Issue #334, opened 2026-04-23
Re-run Phase 2 scaffolding review for hex-matrix and hex-gram-schmidt


kim-em:
**Current state**

Both `hex-matrix` and `hex-gram-schmidt` carry `status/hex-*.scaffolding-reviewed` tokens and have `libraries.yml[*].done_through = 2`. But the Phase 2 reviewer missed multiple wrong-but-plausible scaffolds (see #332), and the [PLAN/Phase2.md](https://github.com/kim-em/hex/blob/main/PLAN/Phase2.md) review questions have since been updated (https://github.com/kim-em/hex/pull/331) to explicitly require flagging such scaffolds. The review attestations for these two libraries are therefore stale.

depends-on: #332

**Deliverables**

In one PR:

1. `rm status/hex-matrix.scaffolding-reviewed status/hex-gram-schmidt.scaffolding-reviewed`.
2. In `libraries.yml`, set both `HexMatrix.done_through` and `HexGramSchmidt.done_through` back from `2` to `1`. This will (correctly) also block `HexMatrix` and dependents from Phase 3 until the re-review lands, matching the SPEC state of the libraries.
3. Verify `python3 scripts/check_dag.py` and `python3 scripts/status.py` still succeed — they should now report both libraries as Phase 2 ready, and all downstream libraries as blocked on `dt >= 2` for their deps.

Then the planner should dispatch two fresh `agent-plan review` issues (one per library) against the updated [PLAN/Phase2.md](https://github.com/kim-em/hex/blob/main/PLAN/Phase2.md) — specifically the new "does every body actually implement the SPEC contract, or is it `sorry`?" question. The reviewer agent produces a fresh token only after that check passes.

**Context**

- Updated Phase 2 rule: https://github.com/kim-em/hex/blob/main/PLAN/Phase2.md#review-questions
- Updated Phase 1 rule: https://github.com/kim-em/hex/blob/main/PLAN/Phase1.md#rules
- Related: https://github.com/kim-em/hex/issues/332 (fixing the bad scaffolds), must land first.

**Verification**

- `status/` directory no longer contains the two tokens.
- `libraries.yml` shows `done_through: 1` for both libraries.
- `python3 scripts/status.py` reports both as Phase 2 ready and all downstream libraries as blocked appropriately.
- `lake build` still succeeds.

**Out of scope**

- Actually doing the re-review — the planner dispatches that as separate `agent-plan` issues once this cleanup lands. This issue is pure state rewinding.

[...]

Issue #333, opened 2026-04-23
Remove scaffold-locking #guard lines from HexMatrix/Conformance.lean and HexGramSchmidt/Conformance.lean


kim-em:
**Current state**

[SPEC/testing.md](https://github.com/kim-em/hex/blob/main/SPEC/testing.md) now forbids scaffold-locking `#guard`s — assertions against current stub output rather than against the SPEC contract (landed in https://github.com/kim-em/hex/pull/331). Two committed conformance modules contain such assertions.

These `#guard`s pass only because the underlying scaffold returns wrong-but-plausible output (issue #332). Once #332 rewrites those bodies as `sorry`, these assertions would either break (if `#eval` is used) or silently accept the `sorryAx` result (if `#eval!` is used). Either way they're not actually testing the contract — they're testing the stub.

depends-on: #332

**Deliverables**

Remove the following `#guard` lines (they come back when the real `rref` / `spanCoeffs` / Gram-Schmidt implementations land, asserting the correct contract at that point):

1. [HexMatrix/Conformance.lean:266-314](https://github.com/kim-em/hex/blob/main/HexMatrix/Conformance.lean#L266-L314) — `rref` rank/echelon/transform assertions, `IsEchelonForm.freeCols` / `spanCoeffs` / `spanContains` / `Matrix.spanCoeffs` assertions, and `nullspaceMatrix` / `nullspace` / `Matrix.nullspace` assertions (these last currently pass only because `rref` returns rank 0 for every input, making every column free; remove until `rref` is real).

2. [HexGramSchmidt/Conformance.lean:139-143](https://github.com/kim-em/hex/blob/main/HexGramSchmidt/Conformance.lean#L139-L143) — `GramSchmidt.Int.coeffs` equals identity (passes because `coeffs` is `Matrix.identity`).

3. [HexGramSchmidt/Conformance.lean:204-209](https://github.com/kim-em/hex/blob/main/HexGramSchmidt/Conformance.lean#L204-L209) — `GramSchmidt.Rat.coeffs` diagonal = 1 and off-diagonal = 0 (same issue).

4. Update the module docstring of each file to remove the now-unsupported coverage claims — e.g. in `HexMatrix/Conformance.lean` the "Covered operations" and "Covered properties" lines that mention `rref`, `spanCoeffs`, `spanContains`, `nullspaceMatrix`, `nullspace`, `Matrix.nullspace`.

**Context**

[...]

Issue #332, opened 2026-04-23
Fix wrong-but-plausible scaffolds in HexMatrix + HexGramSchmidt (replace with sorry)


kim-em:
**Current state**

[PLAN/Phase1.md](https://github.com/kim-em/hex/blob/main/PLAN/Phase1.md) now forbids wrong-but-plausible scaffolds (landed in https://github.com/kim-em/hex/pull/331). Six existing `def`s violate the new rule by returning trivial output (`0`, input unchanged, `Matrix.identity`, `none`) where the SPEC contract promises real computation.

**Deliverables**

Rewrite each of the following as either (a) a correct implementation, or (b) `noncomputable def ... := sorry` (preferred for this issue — a real implementation is Phase 5 work, not Phase 1 rework):

1. [HexMatrix/Rref.lean:17-21](https://github.com/kim-em/hex/blob/main/HexMatrix/Rref.lean#L17-L21) — `rref` returns `rank = 0`, echelon = input, transform = identity.
2. [HexMatrix/Span.lean:43-47](https://github.com/kim-em/hex/blob/main/HexMatrix/Span.lean#L43-L47) — `IsEchelonForm.spanCoeffs` returns `some 0` only on rank-0, else `none`.
3. [HexMatrix/Span.lean:73-75](https://github.com/kim-em/hex/blob/main/HexMatrix/Span.lean#L73-L75) — `Matrix.spanCoeffs` always returns `none`.
4. [HexGramSchmidt/Int.lean:63-64](https://github.com/kim-em/hex/blob/main/HexGramSchmidt/Int.lean#L63-L64) — `coeffs` returns `Matrix.identity`.
5. [HexGramSchmidt/Rat.lean:27-28](https://github.com/kim-em/hex/blob/main/HexGramSchmidt/Rat.lean#L27-L28) — `basis` returns input unchanged.
6. [HexGramSchmidt/Rat.lean:35-36](https://github.com/kim-em/hex/blob/main/HexGramSchmidt/Rat.lean#L35-L36) — `coeffs` returns `Matrix.identity`.

Then run a repo-wide audit to catch anything the first pass missed:

```
git grep -n 'return input unchanged\|return identity\|Matrix.identity' -- '*.lean'
git grep -nE 'def [a-zA-Z]+ .* := [A-Z][a-zA-Z]*.identity$' -- '*.lean'
[...]

PR #384, opened 2026-04-23
spec: strengthen rule — no data-level placeholders of any kind

kim-em:
## Summary

Follow-up to #331, which I wrote too permissively. The rule said "correct implementations, or honest placeholders — no middle ground" and suggested `noncomputable def ... := sorry` as an acceptable placeholder form. Kim's pushback: that's still an escape valve.

This PR removes the escape valve. The Phase 1 rule is now:

> **No data-level placeholders — at all.** Every `def`, `structure` field, `class`, and `instance` must have a body that *correctly* implements the SPEC contract. No `sorry` bodies, no `axiom`s standing in for data, no trivial wrong-but-plausible returns (input unchanged, `0`, `Matrix.identity`, `none`). If you can't implement it correctly in this PR, **don't commit the declaration** — leave it out of Lean until a future PR can implement it correctly. The SPEC file records what's designed; the Lean surface records what's implemented correctly so far.

## Why not `sorry`

A `sorry`-bodied data-level def makes every caller's `#eval` refuse to run (Lean conservatively refuses any term transitively depending on `sorry`). Any conformance test against that caller falls back to `#eval!`, which silently accepts whatever `sorryAx` produces at runtime — neither correct nor obviously wrong, just untrustworthy. An omitted declaration, by contrast, fails the build at every use site, pointing the next agent directly at the implementation gap.

Theorem-level `sorry`s stay allowed (they're Prop-valued and don't propagate into computation). Scope of the ban is data only.

## Corresponding tightening in Phase 2 and SPEC/testing.md

- [PLAN/Phase2.md](https://github.com/kim-em/hex/blob/main/PLAN/Phase2.md) review question: "rewrite as `sorry`" is explicitly NOT an acceptable fix for a flagged placeholder. Either the body becomes correct or the declaration is removed from the tree.

- [SPEC/testing.md](https://github.com/kim-em/hex/blob/main/SPEC/testing.md) scaffold-locking anti-pattern text updated: since placeholders are banned upstream, a conformance test that has to lock in trivial output is really a symptom of an illegally committed placeholder.

## Impact on pending issues

Issues #332, #333, #334 currently describe the fix as "rewrite as `sorry` / noncomputable". That fix is no longer valid under this rule. I'll update those issue bodies to say "either correctly implement the declaration, or delete it (along with every downstream use)" — follow-up.

## Verification

- `lake build` still succeeds — all edits are Markdown.
- `python3 scripts/check_dag.py`, `python3 scripts/status.py` unchanged.

🤖 Prepared with Claude Code

### DIFF
--- a/PLAN/Phase1.md
+++ b/PLAN/Phase1.md
@@ -20,18 +20,30 @@ unless it explicitly fixes the public API or theorem split.
 
 ## Rules
 
-- **Correct implementations, or honest placeholders — no middle
-  ground.** Every `def`, `structure` field, `class`, and `instance`
-  must have a body that either
-  (a) *correctly* implements the SPEC contract for that declaration,
-  or
-  (b) is `sorry` (making the declaration `noncomputable`, trapping
-  at runtime via `sorryAx`, and leaving a visible warning).
-  **Placeholder bodies that typecheck but return wrong-but-plausible
-  output are forbidden** — they lie to downstream callers and invite
-  scaffold-locking conformance tests that pretend the stub is right.
+- **No data-level placeholders — at all.** Every `def`, `structure`
+  field, `class`, and `instance` must have a body that *correctly*
+  implements the SPEC contract for that declaration. There is no
+  placeholder form that's acceptable: not wrong-but-plausible bodies
+  (`def rref M := { rank := 0, echelon := M, transform := 1 }`),
+  not `sorry` bodies (`noncomputable def rref ... := sorry`), not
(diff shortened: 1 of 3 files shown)


kim-em:
Claimed PR repair by session `fb616c6c-8aeb-4f28-83ba-22034e68f4d4` on branch `agent/fb616c6c`

kim-em:
Repair complete (session `fb616c6c-8aeb-4f28-83ba-22034e68f4d4`). Verification passed; pushed.

PR #422, opened 2026-04-23
feat: implement executable HexMatrix rref

### LINKED ISSUE #413: Implement executable HexMatrix.rref over fields with row-operation witness
**Current state**

Issue #397 asks for `HexMatrix.rref` in `HexMatrix/Rref.lean` to stop returning the fake `{ rank := 0, echelon := M, transform := 1, pivotCols := #[] }` scaffold. During worker investigation on 2026-04-23, this turned out to be materially larger than a one-file placeholder swap: the generic API is currently `def rref {R} [Field R]`, but an executable Gaussian-elimination implementation needs a concrete equality/zero test strategy for arbitrary field elements, row-operation bookkeeping for the transform witness, and at least one committed executable conformance example.

**Deliverables**

- Replace the placeholder `rref` body in `HexMatrix/Rref.lean` with a real row-reduction implementation that computes:
  - the rank,
  - the reduced echelon matrix,
  - the accumulated left transform witness, and
  - the sorted pivot-column vector.
- Resolve the generic executability question explicitly rather than burying it in local hacks:
  - either keep the current `[Field R]` API with a valid executable zero-test story, or
  - make the minimal justified API adjustment (for example an added decidable-equality requirement) and propagate that change through downstream generic callers in `HexMatrix`, `HexBerlekamp`, and any other affected modules.
- Add at least one committed `HexMatrix/Conformance.lean` guard for a non-trivial matrix whose RREF result is checked concretely.

**Context**

- Parent issue: #397.
- The worker investigation found that deleting `rref` would cascade into `HexMatrix/Benchmark.lean`, `HexMatrix/Nullspace.lean`, and Berlekamp files, so the practical path is implementation-first.
- `HexMatrix/RowOps.lean` already exposes `rowSwap`, `rowScale`, and `rowAdd`, which should be reused.
- Keep the issue scoped to the executable `rref` body plus the minimal downstream fallout from any API change; do not fold `freeCols`/column-partition work into this issue.

**Verification**

- `lake build HexMatrix` succeeds.
- The new conformance guard(s) in `HexMatrix/Conformance.lean` assert a real non-trivial RREF output.
- `python3 scripts/status.py` is unchanged unless the new work intentionally changes project status.

kim-em:
Closes #413

Implements an executable `HexMatrix.rref` over decidable fields, updates the generic nullspace wrapper to use the honest executability bound, and adds committed RREF conformance guards.

### DIFF
--- a/HexMatrix/Conformance.lean
+++ b/HexMatrix/Conformance.lean
@@ -38,8 +38,14 @@ private def serializeVector (v : Vector α n) : List α :=
 private def serializeMatrix (M : Matrix Int n m) : List (List Int) :=
   M.toList.map Vector.toList
 
+private def serializeRatMatrix (M : Matrix Rat n m) : List (List Rat) :=
+  M.toList.map Vector.toList
+
 private def serializeBasis (B : Vector (Vector α m) n) : List (List α) :=
   B.toList.map serializeVector
+
+private def serializePivotCols (v : Vector (Fin n) k) : List Nat :=
+  v.toList.map Fin.val
 
 private def matrixOfList2D! [Inhabited α] (rows : List (List α)) : Matrix α n m :=
   (Matrix.ofList2D (n := n) (m := m) rows).get!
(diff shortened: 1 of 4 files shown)


kim-em:
Claimed PR repair by session `3c2bf085-e768-40cc-9c12-1909dc958023` on branch `agent/3c2bf085`

kim-em:
Repair complete (session `3c2bf085-e768-40cc-9c12-1909dc958023`). Verification passed; pushed.

PR #601, opened 2026-04-28
fix: tighten Nat extGcd body

### LINKED ISSUE #591: human-oversight: tighten the Nat-level HexArith.extGcd body
## Current state

[`HexArith/ExtGcd.lean:15-24`](https://github.com/kim-em/hex/blob/main/HexArith/ExtGcd.lean#L15-L24) defines

```lean
def extGcd (a b : Nat) : Nat × Int × Int :=
  match b with
  | 0 => (a, 1, 0)
  | b' + 1 =>
      let (g, s, t) := extGcd (b' + 1) (a % (b' + 1))
      (g, t, s - t * Int.ofNat (a / (b' + 1)))
termination_by b
decreasing_by
  simp_wf
  exact Nat.mod_lt _ (Nat.succ_pos _)
```

Two problems with the body shape:

1. It computes `a / (b' + 1)` and `a % (b' + 1)` as **two separate divisions** per recursive frame. On bignum `Nat` that is two heap-allocating operations when one fused `Nat.divMod` (or its in-stdlib equivalent) suffices.
2. It destructs `b = b' + 1` and then **reconstructs `b' + 1`** on every recursive call and division. The recursion can take `b` as a single value with a `b ≠ 0` discharge and avoid the reconstruction entirely.

This is not a wrong-shape scaffold in the strict sense — the algorithm is correct and the asymptotic is fine — but an autonomous agent shipped a body that requires two divisions and an avoidable arithmetic round-trip per frame when the obvious write was tighter. It is the same class of pipeline failure as the other `human-oversight` issues: code that should not have landed in this shape on the first pass.

## Deliverables

Rewrite the body to:

- Use a single fused divMod operation per recursive frame (e.g. `let (q, r) := Nat.divMod a b` or whatever the Mathlib-free stdlib equivalent is — confirm what's exposed at the time of writing).
- Take `b` as a single value with a positivity discharge for the `decreasing_by`, not destruct `b = b' + 1`.

The output contract is unchanged: `(extGcd a b).1 = Nat.gcd a b` and the Bezout identity `s * a + t * b = g`. The (currently `sorry`) theorems may stay `sorry` — only the data-level body is in scope here.

If the Mathlib-free Lean stdlib does not expose a `Nat.divMod` returning `(quot, rem)` in one call, document that finding in the PR description and either (a) write a small helper next to the def that does the fused extraction, or (b) accept the two-call form and add an in-line comment explaining that the fused form is unavailable in this stdlib subset. The point is the choice should be conscious, not accidental.

## Context

- PR #580 — landed doctrine edit naming pipeline-failure-class issues.
- The issue surfaced during the same audit that found `pureUInt64ExtGcd` and the fake `mpz_gcdext` extern.

## Verification

- `lake build HexArith` succeeds.
- `#eval HexArith.extGcd 17 5 = (1, -2, 7)`; further spot-checks for `(0, n)`, `(n, 0)`, `(n, n)` edge cases.
- The body contains exactly one `divMod`-style call per recursive frame (or a documented exception per the deliverables note).

## Out of scope

- Everything else: `pureUInt64ExtGcd`, `pureIntExtGcd`, the fake extern, the docstring vocabulary, theorem proofs.

kim-em:
Closes #591

Session: `68151310-1a2b-4d5f-aae3-2fd6d1ea3b48`

2fe9028 fix: tighten Nat extGcd body

🤖 Prepared with Claude Code

### DIFF
--- a/HexArith/ExtGcd.lean
+++ b/HexArith/ExtGcd.lean
@@ -6,19 +6,31 @@ library.
 namespace HexArith
 
 /--
+Quotient/remainder pairing used by `extGcd`.
+
+Lean 4.30's stdlib exposes `Nat.div` and `Nat.mod`, but not a public fused
+`Nat.divMod`; keep the pairing local so the Euclidean step can switch to a
+fused primitive once one is available.
+-/
+private def natDivMod (a b : Nat) : Nat × Nat :=
+  (a / b, a % b)
+
+/--
 Compute the greatest common divisor of `a` and `b` together with
 Bezout coefficients.
 -/
 def extGcd (a b : Nat) : Nat × Int × Int :=
-  match b with
-  | 0 => (a, 1, 0)
-  | b' + 1 =>
-      let (g, s, t) := extGcd (b' + 1) (a % (b' + 1))
-      (g, t, s - t * Int.ofNat (a / (b' + 1)))
+  if hb : b = 0 then
+    (a, 1, 0)
+  else
+    let _ := hb
+    let qr := natDivMod a b
+    let (g, s, t) := extGcd b qr.2
+    (g, t, s - t * Int.ofNat qr.1)
 termination_by b
 decreasing_by
   simp_wf
-  exact Nat.mod_lt _ (Nat.succ_pos _)
+  simpa [natDivMod] using (Nat.mod_lt a (Nat.pos_iff_ne_zero.2 hb))
 
 theorem extGcd_fst (a b : Nat) : (extGcd a b).1 = Nat.gcd a b := by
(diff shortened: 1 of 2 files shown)


kim-em:
Lean 4.30's stdlib in this repo does not expose a public `Nat.divMod` (confirmed with `#check Nat.divMod`, which fails). I therefore left a local `natDivMod` helper next to `HexArith.extGcd` that pairs `Nat.div` and `Nat.mod` intentionally, so the call site can switch to a fused primitive later without changing the recursive shape again.

Issue #4575, opened 2026-05-17
HO-1 substrate: assemble slow-path quadratic arm umbrella for factor_irreducible_of_nonUnit


kim-em:
## Current state

The HO-1 capstone `factor_irreducible_of_nonUnit`
(`HexBerlekampZassenhausMathlib/Basic.lean:239`, tracked by #4170)
must discharge per-entry irreducibility for every branch of the
public `Hex.factor`. The public entry routes through
`factorWithBound = (factorFastWithBound f B).getD (factorSlowWithBound f B)`
(`HexBerlekampZassenhaus/Basic.lean:5329`), so cases where
`factorFastFactorsWithBound f B = none` fall through to
`factorSlowFactorsWithBound`
(`HexBerlekampZassenhaus/Basic.lean:5009`).

The fast-path arm umbrellas are landed or in flight:

* **Fast-path constant arm** (`squareFreeCore.degree?.getD 0 = 0`) —
  landed in #4565 as
  `factor_constant_branch_entry_irreducible_of_choosePrimeData`
  (`HexBerlekampZassenhausMathlib/IntReductionMod.lean:1123`).
  This umbrella also covers the slow-path constant branch because
  the fast path always succeeds when `hdeg`, so the slow-path
  constant arm is unreachable through `factor`.
* **Fast-path small-mod singleton arm** — landed in #4562 / #4564 as
  `factor_small_mod_singleton_branch_entry_irreducible_of_choosePrimeData`
  (`HexBerlekampZassenhausMathlib/IntReductionMod.lean:1023`).
* **Fast-path quadratic arm** (`quadraticIntegerRootFactors? = some _`,
  `B > 1`) — in flight via #4571.
* **Fast-path BHKS arm** — gated on directive #2567.

The slow-path branches the capstone still needs umbrellas for:

* **Slow-path quadratic arm**
  (`quadraticIntegerRootFactors? squareFreeCore = some coreFactors`,
  fast returned `none`) — **unspecced; this issue**.
* **Slow-path exhaustive arm** — #4561, blocked on substrate
  dischargers #4567 / #4568 / #4569.

### Reachability

The slow-path quadratic branch fires through `factor` exactly when
`factorFastFactorsWithBound f B = none` (so `factorSlow` is consulted)
**and** `(normalizeForFactor f).squareFreeCore.degree?.getD 0 ≠ 0`
**and** `quadraticIntegerRootFactors? squareFreeCore = some coreFactors`.

[...]

Issue #4579, opened 2026-05-17
HO-1 substrate: residual irreducibility helper for quadraticIntegerRootFactors? (sub-issue of #4575)


kim-em:
## Current state

The slow-path quadratic arm umbrella
`factor_slow_quadratic_branch_entry_irreducible_of_choosePrimeData`
in `HexBerlekampZassenhausMathlib/IntReductionMod.lean` (landed by
#4575) handles the non-residual quadratic-core sub-case through the
existing `quadraticIntegerRootFactors?_factor_irreducible_of_ne_residual`
plus the new
`reassemblePolynomialFactors_quadratic_irreducible_or_residual_of_expansionComplete`
classifier in `HexBerlekampZassenhaus/Basic.lean`. The residual
sub-case carries a single localised `sorry` per the #4575 sub-issue
allowance.

The same residual obligation will likely be needed by the in-flight
fast-path quadratic arm umbrella #4571 once that lands (the residual
helper is independent of the fast/slow dispatch). This issue is the
shared substrate.

## Deliverables

A Mathlib-bridge theorem in `HexBerlekampZassenhausMathlib/IntReductionMod.lean`:

```lean
theorem quadraticIntegerRootFactors?_residual_irreducible_of_isPrimitive
    {core : Hex.ZPoly} {coreFactors : Array Hex.ZPoly}
    (hprim : (HexPolyZMathlib.toPolynomial core).IsPrimitive)
    (hquad :
      Hex.quadraticIntegerRootFactors? core = some coreFactors) :
    Hex.ZPoly.Irreducible
      (Hex.splitIntegerRootFactorsAux core
        (Hex.integerRootCandidates core)
        (Hex.integerRootCandidates core).length).2
```

(Or, if `splitIntegerRootFactorsAux` and `integerRootCandidates`
remain `private` in `HexBerlekampZassenhaus/Basic.lean`: expose
public accessors or a wrapper that hides the residual term while
exposing irreducibility.)

### Proof outline

The residual is the optional polynomial appended by
`quadraticIntegerRootFactors?` when
`split.1.size > 0 ∧ split.2 ≠ 1 ∧ split.2.degree?.getD 0 ≤ 1`.
Combined with:

* `splitIntegerRootFactorsAux_product` — `split.2 * polyProduct split.1 = core`
  (over `ZPoly`),
* `HexPolyZMathlib.toPolynomial_mul` — transports the product to
  `Polynomial ℤ`,
[...]

Issue #4577, opened 2026-05-17
HO-1 review: PR #4572 fast-path constant arm umbrella


kim-em:
## Current state

PR #4572 (merged commit `a91b25b0`, closing #4565) landed the
fast-path **constant arm** umbrella for the HO-1 capstone
`factor_irreducible_of_nonUnit` (#4170). Three pieces:

1. Mathlib-free branch-shape lemma
   `Hex.factorWithBound_entry_mem_constant_branch_raw`
   (`HexBerlekampZassenhaus/Basic.lean:5627-5643`): given only
   `hdeg : (normalizeForFactor f).squareFreeCore.degree?.getD 0 = 0`,
   identifies each recorded entry as the sign-normalisation of a raw
   factor in the singleton-core reassembly. No `hB_pos` / `hsmall` /
   `hquadratic` premises (the constant branch is the earliest
   dispatch).
2. Two public substrate helpers in `HexBerlekampZassenhaus/Basic.lean`:
   - `Hex.factorWithBound_entry_shouldRecord` (new helper, relays
     `collectFactorMultiplicities_entry_shouldRecord` through the
     `factorWithBound_eq_factorizationOfFactors` rewrite).
   - `Hex.squareFreeCore_eq_one_of_constant_of_ne_zero` promoted
     from `private` to public.
3. Mathlib-bridge umbrella
   `factor_constant_branch_entry_irreducible_of_choosePrimeData`
   (`HexBerlekampZassenhausMathlib/IntReductionMod.lean:1123-1175`):
   takes `f ≠ 0`, `B`, `entry`, `hdeg`, `hentry_mem`, and
   `hcomplete : reassemblyExpansionComplete (normalizeForFactor f)
                #[(normalizeForFactor f).squareFreeCore]`;
   concludes `Hex.ZPoly.Irreducible entry.1`. No `hchoose` premise
   (the constant branch does not invoke `choosePrimeData`); the
   `_of_choosePrimeData` suffix is retained for naming parity with
   the sibling singleton arm #4564.

The proof route chosen — via
`reassemblePolynomialFactors_mem_xPower_or_core_of_expansionComplete`
under `hcomplete`, then ruling out the singleton-core case as `= 1`
via `factorWithBound_entry_shouldRecord` — is a third route distinct
from the two options noted in the issue body (strengthen umbrella
signature, or use `_constant_irreducible_or_repeated_or_one`). The
worker's session notes
[...]

Issue #4170, opened 2026-05-15
HO-1 capstone: discharge executable factor irreducibility bridge after BHKS fast theorem


kim-em:
library: HexBerlekampZassenhausMathlib

depends-on: #2567

## Current state

This remains the public HO-1 bridge target:

```lean
theorem factor_irreducible_of_nonUnit
    (f : Hex.ZPoly) (hf : ¬ Hex.ZPoly.IsUnit f) :
    ∀ entry ∈ (Hex.factor f).factors, Hex.ZPoly.Irreducible entry.1
```

The slow-path and small-mod singleton substrate gaps previously blocking this capstone have been closed. The old non-monic slow-path prerequisite #4880 is closed, and the factor dispatch has since been rewired to the three-tier `factorFast → factorSlowModular → factorSlowTrial` shape from SPEC PR #6580 / issue #6581.

The remaining blocker is the fast BHKS arm. Directive #2567 is still open and is expected to provide BHKS Theorem 5.2 / equivalent analytic content strong enough to prove irreducibility for factors emitted by the fast recovery branch. A worker audit on 2026-06-03 also found that the declaration site for this theorem in `HexBerlekampZassenhausMathlib/Basic.lean` appears before several later/downstream branch irreducibility bridges. Closing this capstone may therefore require either moving the public capstone later or factoring the raw-entry and per-branch dispatch API into an acyclic module visible at the capstone site.

Until #2567 lands and the declaration-order/API split is settled, this issue is not worker-actionable.

## Deliverables after #2567 lands

1. Prove `factor_irreducible_of_nonUnit` for every factor emitted by `Hex.factor f`.
2. Cover all current `factor` arms under the three-tier dispatch: normalization prefix, slow-modular branch, slow-trial branch, small-mod singleton/quadratic/exhaustive sub-branches as applicable, and the fast BHKS branch.
3. If needed, move or factor the required raw-entry/per-branch irreducibility bridges so the public capstone can see them without an import cycle.
4. Do not change executable behavior.
5. Do not introduce `axiom`, `native_decide`, `TODO`, `FIXME`, or new theorem-level `sorry`.

## Out of scope

- `isIrreducible_iff` (#4172).
- `factor_unique` / corrected factorization uniqueness (#4192, #4193, #4194).
- `checkIrreducibleCert_sound` (#4173).
- Proving BHKS Theorem 5.2 / analytic content tracked by #2567.

## Verification

[...]

PR #4580, opened 2026-05-17
feat: slow-path quadratic arm umbrella for HO-1 capstone (#4575)

kim-em:
Partial progress on #4575

Session: `7537fcdc-481b-4bc5-a093-6978c7f7aa61`

80a306cc chore: add progress note for #4575
ebcf59ae feat: add slow-path quadratic arm umbrella for HO-1 capstone
a8c13a93 feat: add factorWithBound_entry_mem_slow_quadratic_branch_raw

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekampZassenhaus/Basic.lean
+++ b/HexBerlekampZassenhaus/Basic.lean
@@ -7639,6 +7639,34 @@ theorem quadraticIntegerRootFactors?_factor_irreducible_of_ne_residual
         · simp [roots, split, hres_deg] at hquad
   · simp [hdeg] at hquad
 
+/-- Reassembled quadratic-root branch classifier under complete expansion.
+Each raw factor is either irreducible or equals the optional residual emitted
+by the integer-root splitter. The complete-expansion side condition strips out
+the normalization repeated-part fallback case present in the unconditional
+classifier `reassemblePolynomialFactors_quadratic_irreducible_or_repeated_or_residual`. -/
+theorem reassemblePolynomialFactors_quadratic_irreducible_or_residual_of_expansionComplete
+    (d : FactorNormalizationData) {coreFactors : Array ZPoly} {factor : ZPoly}
+    (hquad : quadraticIntegerRootFactors? d.squareFreeCore = some coreFactors)
+    (hcomplete : reassemblyExpansionComplete d coreFactors)
+    (hmem : factor ∈ (reassemblePolynomialFactors d coreFactors).toList) :
+    ZPoly.Irreducible factor ∨
+      factor =
+        (splitIntegerRootFactorsAux d.squareFreeCore
+          (integerRootCandidates d.squareFreeCore)
+          (integerRootCandidates d.squareFreeCore).length).2 := by
+  rcases reassemblePolynomialFactors_mem_xPower_or_core_of_expansionComplete d
+      coreFactors factor hcomplete hmem with hx | hcore
+  · exact Or.inl (xPowerFactorArray_irreducible d.xPower factor hx)
+  · by_cases hres :
+        factor =
+          (splitIntegerRootFactorsAux d.squareFreeCore
+            (integerRootCandidates d.squareFreeCore)
+            (integerRootCandidates d.squareFreeCore).length).2
+    · exact Or.inr hres
+    · exact Or.inl
+        (quadraticIntegerRootFactors?_factor_irreducible_of_ne_residual
+          hquad hcore hres)
+
 /-- Membership classifier for a reassembled quadratic-root branch. A raw factor
 is either one of the already-proved irreducible normalization/root factors, the
 normalization repeated-part fallback, or the optional residual appended by the
--- a/HexBerlekampZassenhaus/Basic.lean
+++ b/HexBerlekampZassenhaus/Basic.lean
@@ -5676,6 +5676,41 @@ theorem factorWithBound_entry_mem_fast_core_success_raw
   simpa only [factorWithBound, factorFastWithBound, hfast, Option.map_some,
     Option.getD_some] using hmem
 
+/-- In the slow-path quadratic integer-root branch, every recorded
+`factorWithBound` entry comes from the normalization reassembly whose core array
+is the `coreFactors` produced by `quadraticIntegerRootFactors?`. This branch
+fires when the fast path returns `none` (e.g. `B = 0`, or `B = 1` with the
+fast-core check missing) and the slow path then sees a non-constant square-free
+core for which `quadraticIntegerRootFactors?` succeeds. The branch-shape lemma
+mirrors `factorWithBound_entry_mem_small_mod_singleton_raw`; it leaves the
+mathematical irreducibility argument for the core factors to the bridge
+layer. -/
+theorem factorWithBound_entry_mem_slow_quadratic_branch_raw
+    (f : ZPoly) (B : Nat) (entry : ZPoly × Nat)
+    (hdeg : (normalizeForFactor f).squareFreeCore.degree?.getD 0 ≠ 0)
+    {coreFactors : Array ZPoly}
+    (hquad :
+      quadraticIntegerRootFactors? (normalizeForFactor f).squareFreeCore
+        = some coreFactors)
+    (hfast_none : factorFastFactorsWithBound f B = none)
+    (hmem : entry ∈ (factorWithBound f B).factors.toList) :
+    ∃ raw ∈
+        (reassemblePolynomialFactors (normalizeForFactor f) coreFactors).toList,
+      entry.1 = normalizeFactorSign raw := by
+  rcases factorWithBound_entry_mem_raw_source f B entry hmem with
+    ⟨rawFactors, hsource, raw, hraw_mem, hraw_norm⟩
+  cases hsource with
+  | inl hfast_some =>
+      rw [hfast_some] at hfast_none
+      cases hfast_none
+  | inr hslow =>
+      rcases hslow with ⟨_, hrawFactors⟩
+      subst rawFactors
+      refine ⟨raw, ?_, hraw_norm⟩
+      unfold factorSlowFactorsWithBound at hraw_mem
+      simp only [if_neg hdeg, hquad] at hraw_mem
+      exact hraw_mem
+
 private def quadraticSquareRegression : ZPoly :=
   let q : ZPoly := DensePoly.ofCoeffs #[-1, 0, 1]
   q * q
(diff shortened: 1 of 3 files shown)


kim-em:
## Review: PR #4580 (slow-path quadratic arm umbrella, #4575)

Review work item: #4584. PR landed at `309a5eaa`; this comment records findings against the seven review deliverables.

### TL;DR

The umbrella is correct as merged and the single localised `sorry` is bounded as the issue text claims. Two non-trivial findings to flag:

1. **The fast-path quadratic arm PR #4582 (in flight) does NOT consume the new `_quadratic_irreducible_or_residual_of_expansionComplete` classifier this PR introduced.** It takes a different route — `quadraticIntegerRootFactors?_factor_irreducible_of_primitive` (new in PR #4582) composed with the generic reassembly lift `reassemblePolynomialFactors_factor_irreducible_of_complete_and_core_irreducible`. Once #4582 lands, the slow-path umbrella here can be rewritten in the same shape and the new classifier becomes dead code. Deliverable 2's premise ("the two arms should share the classifier verbatim") doesn't currently hold.
2. **`(_hf_ne : f ≠ 0)` is underscored prematurely.** The sorry-discharge path (via the `IsPrimitive` of `(normalizeForFactor f).squareFreeCore`) needs `f ≠ 0`. Once #4579 lands, `_hf_ne` becomes used and the underscore must come off. Keep the premise; drop the underscore now and use it in the sorry-bearing branch — or better, restructure per Finding 1 below.

### Deliverable 1 — `hfast_none` plumbing

* No `factorWithBound_uses_slow_iff_factorFast_none` style theorem exists. The closest predicate is `factorWithBoundUsesExhaustiveBranch` (`HexBerlekampZassenhaus/Basic.lean:5118`), whose definition opens with `factorFastFactorsWithBound f B = none` — i.e., any compositional premise would just re-export the same fact. The direct `hfast_none` premise is the cleanest choice for the slow-path umbrella; no refactor needed.
* **Capstone discharge sketch (#4170)**: case-split on `factorFastFactorsWithBound f B`. In the `none` case, `hfast_none` is immediate. In the `some` case, the slow-quadratic arm is unreachable through the public entry path: `factorWithBound = (fast).getD (slow)`, so any `entry ∈ (factorWithBound f B).factors.toList` then came from the fast path, and the fast-path quadratic umbrella (PR #4582) covers that branch. No missing substrate.
* **Cross-check with #4582's branch-shape lemma**: PR #4582 introduces `factorWithBound_entry_mem_quadratic_branch_raw` (fast-path), taking `hB_gt_one : 1 < B` instead of `hfast_none`. The asymmetry is correct and unavoidable: the fast-path lemma needs `1 < B` to enter the `factorFastFactorsWithBound` quadratic dispatch arm; the slow-path lemma needs `hfast_none` to even reach the slow branch. They are not refactorable into a single parametric lemma without inventing a fast/slow dispatch flag that buys nothing. Document the asymmetry in a comment (one-line is enough) on whichever arm lands second; the slow-path docstring already alludes to it indirectly.

### Deliverable 2 — Classifier introduction

* **Naming** is correct and follows the sibling `_mem_xPower_or_core_of_expansionComplete` style.
* **Single caller**: `rg` confirms `reassemblePolynomialFactors_quadratic_irreducible_or_residual_of_expansionComplete` is used **only** by this PR's slow-quadratic umbrella. PR #4582's fast-quadratic umbrella took a different, cleaner route.
* **Older unconditional classifier** `_quadratic_irreducible_or_repeated_or_residual` (`HexBerlekampZassenhaus/Basic.lean:7674`): `rg` shows **zero** external callers. It is only declared. The PR did not promote/specialise correctly — it added a new sibling rather than deriving one from the other. Either (a) derive the conditional one from the unconditional one (under `hcomplete`, the `repeatedPartFactorArray` arm is impossible because the repeated part is consumed), or (b) drop the unconditional one if no caller materialises soon. **Suggested follow-up issue**: audit the two classifiers and pick one as derived from the other.
* **Strongest refactor opportunity** (cross-cutting with PR #4582): once #4582 lands, the slow-quadratic umbrella in this file can be rewritten in the same `_of_primitive`-per-core-factor style as the fast-quadratic umbrella (and as the small-mod singleton arm at `IntReductionMod.lean:1023` already does, via `reassemblePolynomialFactors_factor_irreducible_of_complete_and_core_irreducible`). That rewrite would:
    * eliminate the `sorry` (no separate residual case at the classifier level — every core factor goes through `quadraticIntegerRootFactors?_factor_irreducible_of_primitive`),
    * make `_quadratic_irreducible_or_residual_of_expansionComplete` dead code,
    * unify all four landed arm umbrellas around the same lift,
    * make the sub-issue #4579 unnecessary as a separate deliverable (the residual irreducibility is already part of #4582's `_of_primitive` wrapper).
  **Suggested follow-up**: file an issue "Refactor slow-quadratic arm umbrella onto `_factor_irreducible_of_primitive` once #4582 lands; remove `_of_expansionComplete` classifier." This also subsumes part of #4579.

### Deliverable 3 — Residual sub-case sorry and #4579 framing

* The sorry's local context **does** retain `hentry_mem : entry ∈ (Hex.factorWithBound f B).factors.toList`, so `factorWithBound_entry_shouldRecord` could in principle expose `entry.1 ≠ 1 ∧ entry.1 ≠ -1` and hence `raw ≠ 1 ∧ raw ≠ -1` (because `normalizeFactorSign` is `id`-or-negate). But **#4579's planned proof route does not need this** — it goes through `IsPrimitive` of the core. So the recorded-entry filter being dropped at the umbrella level is not actually a hole.
* **#4579's deliverable is largely already done by PR #4582.** PR #4582 introduces `Hex.quadraticIntegerRootFactors?_residual_irreducible` (`private`) and the public wrapper `_factor_irreducible_of_primitive` in `HexBerlekampZassenhaus/Basic.lean`, plus the reverse primitivity bridge `zpoly_primitive_of_toPolynomial_isPrimitive` in `IntReductionMod.lean`. Once #4582 merges, the slow-quadratic sorry can be closed directly using #4582's helpers; #4579's separate Mathlib-bridge theorem becomes redundant. **Strongly recommend #4579 be re-scoped or closed in favour of just wiring** once #4582 lands.
* The inline commentary at `IntReductionMod.lean:1248-1257` is accurate but predates #4582. Once #4582 lands, the commentary should be tightened to point at the executable `_factor_irreducible_of_primitive` wrapper rather than the abstract proof outline.

### Deliverable 4 — `_hf_ne : f ≠ 0` underscore

* The underscore prefix is **premature and will mislead the next reader**. The sorry-discharge path (per #4579, and per #4582's residual proof) uses `(normalizeForFactor f).squareFreeCore.toPolynomial.IsPrimitive`, which is provided by `IntReductionMod.normalizeForFactor_squareFreeCore_toPolynomial_isPrimitive f hf_ne` and **requires** `f ≠ 0`.
* **Recommended fix**: rename `_hf_ne` to `hf_ne` now (keeping the premise). The current proof body does not consume it, but once the sorry is replaced (either by #4579 wiring or by the refactor in Finding 1), `hf_ne` becomes essential. Lean will emit an unused-variable warning until then — that's the right state for a known-pending discharge.
* Parity with siblings: small-mod singleton (`hf_ne`) and constant (`hf_ne`) arms both already take it without an underscore. Slow-quadratic is the outlier.

### Deliverable 5 — `_of_choosePrimeData` suffix

* The slow-quadratic branch genuinely does not invoke `choosePrimeData`. Keeping the suffix is documentary, not load-bearing.
* The same question is open for the constant arm in review #4577 (also unclaimed). **Defer to that review's resolution** and apply the chosen convention uniformly to all four landed arm umbrellas at that point. Whatever is chosen, the docstring already explains the parity rationale (`IntReductionMod.lean:1170`-ish for the slow-quadratic; `1080`-ish for the constant arm) — that's adequate for now.

### Deliverable 6 — `hcomplete` discharge feasibility

* For the quadratic core shape, `hcomplete : reassemblyExpansionComplete (normalizeForFactor f) coreFactors` (where `coreFactors = quadraticIntegerRootFactors? ...`) is **not generically true**. `reassemblyExpansionComplete` unfolds to `(expandRepeatedPartFactorArray d.repeatedPart coreFactors).2 = 1`, i.e., the repeated part is fully consumed by the supplied core factors. The quadratic splitter's output `(roots, optional_residual)` has no a priori reason to divide `d.repeatedPart` cleanly.
* This is the **same gap** flagged in review #4577 deliverable 3 for the constant arm (whose `hcomplete` is for `#[squareFreeCore]`, also not generically satisfied).
* The capstone assembly will need either a companion umbrella for `¬ hcomplete` (routing through `reassemblePolynomialFactors_mem_normalization_or_core`) or a separate substrate theorem proving `hcomplete` holds in the contexts where the capstone applies the umbrellas. **Suggested follow-up issue**: substrate piece for `hcomplete` discharge across all arm umbrellas — single issue covering the constant, singleton, slow-quadratic, fast-quadratic, and (eventually) exhaustive arms uniformly. Cross-link to #4577.

### Deliverable 7 — Inventory deltas

Verified via `rg` and a fresh `lake build HexBerlekampZassenhausMathlib.IntReductionMod`:

* `HexBerlekampZassenhausMathlib/IntReductionMod.lean`: 1 `sorry` (line 1258, the documented residual sub-case for #4579). 0 `axiom`, 0 `native_decide`, 0 `TODO`/`FIXME`.
* `HexBerlekampZassenhaus/Basic.lean`: 1 pre-existing `sorry` (line 5770, `isIrreducible_iff` placeholder, unchanged by this PR).
* Build green: `lake build HexBerlekampZassenhausMathlib.IntReductionMod` (and full `HexBerlekampZassenhausMathlib`) completes successfully on PR HEAD with only the expected `sorry`-use warnings.

### Suggested follow-up issues

1. **Refactor slow-quadratic umbrella once PR #4582 lands**: rewrite onto `quadraticIntegerRootFactors?_factor_irreducible_of_primitive` + `reassemblePolynomialFactors_factor_irreducible_of_complete_and_core_irreducible`. Removes the sorry, removes the single-use `_of_expansionComplete` classifier, removes the need for separate #4579 work. (Supersedes #4579.)
2. **Audit quadratic classifiers in `Basic.lean`**: pick one of `_quadratic_irreducible_or_repeated_or_residual` vs `_quadratic_irreducible_or_residual_of_expansionComplete` as derived from the other (currently both are stand-alone and the unconditional one has zero callers).
3. **`hcomplete` discharge substrate** (cross-cutting with #4577 deliverable 3): a single issue covering `reassemblyExpansionComplete` discharge — or a companion `¬ hcomplete` umbrella — for all arm umbrellas, since the capstone will need it for every arm.
4. **Drop `_` on `_hf_ne` in the slow-quadratic umbrella signature**: trivial one-line edit; can be folded into Follow-up 1.

### Approval

Review accepts PR #4580 as merged. The findings above are forward-looking refactors and follow-up substrate, not blockers on the merged work.

kim-em:
**Update — PR #4582 landed at `0ffd01cd` while this review was being filed.** Follow-up #4587 (which the review framed as "depends on #4582") is now immediately actionable. The cross-PR finding in the review above stands; the refactor it proposes can proceed without further waiting. Have unblocked #4587 accordingly.

PR #6604, opened 2026-06-09
spec: certified external-dispatch LLL with η-parameterized reducedness

kim-em:
This PR specifies a certified external-dispatch path for `hex-lll`: the public `lll` runs the native algorithm by default and, when an optional external reducer is present in the process, instead certifies that reducer's output with a verified Lean checker, falling back to native on absence or rejection. Both paths satisfy the identical post-condition (same lattice, `(δ, 11/20)`-reducedness, the short-vector bound), so dispatch is invisible to callers and to proofs, and `lll`'s guarantee is property-level rather than value-level.

To accommodate a floating-point reducer's size-reduction threshold without a post-pass, reducedness is parameterized by a size-reduction bound η. The native entry `lllNative` carries the classical η = 1/2 contract (α = 1/(δ − 1/4), δ > 1/4), and the public `lll` carries η = 11/20 (α = 1/(δ − 121/400)), the bound any externally-routed function can guarantee uniformly. The short-vector bound is factored through one lemma over η, instantiated for both surfaces.

Design-principles §4 and the `@[extern]` policy gain carve-outs for a checked external oracle and an untrusted dispatch hook: correctness depends only on the verified checker and the native fallback, never on the external reducer, so the trusted computing base is unchanged. The certificate is `(B', U, V)` with `U·B = B'` and `V·B' = B` for same-lattice (no determinant, no inverse lemma) plus an executable integer reducedness check over the `d`/`ν` representation; `certCheck_sound` is the single trusted bridge the dispatch depends on.

🤖 Prepared with Claude Code

### DIFF
--- a/SPEC/Libraries/hex-lll-mathlib.md
+++ b/SPEC/Libraries/hex-lll-mathlib.md
@@ -3,3 +3,18 @@
 Connects hex-lll to Mathlib's linear algebra:
 - Lattice corresponds to a `Submodule ℤ`
 - Short vector bound holds with respect to Mathlib's `norm`
+
+## Headline correctness theorem
+
+`reduced_first_row_norm_sq_le_of_mem_latticeSubmodule`: the Euclidean
+short-vector bound for the public `lll`, stated with Mathlib's
+`EuclideanSpace` norm and `Submodule ℤ` membership. For a `(δ, 11/20)`-reduced
+output and any nonzero lattice vector `x`,
+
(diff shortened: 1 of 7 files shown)


kim-em:
Claimed PR repair by session `64475e1a-af75-48c5-8f1f-7a1356014ff9` on branch `agent/64475e1a`

kim-em:
Repair complete (session `64475e1a-af75-48c5-8f1f-7a1356014ff9`). Verification passed; pushed.

PR #6640, opened 2026-06-09
spec: HexLLL five-curve comparator plot; fpLLL via fplll-ffi (drop fpylll)

kim-em:
This PR extends the HexLLL headline comparator plot to five curves — fpLLL, Lean native, verified Isabelle native, and the two verification curves (Lean certified and verified Isabelle certified-LLL), the last two certifying the *same* fpLLL output with the Lean versus the Isabelle checker, which is the apples-to-apples yardstick for the certified gating goal.

It also moves the fpLLL comparator from a Python `fpylll` subprocess to the `fplll-ffi` FFI shim: an in-process C++ call into `libfplll`, the preferred comparator pattern per `SPEC/benchmarking.md` and the same reducer the certified dispatch resolves at runtime — so the fpLLL curve and the certified path measure the identical C++ reducer rather than paying Python-interpreter and IPC overhead the runtime never pays. The bench obtains `fplll-ffi` via a setup script plus a path override and `dlopen` (mirroring the Isabelle binary-path overrides), keeping hex free of any build dependency on it. The conformance oracle (`SPEC/testing.md`) continues to use `fpylll` for correctness cross-checks; only the speed comparator changes.

🤖 Prepared with Claude Code


kim-em:
Claimed PR repair by session `1bebd07c-7ed1-4563-a85f-b29f65abca75` on branch `agent/1bebd07c`

kim-em:
Repair complete (session `1bebd07c-7ed1-4563-a85f-b29f65abca75`). Verification passed; pushed.

Issue #6890, opened 2026-06-13
BZ recovery: repair multiplicative closure for recovered lifted predicate


kim-em:
library: HexBerlekampZassenhausMathlib

Residual from #6890 after worker decomposition into #6932, #6819, #6840, and #6804.

depends-on: #6932

## Current state

The broad predicate-swap issue was decomposed because a direct replacement of `RepresentsIntegerFactorAtLift` exposed several independent migration surfaces:

- #6932 owns the immediate public predicate-carrier swap to a proof-level recovered/monic-coordinate shape and the direct abstract structure/projection cluster.
- #6819 owns downstream lifted-partition and substrate consumers after the corrected predicate lands.
- #6840 owns the recursive scaled-recombination search proof migration to the recovered candidate shape.
- #6804 tracks the broader scaled-recombination recovery and coverage chain.

One parent deliverable is not clearly owned by those sub-issues: repairing or restating

```lean
representsIntegerFactorAtLift_mul_of_monic_core
```

after `RepresentsIntegerFactorAtLift` becomes the recovered/monic-coordinate predicate. In the old scaled-product predicate, the theorem destructs a congruence directly. Under the corrected existential carrier, multiplicative closure needs an explicit primitive-part/Gauss argument for multiplying monic-coordinate witnesses, or the theorem should be restated if the current statement is not load-bearing. Do not add a stale scaled-product field to the corrected predicate just to preserve the old proof shape.

## Deliverables

- After #6932 lands, repair or restate `representsIntegerFactorAtLift_mul_of_monic_core` against the recovered/monic-coordinate public predicate.
- If the old multiplicative-closure statement is still load-bearing, prove it by multiplying the monic witnesses and supplying the needed primitive-part/dilation argument explicitly.
- If the statement is not load-bearing under the new predicate, replace it with the shortest theorem that downstream callers actually need, and update those callers.
- Keep exact-recovery, lifted-partition consumer, and recursive scaled-recombination search work in #6819/#6840/#6804 rather than folding those broader surfaces back into this issue.

## Context

Search target:

```bash
rg -n "RepresentsIntegerFactorAtLift|RecoveredAtLift|representsIntegerFactorAtLift_mul_of_monic_core" HexBerlekampZassenhausMathlib/Basic.lean
```

[...]

Issue #6804, opened 2026-06-12
feat: migrate the scaledRecombinationCandidate recovery and coverage chain to the dilation model


kim-em:
library: HexBerlekampZassenhausMathlib

depends-on: #6818
depends-on: #6819

## Context

This is the residual of #6801 after the foundation (#6803, the
`ZPoly.dilate` Mathlib correspondence, now landed) and the predicate
redesign (originally #6772). #6772's Mathlib-layer deliverable was never
landed — it was auto-closed by the executable-only PR #6802, then reopened
and decomposed into #6817 (recovered lifted-factor candidate surface, has
PR), #6818 (redefine `RepresentsIntegerFactorAtLift`/`scaledLiftedFactorProduct`
around the recovered candidate), and #6819 (migrate the lifted-partition
consumers). This issue depends on that chain completing (#6818, #6819);
the dilation recovery lemmas below are unprovable against the current
scale-based predicate (their hypotheses are vacuous for non-monic factors).

Post #6802 the executable slow-path candidate is
`primitivePart (ZPoly.dilate coreLc (centeredLift product))`
(`HexBerlekampZassenhaus/Basic.lean:6550`). The Mathlib model candidate
`scaledRecombinationCandidate` (`HexBerlekampZassenhausMathlib/Basic.lean:2249`)
is still the old `primitivePart (centeredLift (scale coreLc product))`. The
two are different polynomials for non-monic `core`, so the executable seam
breaks: `HexBerlekampZassenhausMathlib/Basic.lean:15990` passes a scale-shaped
`hprefix`/`hcandidate_def` to the now-dilation executable lemma
`Hex.scaledRecombinationSearchModAux_eq_some_of_step_of_prefix_none`.

This is currently the **only** error in the file (the rest of the model is
self-consistent with the old scale shape), but it cannot be patched at the
seam: the executable emits a genuinely different value, so the model candidate
must be redefined to the dilation shape, which cascades through every lemma
that unfolds and rewrites the candidate.

## Current state

- `scaledRecombinationCandidate` (`Basic.lean:2249`) and `liftedFactorProductCandidate` (`:2239`) still build the scale/centeredLift shape.
- The recovery lemmas encode the scale shape and currently compile self-consistently:
  - `centeredLiftPoly_scaledLiftedFactorProduct_eq_factor_of_recovery*` (`:4887`, `:4911`)
[...]

Issue #2564, opened 2026-05-07
HO-1: BZ recombination implements wrong lattice — rollback and rewrite to van Hoeij CLD


kim-em:
## Current state

`HexBerlekampZassenhaus.recombineLLL?` ([HexBerlekampZassenhaus/Basic.lean#L711](../blob/main/HexBerlekampZassenhaus/Basic.lean#L711)) is currently the production recombination path. It builds an additive-coefficient lattice via [`recombinationLattice?`](../blob/main/HexBerlekampZassenhaus/Basic.lean#L633); short-vector decoding via [`decodeShortVector`](../blob/main/HexBerlekampZassenhaus/Basic.lean#L668) reads off `Σ λ_i f_i (mod p^k)`. True integer factors are *products* `∏_{i ∈ S} f_i (mod p^k)`, not sums, so this lattice cannot in principle recover factorisations whose true factors are non-trivial subset products of mod-p factors.

`SPEC/Libraries/hex-berlekamp-zassenhaus.md` (post #2563, further updated by #5767) prescribes the BHKS all-coefficients CLD lattice with rref + equivalence-class recovery and a fallback combinator architecture: `factorSlow` (unconditionally correct exhaustive enumeration), `factorFast : ZPoly → Option (Array ZPoly)` (conditionally correct van Hoeij CLD), and `factor := (factorFast f).getD (factorSlow f)` (unconditionally correct combinator).

`HexBerlekampZassenhaus.done_through: 2` and `HexBerlekampZassenhausMathlib.done_through: 3` overstate readiness against the new spec.

## Symptom

**Declared expectation.** Recombination must implement the BHKS van Hoeij CLD lattice per SPEC §"Fast path: van Hoeij CLD lattice", with the conditional-correctness contract `factorFast f = some gs ⟹ gs is the irreducible factorisation of f`.

**Observed result.** Current `recombineLLL?` builds an additive-coefficient lattice, decoding short vectors as sums of lifted factors. On adversarial inputs (`X⁴ + 1`, Swinnerton-Dyer SD₃, `Φ₁₅`) the algorithm cannot in principle find the correct factorisation. Current conformance hides this because every fixture has true factors coinciding with single mod-p factors.

[...]

PR #6920, opened 2026-06-13
BZ recovery: decouple stale scaled recovery lemmas from representation predicate

### LINKED ISSUE #6888: BZ recovery: decouple stale scaled recovery lemmas from representation predicate
library: HexBerlekampZassenhausMathlib

Parent: #6881. Depends on #6886.

## Current state

After #6886, the corrected monic-coordinate carrier exists alongside the old `RepresentsIntegerFactorAtLift`. The next blocker to redefining `RepresentsIntegerFactorAtLift` is that several old recovery lemmas use `RepresentsIntegerFactorAtLift` to prove exact equality for the scaled selected product. That is not derivable from the corrected carrier.

The first concrete failure from a direct predicate swap is around `HexBerlekampZassenhausMathlib/Basic.lean:2675` in
`centeredLift_scaledLiftedFactorProduct_eq_of_mignottePrecision_of_bound`:

```lean
Hex.centeredLiftPoly_eq_of_reduceModPow_eq ... hrep
```

where `hrep` used to be the old scaled congruence. With the corrected predicate, `hrep` is an existential unscaled congruence plus dilation keystone instead.

## Deliverables

- Refactor the stale scaled-product recovery lemmas so they no longer consume `RepresentsIntegerFactorAtLift` as the source of the old congruence. Use an explicit old-shape congruence hypothesis when the scaled-product equality itself is still needed.
- Preserve downstream theorem names only where their statements remain semantically correct. If a theorem's conclusion is specifically about `scaledLiftedFactorProduct`, its hypotheses should say so explicitly rather than hiding that stale shape behind `RepresentsIntegerFactorAtLift`.
- Add/adjust wrappers for callers that are genuinely monic-core old-shape users, making the monic collapse explicit rather than relying on the public representation predicate.
- Do not redefine `RepresentsIntegerFactorAtLift` in this issue; the goal is to remove the false dependency so the final swap is mechanical.

## Context

Likely affected region starts near:

```bash
sed -n '2650,2760p' HexBerlekampZassenhausMathlib/Basic.lean
sed -n '4980,5455p' HexBerlekampZassenhausMathlib/Basic.lean
```

A direct predicate swap also leaves the pre-existing #6804 failure at `Basic.lean:16402`; this issue should not try to repair that scaled/dilate recursive-search prefix mismatch.

## Verification

- `lake build HexBerlekampZassenhausMathlib.Basic` should reach only the pre-existing #6804 failure at `Basic.lean:16402`, not introduce earlier errors.
- No new `sorry` / `axiom` / `native_decide` / `TODO` / `FIXME` on added lines.
depends-on: #6886

kim-em:
Closes #6888

Session: `d302787e-c97c-4771-8d21-8a9940b25595`

91992821 refactor: decouple scaled recovery congruence (#6888)

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekampZassenhausMathlib/Basic.lean
+++ b/HexBerlekampZassenhausMathlib/Basic.lean
@@ -2669,18 +2669,20 @@ def henselSubsetCorrespondence_of_modPSubsetPartition
 /--
 Abstract-bound variant of
 `centeredLift_scaledLiftedFactorProduct_eq_of_mignottePrecision`: takes an
-arbitrary `B' : Nat` and an explicit validity hypothesis
-`hvalid : ∀ i, (factor.coeff i).natAbs ≤ B'` in place of the core-shape
-`defaultFactorCoeffBound core` precision constraint.  The body just
+arbitrary `B' : Nat`, an explicit validity hypothesis
+`hvalid : ∀ i, (factor.coeff i).natAbs ≤ B'`, and the scaled-product
+congruence in place of the public representation predicate.  The body just
 threads `B'` and `hvalid` into `centeredLiftPoly_eq_of_reduceModPow_eq`
-(which already accepts an abstract bound).  The original core-shape
-theorem is a wrapper around this variant.
+(which already accepts an abstract bound).  The original core-shape theorem is
+a wrapper around this variant.
 -/
 theorem centeredLift_scaledLiftedFactorProduct_eq_of_mignottePrecision_of_bound
     {core factor : Hex.ZPoly} {d : Hex.LiftData} {S : LiftedFactorSubset d}
     (B' : Nat)
     (hvalid : ∀ i, (factor.coeff i).natAbs ≤ B')
-    (hrep : RepresentsIntegerFactorAtLift core d factor S)
+    (hscaled :
+      Hex.ZPoly.reduceModPow (scaledLiftedFactorProduct core d S) d.p d.k =
+        Hex.ZPoly.reduceModPow factor d.p d.k)
     (hprecision : 2 * B' < d.p ^ d.k) :
     Hex.centeredLiftPoly
         (Hex.ZPoly.reduceModPow (scaledLiftedFactorProduct core d S) d.p d.k)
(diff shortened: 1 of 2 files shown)


kim-em:
## Review (issue #6923) — approve

Reviewed the diff for PR #6920 (closes #6888, directive #2564 recovery-model migration). This is a clean, definitionally-transparent hypothesis swap with no soundness concerns.

### Soundness of the refactor (deliverable 1)

The key fact: `RepresentsIntegerFactorAtLift core d factor S` is defined (`Basic.lean:2283-2287`) as *exactly*

```lean
Hex.ZPoly.reduceModPow (scaledLiftedFactorProduct core d S) d.p d.k =
  Hex.ZPoly.reduceModPow factor d.p d.k
```

which is byte-for-byte the new `hscaled` hypothesis every edited lemma now takes. So the old code only typechecked because `hrep` was *defeq* to this congruence (the lemmas already threaded it straight into `centeredLiftPoly_eq_of_reduceModPow_eq` / `centeredLiftPoly_eq_of_reduceModPow_eq`, which consume the raw `reduceModPow = reduceModPow` shape). The PR makes that unfolding explicit. This is precisely the decoupling #6888 wants: once #6890 redefines the predicate to the `RecoveredAtLift` monic-coordinate carrier (`Basic.lean:2298`), these stale scaled-product lemmas are unaffected because they no longer mention the predicate.

All call sites passing the new `hscaled` are within the edited region and consistently updated. The local `have hscaled : scaledLiftedFactorProduct … = liftedFactorProduct …` in `recombinationCandidate_eq_factor_of_recovery_of_monic_core_of_bound` was correctly renamed to `hscaled_eq` to avoid clashing with the new parameter.

### No silent weakening / no predicate redefinition (deliverable 2)

Every theorem's conclusion is unchanged (`recombinationCandidate d S = factor`, `centeredLiftPoly … = factor`, etc.). Only hypotheses changed, and each lemma whose conclusion is about `scaledLiftedFactorProduct` now states that stale shape explicitly in its hypothesis. `RepresentsIntegerFactorAtLift` is **not** redefined here (`def` at `Basic.lean:2283` is untouched) — that remains #6890's job.

### No new sorry/axiom/etc.; no earlier breakage (deliverable 3)

- No `sorry` / `axiom` / `native_decide` / `TODO` / `FIXME` in the added lines.
- `python3 scripts/check_dag.py` → exit 0.
- No `hrep :=` named-argument call sites anywhere in the library, so the `hrep`→`hscaled` rename breaks nothing by name.
- The single pre-seam external call site (`Basic.lean:14570`, `centeredLiftPoly_scaledLiftedFactorProduct_eq_factor_of_recovery hcore_ne hfactor_dvd_core hrep hprecision`) passes its local `hrep` positionally; since the new parameter type is the reducible-`def` unfolding, it still elaborates by defeq — no error before the #6804 seam.

The build is expected to reach only the pre-existing #6804 `scale`→`dilate` mismatch (author reports it now at `Basic.lean:16475`, shifted from `:16402` by the +61 net lines), which is out of scope here.

### Method / residual risk (deliverable 4)

Verified by static analysis (definition check + call-site/defeq audit) rather than a full local `lake build`: the `HexBerlekampZassenhausMathlib` Mathlib layer is not CI-built and a whole-library build is expected-red at the #6804 seam, so a full build adds no signal over the defeq argument for this change. At review time the ubuntu `build` and `conformance` checks were still in progress (`build-macos` green); these do not exercise this file, but let them finish green before merge per branch protection.

Verdict: **approve**, no blocking findings. Self-approval is blocked on the shared agent account, so recording the verdict here.

Issue #6857, opened 2026-06-13
Resultant: monic common-factor Sylvester column transform


kim-em:
library: HexBerlekampZassenhausMathlib

Parent: #6848. The common-factor column-reduction *step* mechanism landed via PR #6885 (partial). This issue now tracks the remaining **assembly** of the full determinant-preserving matrix `A'`.

## Landed so far (PR #6885)

Four reusable lemmas in `HexBerlekampZassenhausMathlib/Resultant.lean`:

- `sylvesterMap_commonFactor_smul` — `c`-scaled syzygy image `C c * ((r*b - s*a) X^t)`;
- `sylvester_mulVec_commonFactor_smul` — the selected column combination is entrywise `c`-divisible;
- `det_updateCol_mulVec` — replacing a column by `A.mulVec w` multiplies the determinant by the pivot coefficient `w p`;
- `sylvester_commonFactor_colReduceStep` — the capstone combining the two: one column-reduction step.

## Remaining deliverable

Iterate `sylvester_commonFactor_colReduceStep` over the `d = q.natDegree` shifts to produce the assembled matrix `A'`, exposed as either an explicit determinant equality or a constructed matrix with:

- `A'.det` related to `(Polynomial.sylvester f g f.natDegree g.natDegree).det`, and
- the chosen `d` columns of `A'` entrywise divisible by `m`,

under the explicit-witness hypotheses reduced modulo `m`.

The pivot coefficient `w p` equals the cofactor leading coefficient `leadingCoeff b` (`= leadingCoeff g`, since `q` is monic). Choose one of:

- **monic-cofactor side condition** (the title's monic-triangular case): each pivot is a unit, so `A'.det = sylvester.det` exactly; or
- **general form**: `A'.det = (leadingCoeff g)^d * sylvester.det`, recording the non-unit pivots.

Document whichever side conditions are used in the theorem docstring so #6858 can consume them. Do not package the final `m ^ d ∣ resultant` theorem here.

## Assembly recipe (from the PR #6885 worker)

1. Pivots `p_t = (⟨(n-d)+t⟩ : Fin n).natAdd m` (top `d` `f`-shift columns), where `m = f.natDegree`, `n = g.natDegree`, `d = q.natDegree`.
2. Process `t` in **decreasing** order: each direction `w_t` references pivots `p_{t'}` only for `t' ≤ t`, so decreasing order keeps every referenced column unreplaced and each combination stays the clean `c`-divisible vector.
[...]

PR #6927, opened 2026-06-13
feat: package map-to-ZMod resultant divisibility wrapper

### LINKED ISSUE #6849: Resultant: package p-adic common-factor divisibility
library: HexBerlekampZassenhausMathlib

Parent: #6843. Depends on #6858.

## Current state

#6843 asks for the load-bearing BHKS Lemma 3.2-style resultant divisibility theorem. The original intermediate issue #6848 was decomposed into:

- #6856: common-factor Sylvester syzygy identities;
- #6857: monic determinant-preserving column transform;
- #6858: explicit-witness common-factor divisibility.

Once #6858 proves the explicit-witness Sylvester statement, this issue packages the public theorem in the form downstream Hensel/CLD code can consume. The witness-extraction helper from #6851 is already available.

## Deliverables

- Add a public theorem over `Polynomial ℤ` with hypotheses shaped for composition:

  ```lean
  q.Monic → q.natDegree = d →
  (q.map (Int.castRingHom (ZMod (p^k))) ∣ f.map (Int.castRingHom (ZMod (p^k)))) →
  (q.map (Int.castRingHom (ZMod (p^k))) ∣ g.map (Int.castRingHom (ZMod (p^k)))) →
  ((p ^ (k * d) : ℕ) : ℤ) ∣ Polynomial.resultant f g
  ```

  Adjust the exact signature if the explicit-witness version from #6858 exposes a cleaner API, but keep a map-to-`ZMod (p^k)` wrapper because downstream Hensel congruence code is expected to feed that shape.
- Prove the wrapper from the explicit quotient/remainder witnesses used by #6858 and `exists_witnesses_of_map_dvd_zmod` from #6851. Do not weaken the theorem to mere `p^k` divisibility.
- Add a short docstring explaining that the exponent is `k * d` because the monic common factor contributes `d` Sylvester directions.

## Context

Parent #6843 explains why `resultant_map_map` alone only gives `p^k ∣ resultant` and is insufficient. Keep this theorem in `HexBerlekampZassenhausMathlib/Resultant.lean` unless the #6858 worker introduced a dedicated helper file.

## Verification

- `lake build HexBerlekampZassenhausMathlib.Resultant`
- `lake build HexBerlekampZassenhausMathlib`
- No new `sorry` / `axiom` / `native_decide` / `TODO` / `FIXME` on added lines.
depends-on: #6858

kim-em:
This PR adds the public BHKS Lemma 3.2-style theorem `pow_dvd_resultant_of_map_dvd` to `HexBerlekampZassenhausMathlib/Resultant.lean`: if a monic degree-`d` polynomial `q` divides both `f` and `g` after reduction modulo `p ^ k`, with `f` monic, then `p ^ (k * d)` divides the integer resultant of `f` and `g`. The exponent is `k * d` rather than `k` because the monic common factor contributes `d` independent Sylvester column directions, each carrying one factor of `p ^ k`. It also adds the private helper `exists_divByMonic_witness`, which produces degree-controlled witnesses `f = q * (f /ₘ q) + C n * s` by honest monic division, the control the existing `commonFactor_dvd_resultant` column reduction needs and which the degree-agnostic `exists_witnesses_of_map_dvd_zmod` cannot supply.

The proof builds monic-division witnesses for both sides, applies `commonFactor_dvd_resultant_of_monic_cofactor` to the swapped pair `(g, f)` so the monic cofactor `f /ₘ q` plays the unit-pivot role, and transfers the conclusion back across `Polynomial.resultant_comm` (the swap differs by a `±1` unit, so divisibility carries through).

Relative to the issue signature this adds `f.Monic` and `d ≤ f.natDegree`, `d ≤ g.natDegree`. The proven column-reduction infrastructure scales the determinant by `(cofactor pivot) ^ d` and needs the pivot to equal `1`, which holds exactly when the divided polynomial is monic; the downstream `inputPolynomial` is the natural monic side. This does not weaken the conclusion (still `p ^ (k * d)`); the fully non-monic form would require new infrastructure to handle the `lc ^ d` pivot factor and is out of scope here.

`lake build HexBerlekampZassenhausMathlib.Resultant` is green and `python3 scripts/check_dag.py` exits 0; the added lines carry no `sorry`/`axiom`/`native_decide`. The whole-library `lake build HexBerlekampZassenhausMathlib` is hard-red on main at `Basic.lean` (a pre-existing `normalizeFactorSign`/`centeredLiftPoly` type mismatch in a file this change does not touch or import); that layer is not CI-built.

Closes #6849

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekampZassenhausMathlib/Resultant.lean
+++ b/HexBerlekampZassenhausMathlib/Resultant.lean
@@ -888,6 +888,94 @@ theorem commonFactor_dvd_resultant_of_monic_cofactor
   exact commonFactor_dvd_resultant f g q a b r s m hf_wit hg_wit hq_monic hq_deg
     hd ha hb hb_pivot
 
+/--
+Degree-controlled witness from a `ZMod n` divisibility by a *monic* `q`.
+
+Unlike `exists_witnesses_of_map_dvd_zmod`, which lifts an arbitrary modular
+quotient, this uses honest monic division: the quotient is `f /ₘ q` (so its
+degree is `f.natDegree - q.natDegree`) and the remainder `f %ₘ q` is
+coefficientwise divisible by `n`, packaged as `Polynomial.C (n : ℤ) * s`. The
+degree control is what the Sylvester column reduction in
+`commonFactor_dvd_resultant` needs.
+-/
+private theorem exists_divByMonic_witness {f q : Polynomial ℤ} {n : ℕ}
+    (hq : q.Monic)
+    (hdvd : q.map (Int.castRingHom (ZMod n)) ∣
+            f.map (Int.castRingHom (ZMod n))) :
+    ∃ s : Polynomial ℤ, f = q * (f /ₘ q) + Polynomial.C (n : ℤ) * s := by
+  have hqmap : (q.map (Int.castRingHom (ZMod n))).Monic := hq.map _
+  -- The monic-division remainder vanishes modulo `n`.
+  have hrem0 : (f %ₘ q).map (Int.castRingHom (ZMod n)) = 0 := by
+    rw [Polynomial.map_modByMonic (Int.castRingHom (ZMod n)) hq]
+    exact (Polynomial.modByMonic_eq_zero_iff_dvd hqmap).mpr hdvd
+  -- Hence each coefficient is divisible by `n`, so `C n` divides `f %ₘ q`.
+  have hCdvd : Polynomial.C (n : ℤ) ∣ (f %ₘ q) := by
+    rw [Polynomial.C_dvd_iff_dvd_coeff]
+    intro k
+    have hc : (((f %ₘ q).coeff k : ℤ) : ZMod n) = 0 := by
+      have h0 : ((f %ₘ q).map (Int.castRingHom (ZMod n))).coeff k = 0 := by
+        rw [hrem0]; simp
+      rwa [Polynomial.coeff_map] at h0
+    exact (ZMod.intCast_zmod_eq_zero_iff_dvd _ n).mp hc
+  obtain ⟨s, hs⟩ := hCdvd
+  refine ⟨s, ?_⟩
+  have hsplit := Polynomial.modByMonic_add_div f q
+  rw [hs] at hsplit
+  linear_combination -hsplit
+
+/--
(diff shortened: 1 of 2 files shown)


kim-em:
Review verdict: approve.

I checked the PR against #6849 and the #6857/#6858 column-reduction infrastructure. The public theorem `pow_dvd_resultant_of_map_dvd` keeps the requested map-to-`ZMod (p^k)` input shape and proves the load-bearing `p ^ (k * d)` divisibility, not the weaker `p ^ k` form. The private `exists_divByMonic_witness` is the right bridge here: it turns modular divisibility by monic `q` into the degree-controlled `f = q * (f /ₘ q) + C n * s` witness that `commonFactor_dvd_resultant_of_monic_cofactor` actually consumes.

I also sanity-checked the added hypotheses. `f.Monic` is justified by the current explicit-witness infrastructure: after swapping to `(g, f)`, the right cofactor is `f /ₘ q`, and the determinant reduction needs that cofactor's pivot coefficient to be `1`. The proof derives this through `Polynomial.leadingCoeff_divByMonic_of_monic`, so this is not cosmetic. The degree side conditions `d ≤ f.natDegree` and `d ≤ g.natDegree` are likewise the side conditions needed for the swapped call's `hd` and the two `natDegree_divByMonic` equalities. Downstream, the bad-vector resultant data uses `inputPolynomial` as the first resultant argument, and the Hensel/toMonic substrate already tracks monic lifted factors; the remaining consumer will still need to provide the input monicity and both modular divisibility facts, but this PR does not silently weaken the directive chain.

Local verification on PR head `39f80fbb`/`1384ceca`:

- `lake build HexBerlekampZassenhausMathlib.Resultant` succeeds. Warnings are pre-existing deprecation/unused-simp warnings in `Resultant.lean` plus pre-existing `HexPolyZ.Basic` linter warnings; no new error surfaced in the target module.
- `python3 scripts/check_dag.py` exits 0.
- `git diff --check` is clean.
- Added Lean source lines contain no `sorry`, `admit`, `axiom`, `native_decide`, `TODO`, or `FIXME`. The only whole-PR scan hit is the progress note text reporting that same clean scan.

No blocking findings. GitHub currently shows `build` and `conformance` still pending, with `build-macos` green.

PR #7008, opened 2026-06-14
doc: warn that tee masks lake build exit code in baseline check

kim-em:
This PR adds a one-line caveat to the hex-lean-mathlib-boundary skill's baseline-build instruction. When an agent captures a baseline `lake build` with `| tee <log>` and runs it in the background, the completion notification reports exit 0 even on a failed build, because the pipeline's exit status is `tee`'s rather than `lake`'s. The caveat tells agents to judge red/green by grepping the log for `error:` / `build failed`, never by the reported exit code. This came up while establishing a baseline for HexBerlekampZassenhausMathlib.Basic, which is red on main; the masked exit code nearly led to reading the failed build as green.

🤖 Prepared with Claude Code


kim-em:
Post-merge review for issue #7012: change is shell-accurate (tee masks lake's non-zero exit absent pipefail), stricter rather than looser (grep log for error:/build failed), well-placed, and scope-clean (skill-doc only). Approve.

PR #7595, opened 2026-06-16
plan: decide BHKS cut separation route

### LINKED ISSUE #7590: plan: decide BHKS cut semantics for true-factor projection
Supersedes #7585. The true-factor cut-projection producer is not currently worker-executable because the analytic heart is under-specified: `bhksWithinGramSchmidtCut` performs a per-row Gram-Schmidt norm cut, so retained rows need not be prefix-closed, and short lattice vectors can require a row that was cut.

## Decision required

Choose the sound executable route for the BHKS cut used by the forward fast-BHKS correctness path:

1. Keep the current per-row cut and add a `CutSeparation` / van Hoeij gap certificate to the relevant producer types. The certificate must be strong enough to prove that all true-block Gram-Schmidt rows are below threshold and all later rows are above threshold, so the retained set behaves like a prefix for the short-vector argument.
2. Change the executable cut in `bhksCutProjectReducedRows` / `bhksWithinGramSchmidtCut` to a sorted or prefix cut whose retained rows are structurally prefix-closed. This changes runtime semantics and needs conformance / benchmark review.

## Evidence from #7585

The false step is survivor-span membership under the current per-row cut. The skipped worker gave the concrete LLL-reduced witness `b0 = (4,0)`, `b1 = (2,3)`: Gram-Schmidt norms are `16` and `9`, and `v = b0 - b1 = (2,-3)` has norm squared `13`. With a cut radius between `9` and `13`, row 0 is cut, row 1 survives, and the short lattice vector `v` is not in the span of survivors.

Restricting the target to true-factor indicators does not by itself fix this: the obstruction is about the reduced-basis expansion of the short vector using a cut row.

## Output

Update the plan with one chosen route and enough concrete successor issue shapes for workers. The successor decomposition should cover:

- the LLL pipeline lattice-membership/span bridge for `lll.shortVectorsUnchecked` through `.toArray`;
- the gap-conditioned or prefix-structured survivor-span theorem;
- the true-factor CLD/lattice bridge placing `(indicatorVector S | cldSum_S)` in the BHKS lattice with the required bound;
- the final producer for `CutProjectionHypotheses` or a narrower local wrapper consumed by the forward-count route.

Do not reissue #7585 as a proof task until this decision is reflected in the types or executable cut semantics.

kim-em:
Closes #7590

Session: `cdc115b0-68ff-4efc-bd2a-7e222dce704c`

a2f7455f plan bhks cut separation route

🤖 Prepared with Claude Code

### DIFF
--- a/progress/20260616T133736Z_cdc115b0.md
+++ b/progress/20260616T133736Z_cdc115b0.md
@@ -1,0 +1,53 @@
+# Progress - 2026-06-16T13:37:36Z (session cdc115b0, work #7590)
+
+## Accomplished
+
+Claimed #7590 after the higher-priority directive #2564 was rejected by
+coordination as blocked. Audited the executable BHKS cut path in
+`HexBerlekampZassenhaus/Basic.lean`, the proof-facing
+`CutProjectionHypotheses` / `CutRetention` surfaces in
+`HexBerlekampZassenhausMathlib/Lattice.lean`, and the forward-count consumers in
+`PartitionRefinement.lean` / `Recovery.lean`.
+
+Chose the sound route for the BHKS cut semantics: keep the current per-row
+executable Gram-Schmidt cut in `Hex.bhksProjectedRows`, and make the needed
+prefix behavior explicit as a proof-facing `CutSeparation` / van-Hoeij gap
+certificate. A first-fail prefix cut would still be false on the known
+LLL-reduced witness, and sorting rows by norm would require a new semantic proof
+for Gram-Schmidt coordinates.
+
+Created critical-path successor feature issues:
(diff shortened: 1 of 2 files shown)


kim-em:
OWNER OVERRIDE: this plan selects option 1 (per-row cut + CutSeparation gap certificate), which is the dead reverse-separation route (see #7590 override comment). Do not merge as the decision-of-record. The chosen route is the BHKS Lemma 5.7 prefix cut (keep b_0..b_t, t=last passing GS index) — an executable change, no gap certificate. Re-do the plan PR for the prefix route or close.

PR #8442, opened 2026-06-29
refactor: encapsulate Hex.Matrix behind a one-field structure

kim-em:
This PR converts `Hex.Matrix` from a transparent `abbrev Matrix R n m := Vector (Vector R m) n` into an opaque one-field structure wrapping the same nested-vector data, and routes every consumer through a published API (`ofFn`/`ofRows`/`rows`/`getRow`/`M[(i, j)]`) so a future flat `Vector R (n*m)` representation becomes invisible outside `HexMatrix/Basic.lean`. Entry access uses the O(1) pair index `M[(i, j)]`; row access `M[i]` routes through a deliberately `noncomputable` instance so any compiled definition that materializes a whole row to read it fails to compile, while proofs may still write `M[i]` freely and executable code uses the computable `getRow`. The data-level row operations (`modify`/`swap`/`mapRows`/`setRow`) destructure the structure so they stay linear in the matrix. A `Mul`-based scalar action, `DecidableEq`/`BEq`, and a `Repr` that displays through the row data keep the structure a drop-in for the former abbrev across the whole graph, benches, and conformance.

🤖 Prepared with Claude Code


kim-em:
Status update (Phase 3, the `abbrev → structure` flip):

`Matrix` is now the opaque one-field `structure` (constructor `ofRows`, field `data`, `rows := M.data`, `@[ext]` on `Matrix.ext`, `deriving DecidableEq, BEq`, and a custom `Repr` that prints through `.rows` so `#eval`/`#guard_msgs` output is unchanged and `data` is never shown).

Green after the flip:
- The whole Mathlib-free library graph (`HexMatrix`, `HexRowReduce`, `HexDeterminant`, `HexBareiss`, `HexGramSchmidt`, `HexLLL`, `HexBerlekamp`, `HexBerlekampZassenhaus`).
- The bench and conformance sub-projects.
- The Mathlib bridges `HexMatrixMathlib`, `HexRowReduceMathlib`, `HexDeterminantMathlib`, `HexBareissMathlib`, `HexLLLMathlib`, `HexBerlekampZassenhausMathlib`.
- Verification: no new `sorry`/`axiom`/`native_decide`; no consumer reads the `data` projection.

Remaining: one file, `HexGramSchmidtMathlib/Int/RowAdd.lean` (~17 errors in ~5 dotProduct/`scaledCoeffMatrix` correspondence proofs). These need per-proof rework from `Matrix.ext; Vector.ext` to `Matrix.ext_getElem` plus body adjustments, because the `Matrix.ofFn`-based defs only reduce cleanly under the pair-form `getElem_ofFn`. Not mechanical; in progress.

Still to do after that: rebase onto current `main` (this branch is based on `0a36a98f`; `main` has since moved to the v4.32 toolchain bump) and re-green, then SPEC updates.

PR #8470, opened 2026-06-30
refactor: migrate the Hex test-kit library to the module system

kim-em:
This PR migrates the shared `Hex` test-kit library (`Hex.Conformance.Emit`, `Hex.BenchOracle.Flint`, and the `Hex` umbrella) to the Lean 4 module system. The mechanical `module`/`public import`/`public section` wrapping was sufficient with no `@[expose]` needed, and the legacy `HexConformance` drivers continue to build against the re-exported API. The full `lake build` and `HexConformance` both build green.

🤖 Prepared with Claude Code

### DIFF
--- a/Hex.lean
+++ b/Hex.lean
@@ -4,8 +4,12 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Kim Morrison
 -/
 
-import Hex.Conformance.Emit
-import Hex.BenchOracle.Flint
+module
+
+public import Hex.Conformance.Emit
+public import Hex.BenchOracle.Flint
+
+public section
 
 /-! Top-level helpers shared by the per-library `Hex<X>` packages.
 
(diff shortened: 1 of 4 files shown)


kim-em:
Closing: migrating the `Hex` test-kit library to the module system breaks the legacy `conformance/*/EmitFixtures.lean` drivers, which `open Hex.Conformance.Emit` and use its helpers — the emit-fixture exe targets (built only in CI's "Build conformance + emit-fixture targets" step, not by `lake build HexConformance`) fail with `Unknown identifier emitPolyFixture` etc. The test kit is a leaf: no `module` file imports it, only these legacy conformance/bench test drivers do. Migrating it would require migrating the whole `conformance/`+`bench/` test-infrastructure (out of the library-migration scope), for no benefit. Leaving it legacy, like HexManual. 🤖 Prepared with Claude Code

PR #8483, opened 2026-06-30
doc: render the kernel-basis recipe output in #m[...] form

kim-em:
This PR wraps the `kernelBasis` how-to recipe's `#eval` in `Matrix.ofRows` so the kernel basis prints as a copy-pasteable `#m[...]` matrix. `Matrix.nullspace` returns a bare `Vector (Vector R m) k`, which has no `#m[...]` `Repr`, so the recipe's expected `leanOutput` had drifted from the raw structure dump the `#eval` actually produced and `lake build HexManual` failed to elaborate it. Collecting the basis vectors as the rows of a matrix restores the intended rendering, and the expected output block is unchanged.

🤖 Prepared with Claude Code


kim-em:
Superseded by #8479, which fixed this recipe on main by switching the `#eval` to `Matrix.nullspaceBasisMatrix` (rendering the basis as the columns of a matrix). Closing as obsolete.

Issue #8531, opened 2026-07-02
test: merge-required conformance case that only the lattice tier can answer


kim-em:
**Parent:** #8417. Requested explicitly: a conformance case baked into every CI run that **only the lattice tier can handle**, so the tier can never be silently lost — not by a dispatch change, not by a lattice regression, not by a precision-cap change like #8521.

## The witness input (verified end-to-end 2026-07-02)

`f = SD5(x)·SD5(x+1)` (Swinnerton-Dyer SD₅ times its shift; deg 64, r = 32 mod the chosen prime, true factorization = two 16-blocks):

- `factorHybridTraced f`: **tier = \"lattice\"**, classical declined after exhausting exactly its 262144-candidate budget, output = the correct 2 factors, product reconstructs. 143 s end-to-end compiled (classical burn 143–193 s run-to-run, lattice alone 20 s).
- Classical **provably** cannot answer it: the size-ordered search needs ΣC(31,≤15) ≈ 2^30 candidates to reach the 16-blocks, ~4000x the budget. No SD-free shape does this job: Swinnerton-Dyer blocks split into ≥ deg/2 factors mod *every* prime, so `choosePrimeData?` cannot dodge the high-r split — that resistance is exactly what forces the lattice arm.
- This was the **first observed** end-to-end traversal of the hybrid's lattice arm: the string \"lattice\" occurs exactly once in `Basic.lean` (the assignment site) and nothing in the corpus, traces, or `#guard`s had ever exercised it.

## Implementation

The existing machinery already does everything once the case exists:

1. Add the case to `conformance/HexBerlekampZassenhaus/EmitFixtures.lean` following the `adv/*` convention (suggest `adv/swinnerton_dyer_sd5_pair`), with pinned coefficients. SD₅ can be generated or pinned as a literal; the product coefficients are 64-bit-exceeding integers, all exact in `ZPoly`.
2. Regenerate the committed fixture (`bz.jsonl`, 100 → 101 cases) and the trace baseline (`bz_trace_gate.py --write`). The trace gate then **pins `tier = \"lattice\"` and the candidate count as a merge requirement** — its stated purpose is failing the merge on \"an unexpected tier downgrade\", which is precisely the protection wanted here.
3. The FLINT oracle (`bz_flint.py`) cross-checks the factorization independently; optionally pin `modFactorPrime`/`modFactorDegrees` metadata like the other `adv/*` cases so the oracle also verifies the r = 32 split.
[...]

Issue #8529, opened 2026-07-02
perf: prune classical recombination candidates before trial division (degree/trailing-coefficient tests)


kim-em:
**Parent:** #8417 (performance side). Measured during the #8521/#8523 review.

## The problem

The classical tier's candidate loop (`scaledRecombinationSmartCandLoop`, `HexBerlekampZassenhaus/Basic.lean`) does zero pruning: **every** enumerated subset builds the full modular product (`Array.polyProduct`), centered-lifts it, dilates by the leading coefficient, takes the primitive part (a content gcd), sign-normalizes, and only then trial-divides (`exactQuotient?`). Additionally `subsetsOfSizeWithComplement` **materializes each size level** (all C(r−1, d) split pairs as a list) before testing any of them.

Measured cost (2026-07-02, compiled, M-series): exhausting the 262144-candidate budget on `SD5(x)·SD5(x+1)` (deg 64, r = 32) takes **143–193 s** — ~5k candidates/s. Reference implementations (FLINT/NTL) reject >99.9% of candidates with O(1) tests before forming any product, running orders of magnitude faster.

## What to do

Add sound rejection tests ahead of product formation, e.g.:
1. **Trailing-coefficient test**: the centered-lifted product of the subset's constant terms (mod p^a) must divide `lc·f(0)` over ℤ — one bignum mul-mod chain and one divisibility check per candidate, no polynomial product.
2. **Degree bound**: skip subsets whose degree sum exceeds `deg(target)/2` (cheap; partially implicit in the size ordering today — make it explicit if not).
3. Enumerate subsets **lazily** instead of materializing each size level.

Soundness requirement: pruning must only reject candidates that provably fail `exactQuotient?`, so the accepted-candidate sequence — and hence every output, fixture, and trace — is **unchanged** except `candidatesTried` semantics: decide (and document) whether pruned candidates still decrement the budget. Keeping budget semantics identical (pruned candidates still count) is the conservative choice — outputs and traces stay byte-identical and only wall-clock improves; changing it alters decline behaviour and belongs to the separate tuning issue.

## Proof-surface caution

The #8413 classical apparatus unfolds these loop definitions. Either preserve the definitional structure (add the tests inside the existing loop bodies and update the unfolding lemmas mechanically), or prove an accepted-candidate-sequence equivalence lemma between old and new loops. Do not weaken any existing theorem statement.

## Why it matters beyond the corpus

[...]

PR #8537, opened 2026-07-02
perf: level-aware classical decline boundary (levelAwareSubsetBudget)

### LINKED ISSUE #8530: perf: level-aware classical decline boundary — BLOCKED on #8417 proof completion
**Parent:** #8417. **BLOCKED: do not start until all #8417 lattice-tier proof work is complete and merged** (the three LatticeTier sorries discharged), and the merge-required lattice conformance case is in CI. Filed now so the measurements and design are not lost.

## The problem (measured 2026-07-02)

For `SD5(x)·SD5(x+1)` (deg 64, r = 32, true factorization two 16-blocks), the hybrid spends **143–193 s** burning the classical tier's full 262144-candidate budget on a search that is hopeless from the first candidate (reaching the size-16 blocks needs ΣC(31,≤15) ≈ 2^30 candidates), then the lattice tier solves it in **20 s**. ~90% of wall time is a wasted probe.

## The design

Classical's candidate budget should be level-aware: once the next subset-size level cannot complete within the remaining budget (C(r−1, d+1) > remaining), finishing the current level and declining certifies everything a partial next level could, so the burn should stop there. Equivalently: an r-aware budget completing size levels only up to d_max(r). For r = 32 this turns a ~150 s decline into a few seconds; for small r nothing changes (the corpus completes far below budget, so fixtures and traces are untouched).

Mechanically the change is proof-light: the #8413 correctness proofs case on `stats.budgetExhausted`, not on the constant `defaultSubsetBudget := 262144`, and the corpus never trips the budget.

## Why it is blocked

1. **Certified-coverage regression until #8417 lands.** Any shrinking of classical's search moves inputs from the tier with unconditional correctness (#8413) to the tier still carrying sorries. Tune only when both tiers are proven.
2. **Dispatch shape is baked into in-flight proofs.** `factorHybridFactors`' structure feeds PR #8517's assembly lemmas; churn invalidates work in progress.
3. **CI must be able to see the boundary first.** The lattice conformance case (separate issue) and #8527's coverage must be merged so tuning cannot silently lose the lattice path.
4. **May be moot after #8529.** If candidate pruning lands first, the burn becomes ~100x cheaper and the wasted probe may be acceptable as-is — re-measure before implementing.

## Verification (when unblocked)

Corpus fixtures and traces byte-identical (small-r behaviour unchanged); the lattice conformance case still dispatches to `tier = \"lattice\"` (and its wall-clock drops — update the trace baseline deliberately); `HexBerlekampZassenhausMathlib` green with no re-broken lemmas; no new `sorry`/`axiom`.

library: HexBerlekampZassenhaus

kim-em:
This PR makes the classical tier's candidate budget level-aware (issue https://github.com/kim-em/hex-dev/issues/8530): `scaledRecombinationSmart` tightens the supplied budget up front to the largest cumulative subset-size-level boundary `∑_{d ≤ k} C(r-1, d)` that fits (`levelAwareSubsetBudget`), so a search that cannot complete declines at the end of the last level it can finish instead of burning the rest of the budget partway into a level it cannot — a partial level certifies nothing beyond the previous boundary. When every level fits (`2^(r-1) ≤ budget`, i.e. `r ≤ 20` at the default 262144), the budget is unchanged, so small-`r` behaviour — the whole corpus — is untouched: fixtures byte-identical, `bz_trace_gate.py` 50/0, `bz_flint.py` 100/0.

Re-measured per the issue's blocker 4 (post-#8529 pruning, M-series): on `SD5(x)·SD5(x+1)` (deg 64, r = 32) the hopeless classical burn now stops at 206368 candidates (the `∑ C(31, ≤5)` boundary) instead of 262144 — 4.71 s → 4.06 s — and the hybrid still dispatches to `tier = "lattice"` with the correct two 16-blocks (probe: hybrid 15.8 s end-to-end, product reconstructs). The win grows with `r`: the cap is 41728 at r = 64 (SD6-shaped) and 1025 at r = 1025. `#guard`s pin the unchanged small-`r` values (r ∈ {3, 19, 20}) and the r = 32 / r = 64 boundaries, cross-checked against an independent Python computation.

Proof impact is as the issue predicted: the correctness lemmas (`RecoveredSmartSearch.{covers,trustworthyNone}_of_bound`) are universally quantified over budget and fuel, so only the two proof sites that spell the budget expression out are re-pointed (`classicalCoreFactorsWithBound_spec`, and the classical irreducibility bridge in `IntReductionMod.lean`). No new `sorry`/`axiom`; `HexBerlekampZassenhaus`, `HexBerlekampZassenhausMathlib`, and `HexConformance` build green.

The cap is applied once at the wrapper from the top-level `r` (a quotient sub-search after a successful peel inherits the remaining budget unaligned to its own levels — harmless for correctness and bounded by the tightened top-level cap; noted in the docstring). The issue's stated pre-condition https://github.com/kim-em/hex-dev/issues/8531 ("test: merge-required conformance case that only the lattice tier can answer") has not landed yet, so lattice dispatch is witnessed here by the measurement probe; #8531's trace assertions should pin the new boundary (`subsetCandidates = 206368`) when it lands.

Closes #8530.

🤖 Prepared with Claude Code


kim-em:
Benchmark data for this change: eight families, before = `main` (8a80dc6a) vs after = this PR, measured with a temporary compiled probe (repurposed `hex_classical_spike`, not committed) on M-series hardware, single-threaded. `classical` times `factorClassicalTraced` alone (the tier this PR touches); `hybrid` is `factorHybridTraced` end-to-end. Reps: 10 for the sub-50ms cases, 3 for the sub-second cases, 1 for the multi-second cases. FLINT = `python-flint` `fmpz_poly.factor()` on the same coefficients (the `bz_flint.py` oracle's engine), for scale.

| family | deg | r | candidates before → after | classical before → after | hybrid before → after | tier before → after | FLINT |
|---|---|---|---|---|---|---|---|
| `(x-1)…(x-16)` | 16 | 16 | 16 → 16 | 25.2 → 23.0 ms | 25.3 → 23.5 ms | classical → classical | 0.17 ms |
| `Φ₁₅` | 8 | 2 | 2 → 2 | 1.1 → 0.9 ms | 1.0 → 1.0 ms | classical → classical | 0.03 ms |
| `SD3` | 8 | 4 | 8 → 8 | 1.7 → 1.6 ms | 1.8 → 1.6 ms | classical → classical | 0.07 ms |
| `SD4` | 16 | 8 | 128 → 128 | 19.0 → 17.1 ms | 19.1 → 17.1 ms | classical → classical | 0.30 ms |
| `SD4(x)·SD4(x+1)` | 32 | 16 | 10503 → 10503 | 279 → 255 ms | 281 → 256 ms | classical → classical | 1.4 ms |
| `SD5` | 32 | 16 | 32768 → 32768 | 363 → 329 ms | 361 → 325 ms | classical → classical | 1.5 ms |
| `SD4(x)·SD4(x+1)·SD4(x+2)` | 48 | 24 | **159028 → 145499** | 1490 → 1238 ms | **1503 → 3286 ms** | **classical → lattice** | 4.0 ms |
| `SD5(x)·SD5(x+1)` | 64 | 32 | 262144 → 206368 | 5071 → 4039 ms | 17924 → 15816 ms | lattice → lattice | 8.5 ms |

All outputs verified correct in every cell (product reconstructs, expected factor counts; FLINT agrees on the factorizations).

**Controls (rows 1–6):** identical candidate counts and identical code path before/after — every level fits the budget, so `levelAwareSubsetBudget` returns it unchanged. The ~5–10% timing deltas are run-to-run noise.

**The target case (SD5 pair):** the hopeless burn stops at the `∑ C(31,≤5) = 206368` boundary instead of 262144: classical decline 5.07 → 4.04 s, hybrid 17.9 → 15.8 s. The win grows with `r` (the cap is 41728 at r = 64, 1025 at r = 1025); after https://github.com/kim-em/hex-dev/pull/8536 "perf: prune classical recombination candidates before trial division" the per-candidate cost is low enough that at r = 32 it is a ~20% decline saving.

**One honest regression the issue's design didn't anticipate (SD4 triple):** on `main`, classical *answers* this input — the winning size-7 subset sits 13529 candidates into the level that doesn't fit (159028 of 262144 tried, then the peeled quotients complete). The level boundary is `∑ C(23,≤6) = 145499`, so this PR declines there and the lattice tier answers instead: hybrid 1.50 → 3.29 s (2.2×), correct either way. This is the inherent flip side of "a partial level certifies nothing": a partial level can still *succeed*, and the boundary forfeits that. Notes: (a) https://github.com/kim-em/hex-dev/issues/8531 "test: merge-required conformance case that only the lattice tier can answer" predicted classical exhausts mid-size-7 on this family; measured on `main` it instead answers within budget (the winning size-7 subset arrives 13529 candidates into the level). Post-this-PR the prediction becomes true, making the SD4 triple a cheap (3.3 s) lattice-dispatch conformance witness, vs 19 s for the SD5 pair; (b) the forfeited-success window is exactly "budget-completable searches whose success lies in the first non-completable level" — nonempty but narrow, and such inputs still get a proven-correct (lattice) answer. The status quo and this rule differ in exactly one thing — the "remainder burn" (the `budget − boundary` candidates the flat budget spends partway into the first non-completable level): paying it keeps the occasional success in that prefix (this family) at the cost of wasting it on every hopeless decline (up to 99.6% of the budget at r = 1025); dropping it is this PR. There is no rule that keeps the prefix successes without paying the prefix. Happy to revert to the flat budget if this trade reads wrong.

Isabelle comparator (`bz_isabelle`, the second oracle) is building locally (AFP session heap); I'll follow up with its per-family times when it finishes.

kim-em:
Follow-up: the second oracle. `bz_isabelle` (the AFP `Berlekamp_Zassenhaus` verified implementation, Haskell extraction, driven through the persistent comparator protocol; fresh local build of the Isabelle2025-2 session heap) on the same eight families, same machine:

| family | deg | FLINT | Isabelle (verified) | hex hybrid (this PR) |
|---|---|---|---|---|
| `Φ₁₅` | 8 | 0.03 ms | 0.45 ms | 1.0 ms |
| `SD3` | 8 | 0.07 ms | 0.69 ms | 1.6 ms |
| `(x-1)…(x-16)` | 16 | 0.17 ms | 1.4 ms | 23.5 ms |
| `SD4` | 16 | 0.30 ms | 1.5 ms | 17.1 ms |
| `SD4(x)·SD4(x+1)` | 32 | 1.4 ms | 13 ms | 256 ms |
| `SD5` | 32 | 1.5 ms | 32 ms | 325 ms |
| `SD4(x)·SD4(x+1)·SD4(x+2)` | 48 | 4.0 ms | 355 ms | 3286 ms |
| `SD5(x)·SD5(x+1)` | 64 | 8.5 ms | **> 120 s (killed)** | **15816 ms** |

All completed cases agree with hex's factorizations (and FLINT's).

Two observations. First, on the classical-tier range (rows 1–7) hex is a constant factor behind the verified reference — the SPEC's stated parity target, unchanged by this PR. Second, the SD5 pair is a genuine crossover: the AFP implementation has no lattice tier, so its recombination grinds the same exponential wall whose burn this PR truncates, and it blew a 120 s cap where the hybrid's lattice arm answers (correctly) in 15.8 s. The one input class where the hybrid beats the verified reference outright is exactly the class this PR's decline boundary routes faster.

PR #8548, opened 2026-07-03
doc: review and revise the six planned-library SPECs in SPEC/Libraries

kim-em:
This PR review and revise the six planned-library SPECs (hex-roots, hex-resultant, hex-number-field, and their Mathlib companions), plus README.md and libraries.yml for consistency. The review verified every Mathlib citation against the pinned revision, every hex-dev API claim against the tree, and Dyadic/decide behaviour against the v4.32.0-rc1 toolchain; the mathematical fixes include a transposed resultant in commonField, the degree-1 case of toQAdjoin, AdjoinRoot taken over Q rather than Z, and corrected complexity contracts and references. Four sections are redesigns rather than corrections: the Pellet witnesses now compare per-term dyadic bounds (the squared form was not decidable), clusters are edge-connected components of grid squares with a structural termination measure and a pairwise-disjointness condition on output discs, SimpleRoot is a Quot over disc intersection of separation-refined isolations, and irreducibility is an executable test ladder (canonical form, rational roots, mod-p Rabin, full factor) so the Mathlib-free layer can construct its own proofs, with decide viable at fixture sizes. The user-facing AlgebraicNumber arithmetic is total, pinning the never-hit failure branch to 0 through an explicit panicWith, with staged soundness / bound-sufficiency theorems in the companion; Hex.NumberField is renamed Hex.QAdjoin to avoid the Mathlib class; and the Rouche-on-circles plan now cites the existing circle-integral lemmas (integral_sub_inv_of_mem_ball, DiffContOnCl.circleIntegral_eq_zero) and deliberately avoids MeromorphicOn. All prose follows SPEC/writing-style.md.

🤖 Prepared with Claude Code

### DIFF
--- a/SPEC/Libraries/README.md
+++ b/SPEC/Libraries/README.md
@@ -1,103 +1,104 @@
 # Libraries
 
-- **hex-arith** — extended GCD, Barrett/Montgomery reduction, binomial coefficients, Fermat's little theorem
-- **hex-poly** — dense `Array`-backed polynomial representation
-- **hex-matrix** — dense matrices, matrix/vector arithmetic, elementary row and column operations, submatrix slicing, the Gram matrix
-- **hex-row-reduce** — row reduction (RREF), rank, span, nullspace
-- **hex-determinant** — the Leibniz determinant and its cofactor/Cauchy-Binet/Plücker theory
-- **hex-bareiss** — the fraction-free Bareiss determinant algorithm
-- **hex-gram-schmidt** — Gram-Schmidt orthogonalization, GS coefficients, Gram determinants, update formulas under row operations
-- **hex-mod-arith** — `ZMod64 p`: `UInt64`-backed arithmetic in `Z/pZ`
-- **hex-poly-fp** — polynomials over `F_p`, Frobenius map, square-free decomposition, lazy reduction for small p
-- **hex-gf2** — packed bitwise polynomials over `F_2` (XOR + CLMUL), `GF(2^n)` elements
-- **hex-poly-z** — polynomials over `Z`, content/primitive part, Mignotte bound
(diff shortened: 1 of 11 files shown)


kim-em:
Codex second-opinion review (full transcript in the session log; reviewer progress note committed as `progress/20260703T045846Z_pr8548-spec-review.md`) found three issues, all addressed in 3f600aa5:

1. **Blocker (agreed): fixed `sepSlack` was unsound without Graeffe.** Near a simple root the higher Taylor coefficients collect contributions from all `n − 1` remote roots (`(1 + ρ/d)^{n−1} − 1` bounds the bad terms), so the Pellet witness needs an isolation ratio of order `deg p`, not a constant. `separationDepth` now includes a `ceilLog2 (max 2 (deg p))` term, the no-Graeffe difference is stated honestly, and hex-roots-mathlib's Pellet-converse item is restated with the linear-in-degree ratio.
2. **Significant (agreed): `commonField?` change of basis was underspecified.** Step 5 is now an executable construction: compute `gcd(β.p(y), α.p(γ − c·y))` in `(QAdjoin r γ)[y]` by HexPoly's Euclidean algorithm over the field; for a non-degenerate shift the gcd is linear, `y − B`, giving `βIn := B` and `αIn := γ − c·βIn`. A degree-≥ 2 gcd is one more degenerate-shift detector feeding the existing retry loop. The matrix-library dependency remains (used by `toAlgebraicNumber`), but `commonField?` no longer needs it.
3. **Minor (agreed): "never re-boxed" was misleading.** The `DyadicRootCluster` docstring now separates refinement geometry (the component squares, which subdivision operates on) from the certificate (attached to the circumscribed disc of `encSquare`).

🤖 Prepared with Claude Code

Issue #8618, opened 2026-07-05
perf: give DensePoly eval/scale/shift/compose kernel-facing specs on toList with @[csimp] Array impls


kim-em:
https://github.com/kim-em/hex-dev/pull/8615 introduced `DensePoly.toList` as the noncomputable spec-level coefficient list view and moved the `mul` specification onto it, with the runtime `Array` loop behind a proved `@[csimp]` equality (design principle 11 in `SPEC/design-principles.md`).

The executable bodies of `DensePoly.eval`, `scale`, `shift`, and `compose` (`HexPoly/Operations.lean`) still read coefficients via the explicit `toArray.toList` round-trip at runtime, and they double as their own kernel-facing definitions. Give each the same split as `mul`:

- a `noncomputable` specification stated over `DensePoly.toList` (cheap kernel reduction, no per-access `Array` list traversal),
- an `Array`-walking (or fold-based, allocation-free) `*Impl` runtime body registered by a proved `@[csimp]` equality,
- characterising lemmas (`coeff_scale`, `coeff_shift`, `eval_*`, `compose_*`) staying on the public names with unchanged statements.

Runtime performance comes first: the impl bodies should stop round-tripping through lists where that is free to do (e.g. a direct `Array` Horner fold for `eval`), and the bench verify checksums must stay green. This also removes the last executable `toArray.toList` uses inside `HexPoly/Operations.lean` itself; cold-path serializers in conformance/bench drivers keep the explicit spelling by design.

🤖 Prepared with Claude Code

kim-em:
Closed by the four-package spec/impl split plan, all merged:

- https://github.com/kim-em/hex-dev/pull/8626 — add/sub/neg/derivative (zipPad/derivList kernel specs, Array.ofFn/Array.map impls)
- https://github.com/kim-em/hex-dev/pull/8627 — eval/compose/composeModMonic (cons-walk specs, downward Array.foldr impls), Quotient.eval spec layer noncomputable on toList
- https://github.com/kim-em/hex-dev/pull/8629 — mulImpl accumulator read via Array.getD
- https://github.com/kim-em/hex-dev/pull/8631 — contentNat/primitivePart (Array.foldl/Array.map impls)

Every DensePoly executable-layer list round-trip named here is gone; the eval/scale/shift/compose scope of this issue is covered by the first two PRs (scale/shift were already Array-path via ofCoeffs after the toList work; their coefficient reads go through the same characterisations).

[...]

PR #8626, opened 2026-07-06
perf: kernel-cheap specs and Array-native @[csimp] impls for DensePoly add/sub/neg/derivative

kim-em:
This PR gives the coefficientwise `DensePoly` operations the kernel-facing-spec / `@[csimp]`-runtime-impl split of design principle 11 (PR 1 of the four-package plan). `add`/`sub` become `zipPad` walks over the two `toList` views: one cons step per coefficient under kernel reduction (these sit on the cross-check `decide` path via `curr + quot * f`), with literal `x + Zero.zero` overhang terms so the `@[csimp]` equalities need no algebraic laws on `R`; the runtime impls are single `Array.ofFn` allocations replacing the `List.range`-map round-trip. `neg` keeps the `0 - p` spec and gains a one-pass `Array.map` impl instead of paying `sub`'s full fold against the zero polynomial. `derivative` becomes a `derivList` cons walk with an `Array.ofFn` impl. All characterising lemma statements (`coeff_add`/`sub`/`neg`/`derivative`, `eval_add`/`eval_sub`, `compose_C`) are unchanged; the six external `derivative` body-unfold sites migrate to the characterisations, supported by the new `size_derivative_le`.

One trap found and fixed during development: the first `zipPad` formulation recursed on different arguments per clause, so Lean compiled it by well-founded recursion, which is opaque to kernel reduction and broke every downstream `decide`; it now recurses structurally on the first list.

Full `lake build` (4142 jobs) plus `HexConformance` are green (`HexGFq.CrossCheck` drops 32 s → 28 s); bench verify checksums pass for `hexpoly`/`hexpolyz`/`hexpolyfp`/`hexmodarith`; the spec symbols are absent from the generated C with the instance closures on the impls. Codex review: no blocking findings.

Part of https://github.com/kim-em/hex-dev/issues/8618.

🤖 Prepared with Claude Code

### DIFF
--- a/HexPoly/Euclid/MulRing.lean
+++ b/HexPoly/Euclid/MulRing.lean
@@ -884,19 +884,8 @@ private theorem rat_weighted_diagonal_fold
     rw [hidx])
 
 private theorem rat_coeff_derivative_generic (p : DensePoly Rat) (n : Nat) :
-    (derivative p).coeff n = ((n + 1 : Nat) : Rat) * p.coeff (n + 1) := by
-  unfold derivative
-  rw [coeff_ofCoeffs_list]
-  change
-    ((List.range (p.size - 1)).map
-        (fun i => ((i + 1 : Nat) : Rat) * p.coeff (i + 1))).getD n 0 =
-      ((n + 1 : Nat) : Rat) * p.coeff (n + 1)
-  by_cases hn : n < p.size - 1
-  · simp [hn, List.getD]
-  · have hp : p.size ≤ n + 1 := by omega
-    have hcoeff : p.coeff (n + 1) = 0 :=
-      coeff_eq_zero_of_size_le p hp
-    simp [hn, List.getD, hcoeff]
+    (derivative p).coeff n = ((n + 1 : Nat) : Rat) * p.coeff (n + 1) :=
+  coeff_derivative p n (Rat.mul_zero _)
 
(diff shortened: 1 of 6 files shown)


kim-em:
Correction to this PR's description, after Kim challenged the claim that well-founded recursion is "opaque to kernel reduction": that is overstated. Verified by experiment on v4.32.0-rc1: a well-founded definition with an explicit Nat measure (`termination_by xs ys => xs.length + ys.length`) reduces fine in the kernel (`decide +kernel` succeeds), and plain `decide` fails only because well-founded definitions are `@[irreducible]` by default — `unseal foo in` unblocks it (`@[semireducible]` is explicitly warned as ineffective for WF definitions). What genuinely does not reduce, even with `+kernel`, is the **auto-derived lexicographic measure** this PR's original four-clause `zipPad` received (`invImage ... Prod.instWellFoundedRelation`), and that is what broke the downstream `decide`s.

The structural rewrite stays the right fix for a kernel-facing spec (no `unseal`/`+kernel` churn at the ~50 cross-check `decide` sites, ordinary defeq, cheaper reduction), but the mechanism in the description should read: the auto-derived lex measure does not kernel-reduce, and WF definitions are irreducible to the elaborator by default; WF over an explicit Nat measure is kernel-reducible.

Issue #8625, opened 2026-07-05
spike(bz): investigate recursive per-remainder re-lift for sub-floor classical irreducibility certification


kim-em:
**Background.** #8620 (https://github.com/kim-em/hex-dev/issues/8620) proposed a per-remainder Mignotte backstop for the classical tier and was closed as unsound: classical-tier irreducibility flows through *coverage* (`RecoveredSmartSearch.covers_of_bound`, `HexBerlekampZassenhausMathlib/SmartSearchCoverage.lean:1396`), which requires a **complete lifted-factor partition** `LiftedFactorSubsetPartition core d J target`. The only initial-partition producer (`liftedFactorSubsetPartition_of_toMonicPrimeData_complete`, `HexBerlekampZassenhausMathlib/IntReductionMod.lean:49`) is gated at `2 * defaultFactorCoeffBound (toMonic core).monic < p ^ …`, and the per-remainder recursion **inherits** that partition at the *same* `LiftData` via `liftedFactorSubsetPartition_transport` (`HexBerlekampZassenhausMathlib/M1Recovery.lean:852`, signature `… core d J target → … core d (J \ S) quotient`). So no returned factor can be certified irreducible below the monic-core floor `2 * mignotte((toMonic core).monic)` within the transported-partition architecture.

**The only sound route to a sub-floor win.** Certify each unsplit remainder against a **fresh** complete partition of its own: independently re-Hensel-lift the remainder's own monic transform `(toMonic remainder).monic` to precision `2 * defaultFactorCoeffBound (toMonic remainder).monic`, select a prime for the remainder, and run coverage there — instead of inheriting the core's global partition. This lets a remainder be certified at `mignotte((toMonic remainder).monic)`, which is below the core's floor when the remainder is a proper factor of a larger core. It is a genuinely different algorithm (a new per-factor lift + partition + coverage stack under well-founded / fuel recursion), NOT the `latticeCoreLoop` mirror #8540/#8620 assumed.

**This is an investigation, not a build order — measure before proving.** The #8620 measurement showed the *cheap* core-local-bound win is marginal on the corpus; the sub-floor win here is also input-dependent and its magnitude is unclear (the total cost is a sum of per-remainder floor-lifts vs one core-floor lift, and precision is logarithmic in the bound). Quantify the win before committing to the ~5-file proof remodel.

**Deliverables.**

[...]

Issue #8621, opened 2026-07-05
perf(bz): balanced product-tree multifactor Hensel lift


kim-em:
**Current state.** The multifactor Hensel lift is a sequential O(n·r) peel: `multifactorLiftQuadraticList` (`HexHensel/QuadraticMultifactor.lean:91`) lifts the head factor `g` against the product of ALL remaining factors and recurses on the complement, so each step multiplies the whole tail. `henselLiftData` (`HexBerlekampZassenhaus/ChoosePrimeData.lean:842`) calls it with a single fixed precision. A working prototype `balancedLift` (`bench/HexBench/ClassicalSpike.lean:101`) reshapes this into an O(log r)-depth / O(n log r)-total product tree using the SAME proven primitives (`ZPoly.normalizedXGCD`, `ZPoly.henselLiftQuadratic`, `FpPoly.liftToZ`, `Array.polyProduct`), with measured wins. Split off from #8540 (https://github.com/kim-em/hex-dev/issues/8540) as the balanced-lift half; independent of the incremental-precision sibling #8620 (https://github.com/kim-em/hex-dev/issues/8620).

**Deliverables.**

1. Reshape `multifactorLiftQuadratic` / `henselLiftData` into a balanced product-tree lift (split the factor list in half, lift each half's product against the target, recurse into each half). Output-preserving: the balanced tree flattens L then R, so the lifted-factor array stays in the original order `g1..gn` with the same reduced values (Hensel lift is unique mod `p^k`).

2. Remodel the Mathlib proof to match the new shape. `Hex.ZPoly.QuadraticMultifactorLiftInvariant` and the chain `multifactorLiftQuadratic_spec` / `_each_congr_mod_base` / `_size_eq_input` / `QuadraticMultifactorLiftInvariant_of_choosePrimeData` are proved by structural induction on the sequential list recursion and are threaded through `HexBerlekampZassenhausMathlib/SubsetCoprimality.lean`, `ForwardHenselTransport.lean`, `ToMonicUniqueness.lean`, `LatticeTier.lean`, and `HenselFactorProps.lean`. A balanced tree needs well-founded recursion, so either reprove the invariant for the new recursion or add an output-equality lemma to the sequential version (via Hensel uniqueness) and keep the existing invariant lemmas keyed to that.

[...]

PR #8628, opened 2026-07-06
bench: measure recursive per-remainder re-lift for sub-floor classical certification

kim-em:
This PR adds the gating measurement for https://github.com/kim-em/hex-dev/issues/8625 (deliverable 1): an executable, unproven prototype of the recursive per-remainder re-lift for sub-floor classical irreducibility certification, plus the findings report. `bench/HexBench/RecursiveReliftSpike.lean` (`lake exe hex_recursive_relift_spike`, following the `ClassicalSpike.lean` precedent: bench-only, not in CI's exe target list, not in the release-sync manifest) prototypes the recursion in a fresh-prime variant (per-remainder `choosePrimeData?`) and a same-prime variant (reuse the parent prime and each remainder's own local factors, with a sub-floor scan-size cap), sharing one recombination scan across all arms including the today-baseline (d-1 trailing filter plus a Mignotte-bounded exact division). Every case re-multiplies the certified factors against the core and checks degree signatures.

Findings (`reports/bz-recursive-relift-findings.md`): the sub-floor lift-work win is real and large on composite cores (4-16x less degree-weighted lift work on the split degree/height families, 9x on `adv/high_multiplicity`, zero on irreducible cores); a naive recursion loses it back 2.5-3.8x end-to-end to failed sub-floor recombination scans and per-node prime selection (18.9 ms `choosePrimeData?` at deg 24, measured); the same-prime variant with sub-floor scans capped at size-2 subsets turns that into a robust 1.8-2.1x end-to-end win on all split families and 1.6x on `adv/high_multiplicity`, at a bounded 1.17-1.25x overhead on the irreducible adversarial inputs (Phi15, SD3, SD4) and 1.32x on `adv/mignotte_swell` (whose sub-floor split needs the size-2 cap; singletons-only loses 2x there). A production arm (`classicalCoreFactorsWithBound`) added after a Codex second-opinion review matches the shared-scan baseline within noise on every family, validating the baseline normalization. The report recommends sequencing deliverable 2 (the proven per-remainder partition/coverage stack) after https://github.com/kim-em/hex-dev/issues/8621 (balanced product-tree lift), which attacks the same lift cost without a certification re-key, and re-measuring the residual win on that baseline before committing to the ~5-file proof remodel.

No production code changes: the only non-bench edits are the `lean_exe` registration in `lakefile.lean` and the report/progress files, so FLINT conformance and the merge-gating `*Mathlib` layers are untouched.

🤖 Prepared with Claude Code

### DIFF
--- a/bench/HexBench/RecursiveReliftSpike.lean
+++ b/bench/HexBench/RecursiveReliftSpike.lean
@@ -1,0 +1,501 @@
+/-
+Copyright (c) 2026 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
+-/
+
+import HexBerlekampZassenhaus
+
+/-!
+Measurement spike (UNPROVEN) for issue #8625: recursive per-remainder re-lift
+for sub-floor classical irreducibility certification.
+
+Today the classical tier runs ONE Hensel lift of the whole core at
+`precisionForCoeffBound (exhaustiveLiftBound core B) p` and certifies
+irreducibility of unsplit remainders through coverage of that single lift,
+whose partition completeness is gated at the monic-core Mignotte floor
+`2 * defaultFactorCoeffBound (toMonic core).monic < p^k` (see #8620). The only
+sound route below that floor is to give each unsplit remainder a FRESH
+lift/partition/coverage stack of its own, at the remainder's OWN monic floor.
+
+This file prototypes that recursion executably (unproven, like
+`classicalFactorIntK` in `ClassicalSpike.lean`) in two variants:
+
+* **fresh-prime**: each remainder gets its own `choosePrimeData?` (prime walk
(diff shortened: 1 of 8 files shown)


kim-em:
Post-merge-window note: the findings report's prime-selection section says the remaining reduction-bound cost is "consistent with https://github.com/kim-em/hex-dev/issues/8642 (packed modByMonic is the remaining lever there)". #8642 has since been closed with the opposite measured finding, recorded in https://github.com/kim-em/hex-dev/issues/8646 ("doc(polyfp): record that packed modByMonic is not the FpPoly lever"): wiring the packed kernel moved end-to-end numbers ~0%. The report's profiling data (prime selection still 13.8 ms at deg 24, single isGoodPrime 299 us post-#8633/#8638/#8641) stands; only the pointer to the presumed fix is superseded. Leaving the text as-is to avoid a CI cycle for one sentence — this comment is the correction.

🤖 Prepared with Claude Code

Issue #8630, opened 2026-07-06
perf(bz): prime selection is boxing-bound; packed FpPoly gcd/nullspace kernels


kim-em:
**Measurement.** Profiling for https://github.com/kim-em/hex-dev/issues/8625 (see `RELIFT_PROFILE=prime` in `bench/HexBench/RecursiveReliftSpike.lean`, https://github.com/kim-em/hex-dev/pull/8628, and the "Prime selection breakdown" section of `reports/bz-recursive-relift-findings.md`) shows `choosePrimeData?` costs, on shifted split families (us/call): deg 8: 818, deg 12: 2113, deg 16: 5372, deg 20: 10414, deg 24: 18666 (roughly cubic in degree); `Phi15`: 484; SD4 (deg 16, r=16): 8417. At deg 24 that is **30% of the entire classical-tier runtime** (18.7 of 61.4 ms) and it is paid once per input by every tier before any lifting starts. A **single `isGoodPrime` check costs 424 us** at deg 24 — one `ZPoly.modP` plus one `DensePoly.gcd` over `ZMod64 p`, which should be a few hundred word operations, i.e. ~1 us. The ~9 failing candidate primes account for ~4 ms of the walk; the Berlekamp factorization (`fixedSpaceKernel` Frobenius matrix + nullspace + the `c = 0..p-1` sweep of `splitFactorAt f witness c = DensePoly.gcd f (witness - FpPoly.C c)`, `HexBerlekamp/Factor.lean:59`) accounts for the remaining ~15 ms.

**Cause.** perf attribution: `lean_apply_1/2` closure dispatch ~12%, allocator traffic ~15%, `lean_is_scalar`/`lean_unbox`/`lean_scalar_to_int64` shims, the generic `nullspaceMatrix`, and `lean_int_extgcd_fallback` + gmp on the modular-inverse path. The whole prime-selection path runs boxed, typeclass-generic arithmetic: `DensePoly.gcd` (`HexPoly/Euclid/DivGcd.lean:1147`) is generic over `[One R] [Add R] [Sub R] [Mul R] [Div R]`, so every coefficient operation is a closure call on a boxed `ZMod64` value, ~0.5-1 us each where a packed word loop costs nanoseconds.

[...]

PR #8633, opened 2026-07-06
perf(modarith): drop dead word-modulus branch from ZMod64 mul/pow externs

kim-em:
This PR removes a per-multiply heap allocation from the core `ZMod64` modular-arithmetic externs, speeding up all `ZMod64` modular multiplication in the codebase.

`lean_hex_zmod64_mul` allocated a bignum (`lean_uint64_to_nat(UINT64_MAX)`) on every call, purely to test whether the modulus was `2^64` (the "word modulus" case, where `% 2^64` cannot be computed in 64 bits). Since https://github.com/kim-em/hex-dev/pull/8612 tightened `ZMod64.Bounds` to `p < 2^64` strictly (`Bounds.pLtR`), that case is impossible — the branch is dead code the Lean-side cleanup left behind in the C extern. This reads the modulus directly with `lean_uint64_of_nat` (no width check, no allocation) and drops the same dead word-modulus machinery from `lean_hex_zmod64_pow`.

Semantics are unchanged for every `Bounds p`, so outputs are byte-identical (the extern's Lean reference `ZMod64.mul` / `ZMod64.pow` and all proofs are untouched). On `hexberlekamp_bench` (compiled binary, measured A/B):

| path | before | after | speedup |
|---|---|---|---|
| gcd (`runBerlekampFactorChecksum`) deg 64 | 1.02 ms | 511 µs | 2.0× |
| gcd deg 256 | 12.8 ms | 5.6 ms | 2.3× |
| Frobenius (`runRabinTestChecksum`) deg 24 | 8.32 ms | 4.53 ms | 1.8× |

This is the runtime win behind https://github.com/kim-em/hex-dev/issues/8630. The issue's proposed packed `Array UInt64` kernels turned out not to be the fix: benchmarking showed the packed gcd was a 1.6-2.7× regression (it re-crossed the boxed boundary every Euclidean step) and packed `modByMonic` was neutral. The actual cost was this allocation in the shared modular multiply, which the packed kernels also paid. See the PR discussion for the full A/B evidence.

Follow-up: Barrett reduction for `p < 2^32` (division-free modmul; the repo already has `HexArith/Barrett` + `HexModArith/HotLoop.lean` gated on `p < 2^32`, currently unused by `ZMod64.mul`).

🤖 Prepared with Claude Code

### DIFF
--- a/HexModArith/ffi/zmod64_mul.c
+++ b/HexModArith/ffi/zmod64_mul.c
@@ -17,48 +17,33 @@ extern void __gmpz_clear(mpz_t);
 extern void* __gmpz_export(void*, size_t*, int, size_t, int, size_t, const mpz_t);
 extern void lean_extract_mpz_value(lean_object*, mpz_t);
 
+/* `ZMod64.Bounds p` guarantees `0 < p < 2^64` (`Bounds.pLtR`), so the modulus
+   always fits in a `uint64_t` and the former `p = 2^64` "word modulus" case is
+   impossible. Reading the modulus with `lean_uint64_of_nat` therefore needs no
+   width check and, in particular, avoids the per-call bignum allocation
+   (`lean_uint64_to_nat(UINT64_MAX)`) the old width check paid on every call. */
+
 LEAN_EXPORT uint64_t lean_hex_zmod64_mul(b_lean_obj_arg p, uint64_t a, uint64_t b) {
-    lean_object* max_word_nat = lean_uint64_to_nat(UINT64_MAX);
+    uint64_t modulus = lean_uint64_of_nat(p);
     __uint128_t product = (__uint128_t)a * (__uint128_t)b;
-    uint64_t result;
-
-    if (lean_nat_lt(max_word_nat, p)) {
-        /* The only bounded modulus above UINT64_MAX is 2^64 itself. */
-        result = (uint64_t)product;
-    } else {
-        uint64_t modulus = lean_uint64_of_nat(p);
-        result = (uint64_t)(product % modulus);
-    }
-
-    lean_dec(max_word_nat);
-    return result;
-}
-
-static bool lean_hex_modulus_is_word(lean_object* p) {
-    lean_object* max_word_nat = lean_uint64_to_nat(UINT64_MAX);
-    bool is_word = lean_nat_lt(max_word_nat, p);
-    lean_dec(max_word_nat);
-    return is_word;
-}
-
-static uint64_t lean_hex_mul_mod_word(uint64_t a, uint64_t b, uint64_t modulus, bool word_modulus) {
-    __uint128_t product = (__uint128_t)a * (__uint128_t)b;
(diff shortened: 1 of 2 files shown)


kim-em:
## Benchmarking revealed this is not a win — converting to draft

I benchmarked `hexberlekamp_bench` (A/B: the two `@[csimp]` swaps ON vs the same build with both swaps disabled, same session, `commit=e2ea0a4`). The correctness proofs hold (outputs byte-identical), but the performance goal is not met:

| Benchmark | swaps OFF | swaps ON | result |
|---|---|---|---|
| `runRabinTestChecksum8` (D1 modByMonic) | 420 µs | 423 µs | neutral |
| `runRabinTestChecksum16` | 2.868 ms | 2.872 ms | neutral |
| `runRabinTestChecksum24` | 8.318 ms | 8.29 ms | neutral |
| `runBerlekampFactorChecksum` @16 (D2 gcd) | 120 µs | 195 µs | **1.6× slower** |
| `runBerlekampFactorChecksum` @48 | 627 µs | 1283 µs | **2.0× slower** |
| `runBerlekampFactorChecksum` @256 | ~12.8 ms | ~34 ms | **2.7× slower** |

**D1 (packed `modByMonic`) is performance-neutral** on the Rabin/Frobenius path — no measurable win.

**D2 (packed `gcd`) is a regression that widens with degree.** Root cause: `gcdAuxPacked` re-packs both operands with `toWords` (an allocating `Array.map` over boxed `ZMod64` values) and unpacks+trims with `ofCoeffs (ofWords …)` on *every* Euclidean iteration. That adds two extra boxed passes per step on top of the packed inner loop — the opposite of de-boxing. The generic `DensePoly.gcd` already runs the `@[csimp]` ceiling-tracking `divModArrayAuxImpl` in place on the boxed arrays and never crosses the boxed boundary mid-loop, so it wins.

A packed gcd only helps if the **entire** Euclidean loop stays in word-space: convert to `Array UInt64` once, run degree-tracking + divmod + leading-coeff inverse + trimming all in words, convert back once. That is a materially larger kernel and correspondence proof than what landed here (essentially re-deriving the whole gcd/divmod/degree stack over words), so it belongs in its own focused effort rather than this swap.

Converting to draft. Options: (a) drop D2 and keep only D1 (but D1 is neutral, so it is added kernel + proof surface for no measured benefit), (b) close this and redo D2 as a stay-in-word-space kernel, (c) close entirely. Leaning toward (c)/(b).

kim-em:
Reshaped: replaced the packed-kernel commits with the single-commit extern fix (force-pushed). Benchmarking showed the packed kernels regressed/were neutral; the real cost was the per-multiply bignum allocation in `lean_hex_zmod64_mul`, a dead word-modulus check obsolete since #8612 made `Bounds` require `p < 2^64`. ~1.6-2.3x on the gcd path and ~1.9x on Frobenius, output byte-identical.

PR #8649, opened 2026-07-07
doc(hex-matrix): specify Strassen-Winograd multiplication

### LINKED ISSUE #8617: write a spec for Strassen-Winograd matrix multiplication
Let's make a performant Strassen-Winograd, and ideally just use it by default in matrix multiplication at runtime.

* We'll need benchmarks
* There should be a (not used by default) backend that lets you plug in a custom scheme for small sizes
* Have to think about how to implement it without unnecessary copying / extracting sub-blocks just to multiply and discard them
* Read the literature some more!

kim-em:
This PR specifies Strassen-Winograd matrix multiplication for `hex-matrix` as a section of `HexMatrix/SPEC/hex-matrix.md`, addressing all four requirements in the issue: benchmarks, a pluggable-but-not-default backend for small sizes, a copy-avoidance plan, and a literature-informed design.

The section specifies the memory-efficient Winograd schedule (seven multiplications, fifteen additions; Boyer-Dumas-Pernet-Zhou, ISSAC 2009); a data-only `StrassenConfig` carrying a cutoff and a customizable base kernel, with correctness factored through a separate `StrassenConfig.Valid` predicate so the default naive base kernel and any hand-tuned small-matrix scheme share one recursion; padded-even recursion that handles odd and rectangular dimensions without breaking the balanced block split; block views inside the recursion with materialization only at the base case, and a flat backing representation noted as the prerequisite for a fully copy-free schedule; correctness stated as `mulStrassen_eq_mul` over `Lean.Grind.Ring` and swapped into compiled code by `@[csimp]`, with no `axiom` and no `native_decide`; and the benchmark (measured crossover cutoff, `n^{log₂ 7}` cost model) and conformance plan.

It records a premise correction: Winograd's schedule subtracts blocks, so `mulStrassen` needs `[Sub R]` and cannot be a type-preserving `@[csimp]` replacement of the semiring-generic `mul`. It is instead a ring-level entry point covering every coefficient type the project multiplies (`Int`, `Rat`, `ZMod64`, `Fp`), while the generic `*` keeps naive `mul` as its universal fallback. The `mul` docstring in `HexMatrix/Basic.lean` now points at the section.

The draft was reviewed with `/second-opinion` (Codex), which confirmed the schedule is algebraically correct and surfaced defects since fixed: the odd-dimension half-split did not type-check (replaced with padded-even recursion), the correctness theorem needed a lawful ring class, and the config coupled data with a proof field.

No Lean implementation is included: per design principle 7 (no scaffolding definitions) and the issue title, the deliverable is the specification. `HexMatrix` builds green; only docstring text changed in source.

Closes #8617.

🤖 Prepared with Claude Code


kim-em:
Follow-up filed: #8652 (flat `Vector R (n*m)` backing for `Matrix`). The Strassen block views are copy-free only on a flat backing, and the crossover measured on the current row-of-rows backing is confounded by quadrant-copy overhead (see the representation-confounder note in the Benchmarks section). #8652 is independent of this SPEC and is not a blocker: Strassen lands on the interim representation, and #8652 measures the switch as a before/after overlay on the same multiply bench.

Issue #8535, opened 2026-07-02
perf: cap classical size-level subset materialization by the running budget


kim-em:
**Parent:** #8529 (deliverable 3; deliverables 1–2 landed as the prefilter PR). #8417 performance side.

## Current state

The classical tier's size loop (`scaledRecombinationSmartSizeLoop`, `HexBerlekampZassenhaus/Basic.lean`) materializes each size level eagerly: `(subsetsOfSizeWithComplement tail d).map (head :: ·.1, ·.2)` builds **all** `C(r−1, d)` split pairs as a strict list before `scaledRecombinationSmartCandLoop` tests any of them. With the #8529 prefilter the per-candidate cost is now small, so this allocation is a larger *fraction* of the remaining budget-exhaustion cost, and for larger `r` a single size level can allocate far beyond the budget's reach (budget 262144 vs `C(31, 15) ≈ 3·10⁸` pairs at `r = 32` — only avoided today because the budget exhausts at `d ≈ 6`).

## What to do

Bound the materialization by the running budget instead of restructuring the loop into a stream:

1. Add `subsetsOfSizeWithComplementCapped : List α → Nat → Nat → List (List α × List α)` generating at most `cap` splits (fuse the cap through the two-branch append; the second branch's cap is `cap − (first branch length)`), with the spec lemma `subsetsOfSizeWithComplementCapped xs k cap = (subsetsOfSizeWithComplement xs k).take cap`.
2. In `scaledRecombinationSmartSizeLoop`, pass `budget` as the cap.
3. Prove behavior identity: the candidate loop consumes at most one split per budget unit, so `scaledRecombinationSmartCandLoop coreLc target modulus (splits.take budget) budget fuel = scaledRecombinationSmartCandLoop coreLc target modulus splits budget fuel` (induction on fuel/splits; every branch decrements the budget or returns).

## Proof-surface caution

[...]

PR #8675, opened 2026-07-08
perf(bz): budget-bounded subset materialization in the classical size loop

### LINKED ISSUE #8648: perf(bz): level-aware budget alignment in quotient sub-searches of the classical size loop
Narrow follow-up to https://github.com/kim-em/hex-dev/issues/8535 (perf: cap classical size-level subset materialization by the running budget), closed because https://github.com/kim-em/hex-dev/pull/8537 (level-aware classical decline boundary, `levelAwareSubsetBudget`) resolved its top-level concern: the top-level search now runs out of budget at the `budget = 0` check before entering a size level it cannot afford.

**Surviving cases**, per the scope note on #8535: eager size-level materialization in `scaledRecombinationSmartSizeLoop` (`HexBerlekampZassenhaus/Basic.lean`) can still build all `C(r-1, k-1)` splits of a level it cannot afford when the running budget is not aligned to a level boundary, which happens for

1. quotient sub-searches after a successful peel (`scaledRecombinationSmartAux` recursion, `HexBerlekampZassenhaus/Recombination.lean:373`), which inherit a mid-level remaining budget, and
2. callers passing a non-default budget.

**Deliverable (hypothesis, verify against the code first).** Re-align the budget to a level boundary on entry to a quotient sub-search (recompute `levelAwareSubsetBudget` from the remaining budget and the sub-search's `r`), or make the level materialization lazy so at most `budget + 1` subsets are built. Value-preserving for all default-budget top-level calls; if the decline boundary shifts on sub-searches, regenerate the trace baseline per the #8537 precedent.

**Verification.** FLINT conformance byte-identical (`bz_flint.py`, `bz_trace_gate.py`, trace baseline updated only if the decline boundary provably moves); classical-tier bench targets in `bench/HexBerlekampZassenhaus/Bench.lean` unchanged or improved on the recombination-heavy targets.

🤖 Prepared with Claude Code

kim-em:
This PR caps size-level subset materialization in `scaledRecombinationSmartSizeLoop` by the running candidate budget, so a quotient sub-search after a peel (or a caller passing a non-default budget) that inherits a mid-level budget no longer builds an entire `C(r-1, d)` size level it cannot afford only to try a handful of splits. These are the cases that survived the level-aware decline boundary of https://github.com/kim-em/hex-dev/pull/8537 (level-aware classical decline boundary, `levelAwareSubsetBudget`), as noted in the scope of https://github.com/kim-em/hex-dev/issues/8535 (perf: cap classical size-level subset materialization by the running budget), and close https://github.com/kim-em/hex-dev/issues/8648 (perf(bz): level-aware budget alignment in quotient sub-searches of the classical size loop).

The fix adds `subsetsOfSizeWithComplementTake`, a budget-bounded generator that produces at most `n` size-`k` splits in the same order as `subsetsOfSizeWithComplement` and never enumerates the level above `n` (a subtree asked for `0` more splits returns immediately), and feeds the size loop the running `budget`. It chooses lazy materialization over re-aligning the sub-search budget precisely because it is value-preserving: the candidate loop already consumes at most `budget` splits before its `budget = 0` guard fires, so restricting the split list to its length-`budget` prefix cannot change what the search tries. `scaledRecombinationSmartCandLoop_take` proves the candidate-loop result is unchanged (using the existing budget-monotonicity lemma so a sub-search cannot refund budget past the built prefix), `subsetsOfSizeWithComplementTake_eq` identifies the generator with `(subsetsOfSizeWithComplement xs k).take n`, and the two size-loop coverage proofs in `SmartSearchCoverage` realign the now-bounded hypothesis back to the full enumeration so the covering-subset argument is unchanged. The existing size-loop soundness proofs are generic over the split list and need no change.

Because nothing the search tries changes, the decline boundary does not move: the regenerated `bz.jsonl` conformance fixtures are byte-identical to the committed ones, `bz_trace_gate.py` passes against the existing baseline, and no trace baseline regeneration is required. The generator is empirically efficient (it returns the first 5 splits of `C(40, 20)` instantly and the full `C(12, 6) = 924` when the budget exceeds the level).

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekampZassenhaus/Recombination.lean
+++ b/HexBerlekampZassenhaus/Recombination.lean
@@ -260,6 +260,46 @@ def subsetsOfSizeWithComplement {α : Type} : List α → Nat → List (List α 
       (subsetsOfSizeWithComplement xs k).map (fun sc => (x :: sc.1, sc.2)) ++
       (subsetsOfSizeWithComplement xs (k + 1)).map (fun sc => (sc.1, x :: sc.2))
 
+/-- Budget-bounded prefix of `subsetsOfSizeWithComplement xs k`: it produces the
+first `n` size-`k` sublists (with complements), in the same order, and stops as
+soon as that prefix is full. It returns at most `n` splits
+(`subsetsOfSizeWithComplementTake_length_le`) and, crucially, does not build the
+whole `C(xs.length, k)` level before taking the prefix — a subtree asked for `0`
+more splits returns immediately (the `n = 0` arm), so the level above `n` is
+never enumerated. Equal to `(subsetsOfSizeWithComplement xs k).take n`
+(`subsetsOfSizeWithComplementTake_eq`).
+
+The size-ordered classical search feeds this the running candidate `budget`. The
+candidate loop consumes at most `budget` subsets before its `budget = 0` guard
+fires, so building only the first `budget` of a size level is value-preserving
+while keeping the whole level from being realised when the running budget is not
+aligned to the level boundary (a quotient sub-search after a peel, or a
+non-default caller budget) — the surviving cases from #8535/#8537. -/
+@[expose]
+def subsetsOfSizeWithComplementTake {α : Type} :
+    List α → Nat → Nat → List (List α × List α)
+  | _, _, 0 => []
+  | xs, 0, _ + 1 => [([], xs)]
+  | [], _ + 1, _ + 1 => []
+  | x :: xs, k + 1, n + 1 =>
+      let withHead :=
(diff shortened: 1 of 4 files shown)


kim-em:
Closing: the change is correct and provably value-preserving, but end-to-end benchmarking shows the over-materialisation it targets never fires through the shipping `factor` dispatcher (before ≈ after, 1.0× on every recombination-heavy input). Full rationale and the before/after numbers are on https://github.com/kim-em/hex-dev/issues/8648#issuecomment-4921997601.

Issue #8652, opened 2026-07-08
perf(hex-matrix): flat Vector R (n*m) backing for Matrix


kim-em:
This issue switches `Hex.Matrix R n m` from its row-of-rows `Vector (Vector R m) n` backing to a flat `Vector R (n*m)` backing, behind the unchanged opaque API, so quadrant sub-blocks become offset-and-stride windows and the dense multiply hot paths stop paying per-row allocation.

**Current state.** `Hex.Matrix R n m` wraps `Vector (Vector R m) n`. Design principle 10 and the `hex-matrix` SPEC already name a future flat `Vector R (n*m)` backing, and the type was deliberately encapsulated (opaque one-field structure; `ofFn`/`ofRows`/`getRow`/`rows`/`M[(i,j)]` API; `noncomputable` `M[i]` row access) so the switch stays invisible to consumers. It is stated as intent, not required or implemented.

**Deliverables.**
1. Settle the layout order, row-major `n*m` versus column-major `m*n`, and record the decision and rationale in `HexMatrix/SPEC/hex-matrix.md`. The deciding factor is which axis must be contiguous for the multiply hot paths: whether `mulImpl`'s transpose step survives, and how Strassen block strides fall out.
2. Update the `hex-matrix` SPEC to require the flat backing. This is an intentional representation change, called out per design-principles §"Scope of autonomous SPEC edits".
3. Migrate the `Matrix` internals behind the unchanged API. `getRow`, `col`, `transpose`, `principalSubmatrix`, the elementary operations, and `mulImpl` keep their signatures; only their bodies and the backing change. The `noncomputable` `M[i]` guard stays, so no compiled path materializes a row to read one entry.
4. Add a `BlockView` (backing matrix plus row and column offsets plus two strides) so quadrant sub-blocks are stride windows, which is what the copy-free Strassen schedule needs (see the "Avoiding sub-block copies" section of the `hex-matrix` SPEC).
5. Measure the switch as a before/after overlay on the existing multiply bench (`bench/HexMatrix/Bench.lean` `runSquareMulChecksum`, overlay mode per the plotter), reporting the effect on the naive baseline and, once it lands, on `mulStrassen`.

[...]

PR #8678, opened 2026-07-08
feat(hex-matrix): mulStrassen with StrassenConfig and mulStrassen_eq_mul

### LINKED ISSUE #8676: feat(hex-matrix): mulStrassen with StrassenConfig and mulStrassen_eq_mul
This issue implements the Strassen-Winograd multiplication entry point in `hex-matrix`: `StrassenConfig`, its validity predicate, the recursive `mulStrassen`, the default configuration, and the correctness theorem `mulStrassen_eq_mul`. This is wave 2 of the Strassen SPEC implementation (tracking: #8665); all three proof-infrastructure lemmas from wave 1 are on `main` (#8674 blocks, #8673 padding, #8672 Winograd identity). Read [HexMatrix/SPEC/hex-matrix.md § Strassen-Winograd multiplication](https://github.com/kim-em/hex-dev/blob/main/HexMatrix/SPEC/hex-matrix.md#strassen-winograd-multiplication) in full before starting; it is the authority for every signature below.

**Current state.** On `main`: `Matrix.fromBlocks` / `toBlocks₁₁…₂₂` with `getElem` lemmas keyed on `Fin.castAdd`/`Fin.natAdd`, round-trips, and the block-decomposition `fromBlocks_mul_fromBlocks` (`HexMatrix/Block.lean`); `Matrix.pad` / `takeCols` / `takeCols_takeRows_pad` / `row_pad` / `col_pad` and the padding lemma `takeCols_takeRows_mul_pad` (`HexMatrix/Pad.lean`, `HexMatrix/Submatrix.lean`); the `Winograd` structure with output-block theorems `Winograd.c11/c12/c21/c22` and matrix distributivity `add_mul`/`mul_add`/`sub_mul`/`mul_sub` (`HexMatrix/Winograd.lean`, `HexMatrix/MatrixAlgebra.lean`). `mul` is the `noncomputable` reference with the `@[csimp]` twin `mulImpl` and `mul_eq_mulImpl` (`HexMatrix/Basic.lean`).

**Deliverables.** A new file `HexMatrix/Strassen.lean` (add it to the `HexMatrix.lean` umbrella), containing, per the SPEC:

1. `structure StrassenConfig (R : Type u) where cutoff : Nat; baseMul : {n m k : Nat} → Matrix R n m → Matrix R m k → Matrix R n k` — data-only, no algebraic instances.
2. `def StrassenConfig.Valid [Mul R] [Add R] [OfNat R 0] (cfg : StrassenConfig R) : Prop := ∀ {n m k} (X : Matrix R n m) (Y : Matrix R m k), cfg.baseMul X Y = mul X Y`.
3. `def strassenDefault : StrassenConfig R` with `baseMul := mulImpl` and a provisional `cutoff` (pick 64; the docstring must say the value is provisional until the wave-3 bench measures the real crossover, per SPEC § Benchmarks).
4. `theorem strassenDefault_valid : (strassenDefault (R := R)).Valid`, proved from `mul_eq_mulImpl` (needs `[Mul R] [Add R] [OfNat R 0]`).
5. `def mulStrassen [Mul R] [Add R] [Sub R] [OfNat R 0] (cfg : StrassenConfig R) (M : Matrix R n m) (N : Matrix R m k) : Matrix R n k`, computable and `@[expose]`, recursing on the runtime dimensions:
   - **Base:** when `n ≤ 1 ∨ m ≤ 1 ∨ k ≤ 1 ∨ n < cfg.cutoff ∨ m < cfg.cutoff ∨ k < cfg.cutoff`, return `cfg.baseMul M N`. The `≤ 1` disjuncts are config-independent, so `cutoff = 0` cannot break termination (SPEC § "Recursion, base case, and the customizable backend").
   - **Recursive step:** pad-and-split. The recommended shape avoids all dimension casts: set `h := (n+1)/2`, `w := (m+1)/2`, `d := (k+1)/2` and pad straight to the sum-shaped types, `pad M (h+h) (w+w) : Matrix R (h+h) (w+w)` and `pad N (w+w) (d+d)` — `pad` takes arbitrary target dimensions, `n ≤ h+h` holds by `omega`-trivial arithmetic, and `toBlocks₁₁…₂₂` then apply with no `Nat` cast anywhere. Extract the eight quadrants, run the Winograd schedule (S₁…S₄, T₁…T₄ via matrix `+`/`-`, P₁…P₇ via **seven recursive `mulStrassen` calls**, U₁…U₇), assemble with `fromBlocks`, and crop with `takeCols (takeRows … n _) k _` back to `Matrix R n k`. Follow the SPEC's schedule exactly; the fifteen-addition count is the point of Winograd's variant. Naive per-sum storage is explicitly fine for this first implementation (SPEC § "Avoiding sub-block copies"); the two-buffer storage schedule is later follow-up.
   - **Termination:** well-founded on `n + m + k` (`termination_by`). The recursion fires only when all of `n`, `m`, `k` are `≥ 2` and `≥ cutoff`; each recursive call has dimensions `(h, w, d)` with `h ≤ (n+1)/2 < n` for `n ≥ 2`, so the sum strictly decreases (`decreasing_by` with `omega` should close it).
6. `theorem mulStrassen_eq_mul [Lean.Grind.Ring R] (cfg : StrassenConfig R) (h : cfg.Valid) (M : Matrix R n m) (N : Matrix R m k) : mulStrassen cfg M N = mul M N`. Proof skeleton: induction following the recursion (`fun_induction mulStrassen` if it applies to the WF definition, else strong induction on `n + m + k` unfolding with the equation lemma). Base case is `h …`. Recursive case: the seven `P`s equal the corresponding `mul` of quadrants by the induction hypotheses; instantiate the `Winograd` structure (its `hS…`/`hU…` equations hold by `rfl` for the computed intermediates, `hP…` by the IHs), read off `c11/c12/c21/c22`, rewrite the assembled `fromBlocks` against `fromBlocks_mul_fromBlocks` backwards, then `fromBlocks_toBlocks` collapses the quadrant products of the padded operands, and `takeCols_takeRows_mul_pad` recovers the unpadded product.
7. In `HexMatrix/Basic.lean`, update the `mul` docstring sentence "It is not yet implemented." to point at the now-real `mulStrassen`.

Do **not** register any `@[csimp]` from `mul` to `mulStrassen` — the SPEC is explicit that this is type-impossible (`mul` lacks `[Sub R]`) and that `mulStrassen` is a separate ring-level entry point. Do not switch any callers; that is separate follow-up work. No `mulStrassenImpl` twin is needed now (SPEC point 1).

**Constraints.**
- `HexMatrix` is Mathlib-free: no Mathlib imports or tactics; ring reasoning through `grind` / `Lean.Grind.Ring`. Read `.claude/skills/hex-lean-mathlib-boundary/SKILL.md` § "Proving inside the Mathlib-free files" first.
- Module-system conventions as in the existing files (`module`, `public import`, `public section`, `@[expose]`, copyright header `Authors: Kim Morrison`).
- Naming per SPEC § "New public names": `Matrix.StrassenConfig`, `Matrix.StrassenConfig.Valid`, `Matrix.strassenDefault`, `Matrix.strassenDefault_valid`, `Matrix.mulStrassen`, `Matrix.mulStrassen_eq_mul`.
- No `axiom`, no `native_decide`, no `sorry` in the final PR. Termination must be genuine WF recursion, not `partial`.
- `mulStrassen` is the future runtime entry point: keep the recursive step free of gratuitous copies beyond the sanctioned quadrant materialization (use `let`s to share the padded operands and intermediates).

**Verification.** Whole-graph `lake build` green (downstream libraries including the `*Mathlib` bridge layers recompile). As a fast executable sanity check while developing, `#guard`-style evaluation of `mulStrassen strassenDefault A B == mul A B`-shaped comparisons on a couple of small `Int` matrices in a scratch file is encouraged, but committed conformance fixtures are a separate wave-3 issue — do not add them here.

kim-em:
This PR adds the Strassen-Winograd multiplication entry point to `hex-matrix`, the core deliverable of the Strassen-Winograd SPEC: the data-only `StrassenConfig` (recursion `cutoff` plus a pluggable `baseMul` base kernel) with its `Valid` predicate, the default configuration `strassenDefault` (naive `mulImpl` base kernel and a provisional `cutoff` of 64 pending the wave-3 bench) proved valid by `strassenDefault_valid`, the computable well-founded-recursive `mulStrassen` following the fifteen-addition seven-product Winograd schedule with a config-independent `≤ 1` base condition and per-level zero-padding of odd dimensions, and the correctness theorem `mulStrassen_eq_mul` proving the whole recursion equal to the reference `mul` for every valid config by composing the three wave-1 lemmas (the Winograd schedule identity, the block decomposition `fromBlocks_mul_fromBlocks`, and the padding lemma `takeCols_takeRows_mul_pad`). `mulStrassen` is a separate ring-level entry point rather than a `@[csimp]` replacement of `mul`, because the Winograd schedule subtracts blocks and so needs `[Sub R]` which `mul` does not have; the `mul` docstring is updated to point at it.

Closes #8676.

🤖 Prepared with Claude Code

### DIFF
--- a/HexMatrix.lean
+++ b/HexMatrix.lean
@@ -16,6 +16,7 @@ public import HexMatrix.Winograd
 public import HexMatrix.Elementary
 public import HexMatrix.Submatrix
 public import HexMatrix.Pad
+public import HexMatrix.Strassen
 public import HexMatrix.Gram
 
 public section
--- a/HexMatrix/Basic.lean
+++ b/HexMatrix/Basic.lean
@@ -333,10 +333,12 @@ reason about. Compiled code runs `mulImpl`, which transposes `N` once (via the
 rebuilt for every row of `M`.
 
 Strassen-Winograd multiplication, with a customizable base kernel for small
-sizes, is specified in `HexMatrix/SPEC/hex-matrix.md` under "Strassen-Winograd
-multiplication" as `mulStrassen`. It is a separate ring-level entry point (it
-needs `[Sub R]`, which this `mul` does not), proved equal to `mul` and swapped
-in for compiled code by `@[csimp]`. It is not yet implemented.
+sizes, is implemented as `mulStrassen` in `HexMatrix/Strassen.lean` and proved
+equal to this `mul` by `mulStrassen_eq_mul`. It is a separate ring-level entry
+point rather than a `@[csimp]` replacement of `mul`: the Winograd schedule
+subtracts blocks, so `mulStrassen` needs `[Sub R]`, which this `mul` does not
+have, and a `@[csimp]` replacement must preserve the declaration's type. Callers
+over a ring opt into it explicitly.
 -/
 @[expose]
 noncomputable def mul [Mul R] [Add R] [OfNat R 0] (M : Matrix R n m) (N : Matrix R m k) :
--- a/HexMatrix/Strassen.lean
+++ b/HexMatrix/Strassen.lean
@@ -64,6 +64,7 @@ structure StrassenConfig (R : Type u) where
 /-- A configuration is **valid** when its base kernel agrees with the reference
 `mul` on every input. The correctness theorem `mulStrassen_eq_mul` is stated under
 this hypothesis, keeping the proof out of the `StrassenConfig` data record. -/
+@[expose]
 def StrassenConfig.Valid [Mul R] [Add R] [OfNat R 0] (cfg : StrassenConfig R) : Prop :=
   ∀ {n m k} (X : Matrix R n m) (Y : Matrix R m k), cfg.baseMul X Y = mul X Y
 
--- a/HexMatrix/Strassen.lean
+++ b/HexMatrix/Strassen.lean
@@ -1,0 +1,182 @@
+/-
+Copyright (c) 2026 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
+-/
+
+module
+
+public import HexMatrix.Block
+public import HexMatrix.Pad
+public import HexMatrix.Winograd
+
+public section
+
+/-!
+Strassen-Winograd matrix multiplication.
+
+`mulStrassen` is the recursive, ring-level multiplication entry point specified in
+`HexMatrix/SPEC/hex-matrix.md` § "Strassen-Winograd multiplication". It computes a
+2×2 block product with **seven** recursive block multiplications (`P₁…P₇`) and
+**fifteen** block additions/subtractions (`S₁…S₄`, `T₁…T₄`, `U₁…U₇`), following
+Winograd's memory-efficient schedule, giving `Θ(n^{log₂ 7})` coefficient
+multiplications.
+
+The cutoff below which the recursion falls back to a base kernel, and the base
+kernel itself, live in the data-only `StrassenConfig`. A config is `Valid` when
+its base kernel agrees with the reference `mul`; the default config
+`strassenDefault` uses the naive `mulImpl` and `strassenDefault_valid` proves it
+valid. The correctness theorem `mulStrassen_eq_mul` proves the whole recursion
+equal to `mul` for every valid config, composing the three wave-1 lemmas: the
+Winograd schedule identity (`Winograd.c11…c22`), the block decomposition
+(`fromBlocks_mul_fromBlocks`), and the padding lemma
+(`takeCols_takeRows_mul_pad`).
+
+`mulStrassen` needs subtraction on `R` (Winograd subtracts blocks), so it is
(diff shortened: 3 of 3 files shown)


kim-em:
Review round (Claude session + Codex second opinion): no correctness findings — the schedule transcription, base-condition semantics, termination, and the fun_induction/Winograd-record proof all check out. One fix applied from the Codex pass: `StrassenConfig.Valid` now carries `@[expose]`, matching the repo's convention for public Prop defs (`isLLLReduced`, `memLattice`), since wave-3 custom-config validity proofs are written in downstream modules; verified with a downstream-style smoke proof. The materialized-quadrant recursion is the SPEC-sanctioned interim representation; the view-based refinement is tracked with the flat backing in #8652.

PR #8682, opened 2026-07-08
test(hex-matrix): conformance guards for mulStrassen

### LINKED ISSUE #8679: test(hex-matrix): conformance guards for mulStrassen
This issue adds the conformance guards for `mulStrassen`, per [HexMatrix/SPEC/hex-matrix.md § Conformance](https://github.com/kim-em/hex-dev/blob/main/HexMatrix/SPEC/hex-matrix.md#conformance). Wave 3 of the Strassen implementation (tracking: #8665); requires `mulStrassen` on `main` (#8676).

**Current state.** `conformance/HexMatrix/Conformance.lean` has `#guard` checks for the dense base (constructors, `mul`, elementary ops) on committed fixtures, with the module docstring cataloguing covered operations/properties/edge cases. Nothing exercises `mulStrassen`.

**Deliverables.** Extend `conformance/HexMatrix/Conformance.lean` (same file, same idiom — no new drivers, no new CI jobs, per `SPEC/CI.md`):

1. `#guard` checks that `mulStrassen cfg A B = mul A B` on committed fixtures spanning: even dimensions, odd dimensions, a prime dimension, the empty matrix (a zero axis), 1×1, rectangular shapes with distinct `n`, `m`, `k`, and a dimension straddling the default cutoff (i.e. one fixture at ≥ `strassenDefault.cutoff` so the recursion actually fires with the default config — mind the wallclock cost of the fixture size you pick; one recursion level is enough).
2. Both configs: `strassenDefault` and a non-default custom base kernel (a deliberately different but valid `baseMul` — e.g. a transpose-twice or `ofFn`-direct naive kernel — plus a small-cutoff variant such as `cutoff := 0` so deep recursion is exercised on small fixtures). The 1×1 and empty fixtures must run under `cutoff := 0` too, exercising the config-independent `≤ 1` base condition that keeps the recursion terminating.
3. Update the module docstring catalogue (covered operations/properties/edge cases) accordingly. Oracle: none (structural-layer exact arithmetic, as for the existing multiplication guards); Mode: always.

These checks run through `#guard`, the compiled-evaluator path, not kernel `decide` — `mulStrassen` is defined by well-founded recursion and does not reduce cheaply in the kernel (SPEC § Conformance). `mul` is `noncomputable` with a `@[csimp]` twin, so `mulStrassen cfg A B = mul A B` inside `#guard` evaluates both sides compiled; follow the existing guards' equality idiom.

**Constraints.** Conformance sub-project conventions per `SPEC/testing.md`; no Mathlib imports; no `axiom` / `native_decide` / `sorry`; keep total added wallclock small (these guards run in CI's single job).

**Verification.** The conformance Lake target passes locally with the new guards; whole-graph `lake build` green.

kim-em:
This PR extends `conformance/HexMatrix/Conformance.lean` with differential `#guard` checks that `mulStrassen cfg A B = mul A B` (the reference product) on committed fixtures, following the existing file's compiled-evaluator idiom. The fixtures span even, odd, and prime square dimensions, a distinct-`n`,`m`,`k` rectangle, a zero contraction axis and a zero output axis, 1×1, and a 96×96 square past the default cutoff so the recursion fires under `strassenDefault`. They run under the default config, a custom config carrying a deliberately different but valid triple-loop base kernel with an explicit `cutoff := 4`, and a deep-recursion config with `cutoff := 0` that drives the config-independent `≤ 1` base condition (the 1×1 and empty fixtures exercise exactly that terminating case). The custom configs pin explicit cutoffs rather than the provisional `strassenDefault.cutoff` of 64, and the one default-config recursion test is a 96×96 square comfortably above any plausible measured crossover, so a later re-measurement of that constant cannot silently turn a recursion test into a base-kernel test. The module docstring catalogue is updated accordingly.

Closes #8679.

🤖 Prepared with Claude Code

### DIFF
--- a/conformance/HexMatrix/Conformance.lean
+++ b/conformance/HexMatrix/Conformance.lean
@@ -167,13 +167,14 @@ forces one-or-two recursion levels on the small square/rectangular fixtures with
 deliberately different — but valid — textbook triple-loop base kernel; `cfgDeep`
 (`cutoff := 0`) drives the recursion down to the config-independent `≤ 1` base
 condition (the 1×1 and empty fixtures exercise exactly that terminating case). The
-single `strassenDefault` recursion test uses a 96×96 square, comfortably above any
-plausible measured crossover so it keeps firing the recursion regardless of #8680.
+single `strassenDefault` recursion test derives its dimension from
+`strassenDefault.cutoff` itself, so it keeps firing the recursion whatever value a
+future re-measurement assigns to that constant.
 -/
 
 /-- Deterministic Int fixture generator; `seed` distinguishes the two operands so
 the product is not accidentally symmetric. Entries are kept small to keep the
-96×96 default-config guard in the millisecond range. -/
+cutoff-derived default-config guard in the millisecond range. -/
 private def genInt (seed n m : Nat) : Matrix Int n m :=
   Matrix.ofFn fun i j => (((i.val + 1) * (j.val + 2) + seed * (i.val + j.val) : Int) % 13) - 6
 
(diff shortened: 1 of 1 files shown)


kim-em:
Review round (Claude session + Codex second opinion): coverage and fixture design confirmed against the SPEC's conformance list. One fix applied from the Codex pass: the `strassenDefault` recursion guard now derives its dimension from `strassenDefault.cutoff` itself (`max 2 (cutoff + 1)`) instead of a fixed 96×96, so a re-measured cutoff above 96 cannot silently turn it into a base-kernel-only check; also trimmed issue-number process commentary from the source per repo doctrine.

PR #8683, opened 2026-07-08
feat(arith): full-range Barrett reduce and two-word delayed-reduction accumulator

### LINKED ISSUE #8681: feat(berlekamp): delayed-reduction Barrett base kernel as the demonstration StrassenConfig
This issue delivers the SPEC's demonstration non-default `StrassenConfig`: a delayed-reduction (periodic-reduction) Barrett base kernel for the prime-field residues, with its validity proof and the comparison measurement that decides whether it ships. Read [HexMatrix/SPEC/hex-matrix.md § A demonstration non-default config](https://github.com/kim-em/hex-dev/blob/main/HexMatrix/SPEC/hex-matrix.md#a-demonstration-non-default-config) in full — every design decision below is stated there, including the two honesty constraints. Wave 3 of the Strassen implementation (tracking: #8665); requires `mulStrassen` on `main` (#8676).

**Current state.** `StrassenConfig` / `mulStrassen` / `Valid` live in `HexMatrix/Strassen.lean`. The ZMod64-level `Hex.BarrettCtx p` wrapper is `HexModArith/HotLoop.lean` over the `UInt64`-level `BarrettCtx` in `HexArith/Barrett/{Context,Reduce}.lean`; `BarrettCtx.toNat_mulMod`-style single-multiply lemmas exist, but there is **no wide-accumulator layer** (the SPEC is explicit that the small verified `UInt128`-or-`UInt64`-pair add-and-reduce lemmas are themselves a deliverable here). The consuming multiply is the `HexBerlekamp` nullspace computation (`HexBerlekamp/RabinSoundness/KernelWitness.lean`).

**Placement.** The kernel and config need both `HexMatrix` and the Barrett layer in scope. `HexBerlekamp` already imports both (`HexBerlekamp/Basic.lean` imports `HexMatrix`); `hex-mod-arith` and `hex-arith` must NOT gain a `HexMatrix` dependency (that would invert the released-library dependency order in `libraries.yml`). So: the machine-word accumulator lemmas go beside the existing Barrett layer in `HexArith`/`HexModArith`; the `StrassenConfig` value, its kernel, and its `Valid` proof go in a new `HexBerlekamp` file (or another library above both — justify the choice in the PR body).

**Deliverables.**

1. The **periodic-reduction kernel**: a base multiply for the Barrett-reduced prime field (`ZMod64 p` with a `BarrettCtx`) that accumulates each dot product in a wide accumulator and reduces modulo `p` every fixed number of terms chosen to preclude overflow, rather than reducing after every multiply-add. It must be the periodic form, not a single end reduction: `Valid` quantifies over all `n m k`, and at a base-case leaf the inner dimension `m` is unbounded, so only periodic reduction is provably correct for every `m` (SPEC). `BarrettCtx` keeps `p < 2^32`, so each product fits in 64 bits and a bounded run fits in the wide accumulator between reductions.
2. The verified accumulator layer the proof needs: the accumulator invariant (running sum's residue mod `p`), the per-window no-overflow bound, the final-reduction step, and the equality to the `ZMod64` dot product the naive kernel computes — as named lemmas beside the Barrett layer. `BarrettCtx.toNat_mulMod` is the single-multiply building block, not the whole chain.
3. The demonstration config (a `StrassenConfig (ZMod64 p)`-producing definition taking the `BarrettCtx`), with its `Valid` proof, and dispatch: the size-specialized kernel handles its fast path and falls back to the naive kernel off it (SPEC: one function dispatching on runtime dimensions).
4. The **comparison measurement**: default-base vs delayed-reduction-base `mulStrassen` on the prime field, as a separate comparison plot (not a third series on the `Int` scaling figure), committed under `reports/figures/` with its script under `scripts/plots/`, following the repo's plotting idiom. Extend the existing single bench job only within its wallclock budget (`SPEC/CI.md`); the measurement itself is local.
5. **Honesty constraints** (SPEC, verbatim obligations): (a) a base kernel moves constants and the crossover, never the asymptotic slope — present it that way; (b) it ships only if it measurably beats the default naive base kernel on that type. If the measured win does not materialize, revert the demonstration to a trivial alternate config that still exercises the plug-in path, commit the negative measurement, and file the delayed-reduction kernel as follow-up work with the evidence — do not force it in.

**Constraints.** The Barrett/accumulator lemmas live in Mathlib-free layers: core-Lean tactics + `grind` only; read `.claude/skills/hex-lean-mathlib-boundary/SKILL.md` first. Module-system conventions as in the surrounding files. No `axiom` / `native_decide` / `sorry`. Names: short verb-noun, qualifiers in namespaces. This is the largest wave-3 item — if, on reading the Barrett layer, you conclude the verified wide-accumulator chain is too large for one PR, STOP and comment on this issue with the natural split (accumulator lemmas first, kernel+config second) rather than opening an oversized or sorry-carrying PR.

**Verification.** Whole-graph `lake build` green; the comparison plot committed; the PR body states the measured outcome (speedup or the honest negative) and which branch of honesty-constraint (b) applies.

kim-em:
This PR adds the machine-word wide-accumulator substrate that the delayed-reduction Barrett base kernel of https://github.com/kim-em/hex-dev/issues/8681 folds over, landing the "accumulator lemmas first" half of that issue's own suggested split rather than an oversized or sorry-carrying single PR (see the issue comment for why the full chain — periodic-reduction fold invariant, ZMod64 bridge, kernel dispatch, `StrassenConfig` + `Valid`, and a config-deciding bench/plot measurement — is two PRs, and for the continuation design). It observes that `barrettReduce` already computes `T % p` for every input word and not only for `T < p^2` (the `p^2` hypothesis was only ever used to derive the always-true `T < 2^64`), stating that relaxed full-range correctness as `toNat_barrettReduce_mod` with `barrettReduce_lt_mod` for the residue bound; on top of it `accVal` / `accAddWord` / `accVal_accAddWord` give an exact two-word accumulate-a-product step under a no-carry-overflow bound, and `accReduce` / `toNat_accReduce` / `accReduce_lt` reduce the two-word accumulator modulo `p` by composing three full-range Barrett reductions through the precomputed constant `2^64 % p`. The whole-graph `lake build` is green and the new file carries no `sorry`, `axiom`, or `native_decide`; it does not close #8681, which stays open for the kernel + config + measurement.

🤖 Prepared with Claude Code

### DIFF
--- a/HexArith.lean
+++ b/HexArith.lean
@@ -12,6 +12,7 @@ public import HexArith.Nat.Prime
 public import HexArith.Barrett.ReduceNat
 public import HexArith.Barrett.Reduce
 public import HexArith.Barrett.Context
+public import HexArith.Barrett.Accumulator
 public import HexArith.ExtGcd
 public import HexArith.Montgomery.Context
 public import HexArith.Montgomery.InvNat
--- a/HexArith/Barrett/Accumulator.lean
+++ b/HexArith/Barrett/Accumulator.lean
@@ -1,0 +1,245 @@
+/-
+Copyright (c) 2026 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
+-/
+
+module
+
+public import HexArith.Barrett.Reduce
+
+public section
+
+/-!
+Wide-accumulator layer for delayed-reduction Barrett multiplication.
+
+The single-multiply Barrett reducer `barrettReduce` reduces after every product.
+A delayed-reduction dot product instead accumulates several products in a
+two-word (`128`-bit) accumulator and reduces modulo `p` only periodically. This
+module provides the verified pieces that make that sound:
+
+* `barrettReduce` is correct on the **whole** `UInt64` range, not just on inputs
+  below `p^2` (`toNat_barrettReduce_mod`): every `UInt64` is already below the
+  radix `2^64`, which is the only bound the reducer needs.
+* a two-word accumulator `accVal lo hi := lo + hi * 2^64`, an add-a-product step
+  `accAddWord`, and its exact-value lemma under the no-carry-overflow bound
+  (`accVal_accAddWord`);
+* a two-word reduction `accReduce` that returns `(lo + hi * 2^64) % p` using the
+  precomputed constant `2^64 % p` (`toNat_accReduce`, `accReduce_lt`).
+
+`BarrettCtx` keeps `p < 2^32`, so each residue product fits in one `UInt64` and a
+bounded run of them fits below `2^128` between reductions; that is why the
+accumulator never overflows and periodic reduction is exact for every inner
+length.
+-/
+
+namespace BarrettCtx
+
+variable {p : UInt64}
+
+/--
+`barrettReduce` computes `T % p` for **every** input word, not only for inputs
+below `p^2`. The single-word reducer only needs `T < 2^64`, which holds for every
+`UInt64`; the `p^2` bound in `toNat_barrettReduce_eq_mod` is stronger than
+necessary. This is the reduction the wide accumulator flushes through.
+-/
+@[grind =]
+theorem toNat_barrettReduce_mod (ctx : BarrettCtx p) (T : UInt64) :
+    (barrettReduce ctx T).toNat = T.toNat % p.toNat := by
+  have hp0 : 0 < p.toNat := Nat.lt_trans (by decide : 0 < 1) ctx.p_gt
+  have hT_word : T.toNat < barrettRadix := by
+    simpa [barrettRadix] using UInt64.toNat_lt_word T
+  let q := UInt64.mulHi T ctx.pinv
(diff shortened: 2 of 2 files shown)


kim-em:
Review round (Claude session + Codex second opinion): the full-range `barrettReduce` correctness claim is confirmed on both reviews (for exact `μ = ⌊2^64/p⌋` and any `T < 2^64`, `q ≤ ⌊T/p⌋ ≤ q+1`, so the remainder is `< 2p` and one conditional subtraction suffices; edge cases at tiny `p` and `T = 2^64−1` covered by the same bound). No merge-blocking findings. Noted for the #8681 continuation: the periodic-reduction fold needs an explicit per-window bound discharging `accVal_accAddWord`'s `hi + 1 < 2^64` hypothesis at every step, and a `ctx`-level `2^64 % p` helper would cut the `cR`-plus-proof plumbing. Optional later refactor: the full-range theorems could migrate into `Reduce.lean` with the `p²`-hypothesis lemma delegating to them.

PR #8686, opened 2026-07-08
feat(berlekamp): delayed-reduction Barrett base kernel, config, and comparison

### LINKED ISSUE #8681: feat(berlekamp): delayed-reduction Barrett base kernel as the demonstration StrassenConfig

kim-em:
This PR completes the SPEC's demonstration non-default `StrassenConfig` for the Barrett-reduced prime field by adding, on top of the wide-accumulator layer, the periodic-reduction fold `BarrettCtx.foldReduce` with its per-window no-overflow invariant `accFold_spec` (the fold keeps `hi ≤ count < barrettWindow`, discharging `accVal_accAddWord`'s `hi + 1 < 2^64` hypothesis at every step, with `barrettWindow := 2^32 < 2^64`), a context-level `radixResidue = 2^64 % p` helper that cuts the `cR`-plus-proof plumbing, the ZMod64 dot-product bridge `delayedDot` (with an allocation-free `delayedDotImpl` swapped in by `@[csimp]`) proved equal to the naive `Vector.dotProduct` in `delayedDot_eq_dotProduct`, the size-dispatching base kernel `barrettBaseMul` (delayed fast path with a `mulImpl` fallback), and the verified config `strassenBarrett` with `strassenBarrett_valid`, all placed in `HexBerlekamp/DelayedKernel.lean` (above both HexMatrix and the Barrett layer, so hex-arith and hex-mod-arith gain no HexMatrix dependency). The committed comparison measurement (`hexstrassen_compare` driver, `reports/bench-results/strassen-base-kernel-comparison.json`, plotted by `scripts/plots/strassen-base-kernel-comparison.py` into `reports/figures/strassen-base-kernel-comparison.svg`) is an honest negative: default-base versus delayed-reduction-base `mulStrassen` on `ZMod64 2147483647` shows the delayed kernel is roughly 7-10x slower across dimensions 64..256, because its boxed two-word accumulator bookkeeping and per-term `UInt64.addCarry` extern call outweigh the modular reductions it defers relative to the default's tight single-extern `ZMod64.mul` (a base kernel moves the constant factor and crossover, never the asymptotic slope, and it is presented that way). Honesty constraint (b) therefore applies: the shipped demonstration config is the trivial `strassenDemo` (naive base kernel, non-default cutoff) that still exercises the pluggable-kernel path, and the verified `strassenBarrett` stays as follow-up work for its constant-factor optimization (or a bit-packed GF(2) four-Russians kernel). Whole-graph `lake build` is green; no `sorry`/`axiom`/`native_decide`.

Closes #8681.

🤖 Prepared with Claude Code

### DIFF
--- a/HexArith/Barrett/Accumulator.lean
+++ b/HexArith/Barrett/Accumulator.lean
@@ -242,4 +242,165 @@ theorem accReduce_lt (ctx : BarrettCtx p) (cR lo hi : UInt64)
   rw [UInt64.lt_iff_toNat_lt, toNat_accReduce ctx cR lo hi hcR]
   exact Nat.mod_lt _ (Nat.lt_trans (by decide : 0 < 1) ctx.p_gt)
 
+/-! ### The radix residue as a context-level constant
+
+`accReduce` needs the constant `2^64 % p`; storing it on the context (rather than
+threading a `cR` word plus its `hcR` proof through every caller) keeps the
+delayed-reduction fold below free of that plumbing. -/
+
+/-- `2^32 < 2^64`, the radix headroom that keeps a reduced residue and the
+window inside a single machine word. -/
+private theorem two_pow_32_lt_word : (2 : Nat) ^ 32 < UInt64.word := by
+  rw [UInt64.word]; exact Nat.pow_lt_pow_right (by decide) (by decide)
+
+/-- The radix residue `2^64 % p`, stored as a context-level constant so callers
+of `accReduce`/`foldReduce` need not carry the word and its correctness proof.
+Computed with a `Nat` reduction of the `2^64` radix (done once per context), so
+no `128`-bit machine literal is required. -/
+@[expose]
+def radixResidue (_ctx : BarrettCtx p) : UInt64 :=
+  UInt64.ofNat (UInt64.word % p.toNat)
+
+/-- `radixResidue` stores exactly `2^64 % p`, the `hcR` side condition every
+`accReduce`/`foldReduce` lemma needs. -/
+@[grind =]
+theorem toNat_radixResidue (ctx : BarrettCtx p) :
+    (radixResidue ctx).toNat = UInt64.word % p.toNat := by
+  have hp0 : 0 < p.toNat := Nat.lt_trans (by decide : 0 < 1) ctx.p_gt
+  have hlt : UInt64.word % p.toNat < p.toNat := Nat.mod_lt _ hp0
+  have hpw : p.toNat < UInt64.word := Nat.lt_trans ctx.p_lt two_pow_32_lt_word
+  have hb : UInt64.word % p.toNat < UInt64.size := by
+    simpa [UInt64.size, UInt64.word] using Nat.lt_trans hlt hpw
+  rw [radixResidue, UInt64.toNat_ofNat_of_lt' hb]
+
+/-! ### Periodic-reduction fold
+
+A delayed-reduction dot product accumulates one-word products into the two-word
+accumulator and flushes through `accReduce` every `barrettWindow` additions. The
+window is fixed and independent of the inner dimension, so a run of any length
+stays below `2^128`: each add grows the high word by at most one, and a flush
+resets it, so the high word never reaches the wraparound boundary
+(`accFold_spec`, the per-window no-overflow invariant). This is why periodic
+reduction is correct for *every* inner length, where a single end reduction is
+not. -/
+
+/-- Number of one-word products accumulated between reductions. Any value below
+`2^64` keeps the two-word accumulator's high word from overflowing (the add
+grows it by at most one per step); `2^32` sits comfortably above every realistic
+inner dimension, so a base-kernel dot product reduces exactly once. -/
+@[irreducible]
+def barrettWindow : Nat := 2 ^ 32
+
+theorem barrettWindow_pos : 0 < barrettWindow := by
+  unfold barrettWindow; exact Nat.two_pow_pos 32
+
+theorem barrettWindow_lt_word : barrettWindow < UInt64.word := by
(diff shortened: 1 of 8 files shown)


kim-em:
Delayed-kernel optimization filed as follow-up: https://github.com/kim-em/hex-dev/issues/8687 (perf(berlekamp): make the delayed-reduction Barrett base kernel beat the default), with the measured slowdown table and diagnosis.

kim-em:
Review round (Claude session + Codex second opinion): both reviews confirm the per-window invariant (`hi ≤ count < barrettWindow`, reset on flush, empty-list and boundary cases covered), the ZMod64 bridge (single-word products from `p < 2^31`, generalized-accumulator fold induction, `ofNat` round-trip), and — explicitly — that keeping the verified-but-slow `strassenBarrett` in-tree as non-default code with the committed negative measurement is the faithful reading of the SPEC's honesty constraint (b). Two low-severity fixes applied: the `strassenBarrett` docstring now labels it the verified candidate rather than "the demonstration" (that role belongs to `strassenDemo`), and the comparison driver's identical-branch forcing trick was replaced with a checksum sink printed outside the timed region.

Issue #8692, opened 2026-07-09
refactor(arith): give binom and choose a common home and dedup the two binomial folds


kim-em:
The repo now has two nearly identical executable binomial-coefficient folds in two libraries:

- `Hex.ZPoly.binom` (`HexPolyZ/Mignotte.lean:29`): `if n < k then 0 else foldl over range (min k (n-k)) of acc * (n-i)/(i+1)` — the `min` is a symmetry optimization.
- `Hex.Nat.chooseFast` / `Hex.Nat.choose` (`HexArith/Nat/Prime.lean`): the same fold without the `min`, with a Pascal spec twin `choose` and a `@[csimp]` swap.

https://github.com/kim-em/hex-dev/pull/8684 already bridged them (`binom_le_central` bridges `binom` to `Hex.Nat.choose`, and `HexPolyZMathlib/Mignotte.lean` has `binom_eq_choose : binom n k = Nat.choose n k`), and made `HexPolyZ` depend on `HexArith`. So they are now one concept living in two places.

Give them a common home and dedup. Suggested target: `HexArith.Nat` (foundational, Mathlib-free, `deps: []`, already owns `choose`). Options to weigh:
- Move `binom` (the min-optimized fold) into `HexArith.Nat` next to `choose`, renaming `Hex.ZPoly.binom -> Hex.Nat.binom`, and prove `binom = choose` there (folding the existing `binom_eq_choose` bridge down into the Mathlib-free layer if the symmetry lemma is available, or keeping the Mathlib bridge thin).
- Or keep one fold and express the other as a definitional wrapper.

Consumers to update (the rename ripples): `HexPolyZ/Mignotte.lean`, `HexPolyZMathlib/Mignotte.lean` (`binom_eq_choose`), `HexBerlekampZassenhaus/Recombination.lean`, `HexBerlekampZassenhausMathlib/PublicSurface.lean`, `HexManual/Chapters/HexPolyZ.lean`, `bench/HexPolyZ/Bench.lean`, `conformance/HexPolyZ/{Conformance,EmitFixtures}.lean`. Values must be unchanged (conformance fixtures untouched).

Note: overlaps with #8693 (which moves the pure-`choose` monotonicity lemmas into `HexArith.Nat`). If both land, coordinate so the final home is consistent; whichever lands second rebases.

kim-em:
[...]

Issue #8693, opened 2026-07-09
refactor(arith): move the central-binomial monotonicity lemmas upstream to HexArith.Nat


kim-em:
https://github.com/kim-em/hex-dev/pull/8684 introduced `Hex.ZPoly.binom_le_central` (`HexPolyZ/Mignotte.lean`) and a chain of supporting lemmas to prove the central binomial dominates the factor rectangle. Several of these are pure facts about the Pascal `Hex.Nat.choose` and use only recurrences already in `HexArith/Nat/Prime.lean` (`choose_succ_succ`, `succ_mul_choose_succ`), so they can move upstream next to `choose` with **no new imports**, making them reusable and shortening `HexPolyZ/Mignotte.lean`.

Move upstream into `HexArith.Nat` (the `Hex.Nat` namespace, `HexArith/Nat/Prime.lean` or a sibling), as far as each goes without needing imports beyond what that file already has:

- `choose_le_succ_left` — a row increases up to its centre.
- `choose_le_center` (currently fuel-recursion) — left half is bounded by the central entry.
- `central_even`, `central_odd`, `centralChoose_le_succ`, `centralChoose_mono` — the central binomial is monotone in the row.
- `foldl_choose` overlaps with the existing private `chooseFast_foldl` in the same file; dedup rather than duplicate.

Leave `binom_le_central`, `binom_symm`, and `binom_eq_choose_left` in `HexPolyZ/Mignotte.lean` (they mention `Hex.ZPoly.binom`, which lives in `HexPolyZ`), rewritten as thin wrappers delegating to the moved `Hex.Nat` lemmas.

No behavioural change; this is a pure relocation/dedup. Verify the whole graph still builds (`HexArith`, `HexPolyZ`, `HexBerlekampZassenhaus*`) and no `sorry`/`axiom` introduced.

Note: overlaps with #8692 (unifying `binom` and `choose`). If #8692 moves `binom` into `HexArith.Nat` too, `binom_le_central` can follow; coordinate so the second to land rebases.

PR #8696, opened 2026-07-09
refactor(arith): give binom and choose a common home and dedup the two binomial folds

kim-em:
This PR gives the two executable binomial-coefficient folds a single home. The min-optimized fold `Hex.ZPoly.binom` and the plain fold `Hex.Nat.chooseFast` both computed the Pascal `Hex.Nat.choose`; this collapses them into one `Hex.Nat.binom` in `HexArith/Nat/Prime.lean`, next to `choose`. `binom` folds over the shorter half `min k (n - k)` of the row, is proven equal to `choose` (`binom_eq_choose`, Mathlib-free, using a new `choose_symm`), and is registered `@[csimp]` as the compiled implementation of `choose`, replacing `chooseFast`. Every compiled caller of `choose` now runs the shorter fold.

`Hex.ZPoly.binom` and its `chooseFast` twin are removed. Consumers move to `Hex.Nat.binom`: `HexPolyZ/Mignotte.lean`, `HexPolyZMathlib/Mignotte.lean`, `HexBerlekampZassenhausMathlib/PublicSurface.lean`, `HexManual/Chapters/HexPolyZ.lean`, `bench/HexPolyZ/Bench.lean`, and `conformance/HexPolyZ/Conformance.lean`. In `Mignotte.lean` the now-redundant `foldl_choose`, `binom_symm`, and `binom_eq_choose_left` helpers drop out, since `binom_le_central` reduces to the pure-`choose` statement through the full `binom_eq_choose`. Values are unchanged, so conformance fixtures are untouched.

Overlaps with #8693 (moving the central-binomial monotonicity lemmas upstream); the two touch `binom_le_central`, so whichever lands second rebases.

🤖 Prepared with Claude Code

### DIFF
--- a/HexArith/Nat/Prime.lean
+++ b/HexArith/Nat/Prime.lean
@@ -171,7 +171,7 @@ private theorem choose_succ_mul_eq (n k : Nat) :
 
 /-- Within-row multiplicative recurrence `(k+1) · choose n (k+1) = (n - k) · choose n k`.
 Reading it left to right computes one Pascal row in a single linear pass, which is
-what `chooseFast` exploits. -/
+what `binom` exploits. -/
 theorem succ_mul_choose_succ (n k : Nat) :
     (k + 1) * choose n (k + 1) = (n - k) * choose n k := by
   have hcross := choose_succ_mul_eq n k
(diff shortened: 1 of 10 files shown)


kim-em:
Closes #8692.

PR #8699, opened 2026-07-09
perf(berlekamp): make the delayed-reduction Barrett base kernel beat the default

### LINKED ISSUE #8687: perf(berlekamp): make the delayed-reduction Barrett base kernel beat the default
Follow-up from #8681 / #8686 (honesty constraint (b)). The delayed-reduction Barrett base kernel `Hex.strassenBarrett` landed verified-correct in #8686 (`HexBerlekamp/DelayedKernel.lean`, `delayedDot_eq_dotProduct` + `strassenBarrett_valid`), but the committed comparison measurement showed it does NOT beat the default base kernel, so it did not ship as the demonstration config (the trivial `strassenDemo` did). This issue tracks making it an actual speedup.

Measured `mulStrassen` on `ZMod64 2147483647` (best-of-iters wall time, cutoff 64), default-base vs delayed-base:

| n | default | delayed | delayed/default |
|---|---|---|---|
| 64 | 1.61 ms | 14.40 ms | 8.9x slower |
| 96 | 4.40 ms | 46.94 ms | 10.7x slower |
| 128 | 13.69 ms | 104.29 ms | 7.6x slower |
| 160 | 23.48 ms | 197.14 ms | 8.4x slower |
| 192 | 36.13 ms | 336.52 ms | 9.3x slower |
| 256 | 107.96 ms | 757.18 ms | 7.0x slower |

Committed evidence: `reports/bench-results/strassen-base-kernel-comparison.json`, `reports/figures/strassen-base-kernel-comparison.svg`; regenerate the data with `lake exe hexstrassen_compare` and the plot with `scripts/plots/strassen-base-kernel-comparison.py`.

Diagnosis (hypotheses to confirm/refute): the theoretical win is fewer modular reductions (one per window instead of one per multiply-add), but the current fold pays more than it saves in Lean-level overhead:

- The accumulator is a boxed `(UInt64 × UInt64 × Nat)` tuple threaded through `Fin.foldl`, allocated once per term (`O(n^2 m)` allocations), versus the default's single unboxed `ZMod64`.
- `accAddWord` calls the `UInt64.addCarry` extern once per term, comparable in cost to the default's single `lean_hex_zmod64_mul` extern, so no extern is actually saved per term.
- The window counter is a `Nat` (`barrettWindow := 2^32`); confirm whether `2^32` is recomputed per step and whether a `UInt64` counter helps.

Candidate directions:
- Replace the tuple accumulator with an unboxed structure / pack `(lo, hi)` and use a `UInt64` window counter, so the hot loop is allocation-free.
- Provide an inlined (non-extern) carry-add for the accumulate step, or a `UInt128`-style intrinsic if the toolchain gains one.
- Reconsider whether beating the C-extern `ZMod64.mul` is achievable at all for this element type at these sizes; the SPEC notes a bit-packed GF(2) four-Russians kernel would be the strongest showcase but the project has no bit-packed GF(2) matrix representation yet.

The verified `foldReduce`/`accFold_spec`/`delayedDot`/`strassenBarrett` API is already in place, so this is a performance-only remodel of the executable path plus a re-measurement; correctness is preserved by the existing theorems (and any restructured fold must keep them green).

kim-em:
This PR makes the delayed-reduction Barrett base kernel `strassenBarrett` measurably beat the default base kernel, promoting it to the shipped performance demonstration config (the win branch of the issue's decision rule). The verified reference (`foldReduce`, `accFold_spec`, `delayedDot`, `delayedDot_eq_dotProduct`, `strassenBarrett_valid`) is untouched; only the executable side of the `@[csimp]` boundary is remodeled: the boxed `(UInt64 × UInt64 × Nat)` accumulator threaded through `Fin.foldl` (one tuple allocation per term, confirmed in the generated C) becomes allocation-free scalar tail-recursive loops with the carry inlined as a wraparound compare instead of the `UInt64.addCarry` extern and the window counter dropped entirely on the `m < barrettWindow` fast path, proved equal to the `accStep` fold via new exported step equations (`accAddWord_eq_inline`, `accStep_eq_inline`) and loop bridges (`delayedDotLoop_eq_foldl`, `delayedDotRun_eq_loop`); `barrettBaseMul` transposes `N` once instead of materializing a strided column per output entry; `radixResidue` computes `2^64 % p` in machine words (`(0 - p) % p`) instead of a bignum `Nat` reduction per dot product (`toNat_radixResidue` re-proved, same statement); and the config cutoff moves to the delayed kernel's own measured crossover of 128. The comparison driver was also silently broken — the compiler floated the pure `let r := cmpRun …` past the closing timestamp, leaving the two `lean_io_mono_nanos_now()` calls adjacent in the generated C — so a `@[noinline]` IO identity now pins the computation inside the timed region; the fixed driver first reproduced the issue's 9–15× slowdown for the old kernel, and the new kernel then measures (best-of-iters wall time, `p = 2147483647`, checksums identical between kernels at every dimension):

| n | default | delayed (before, #8686) | delayed (after) | speedup after (default/delayed) |
|---|---|---|---|---|
| 64 | 0.88 ms | 15.44 ms | 0.47 ms | 1.88× |
| 96 | 4.04 ms | 49.81 ms | 1.25 ms | 3.22× |
| 128 | 8.27 ms | 105.14 ms | 5.35 ms | 1.55× |
| 160 | 14.63 ms | 200.14 ms | 8.74 ms | 1.67× |
| 192 | 33.91 ms | 341.11 ms | 13.72 ms | 2.47× |
| 256 | 69.04 ms | 750.76 ms | 47.98 ms | 1.44× |

The committed JSON and figure are regenerated from the final code, the docstrings now tell the affirmative honesty-constraint-(b) story, and `strassenDemo` stays as the minimal plug-in example. Full `lake build` (4158 jobs, including the Mathlib bridge layers) and `HexConformance` are green; `check_benches_mathlib_free` passes; no `sorry`/`axiom`/`native_decide`.

Closes #8687.

🤖 Prepared with Claude Code


kim-em:
Review round (Claude session + Codex second opinion): all proofs confirmed on both reviews — the `delayedDotRun_eq_loop` invariant (`count + m < window + i`, sufficient with `barrettWindow` opaque), the flush branch's `(reduced, 0, 0)` reset matching `accStep`, the `radixResidue` wraparound argument at both edges (`p = 2`, `p` near `2^32`), and the `cmpForce` timing fix. One medium finding applied: the shipped configs differ in cutoff (96 vs 128) as well as base kernel, so the comparison artifacts were conflating the two effects — the JSON/figure now carry matched-cutoff control series (each kernel at the other config's cutoff), and the controls confirm the base-kernel effect stands on its own: the delayed kernel wins 1.4–1.9× with both configs at cutoff 96 and 1.6–2.2× with both at 128. Artifacts reworded as a shipped-config comparison with kernel-isolating controls; docstring history notes trimmed per repo doctrine. Also for the record: the timing-harness float this PR fixed was introduced by the reviewer's sink refactor in the #8686 review round — the numbers committed in #8686 predate that edit and were sound.

PR #8702, opened 2026-07-10
perf(hex-matrix): flat row-major Vector R (n*m) backing for Matrix

kim-em:
This PR switch `Hex.Matrix` from the row-of-rows `Vector (Vector R m) n` backing to a single flat row-major `Vector R (n*m)` buffer behind the unchanged opaque API, settling the layout decision in the SPEC (row-major: the elimination consumers' row operations and `getRow` dominate and stay contiguous and in-place; `mulImpl` keeps its one-time transpose; quadrant sub-blocks become offset-and-stride windows). Every public lemma statement is unchanged — the migration's proof repairs across HexMatrix, HexRowReduce, HexDeterminant(+Mathlib), HexBareiss(+Mathlib), HexGramSchmidt(+Mathlib), HexLLL(+Mathlib), HexBerlekamp(ZassenhausMathlib) and HexMatrixMathlib are proof-bodies-only, routed through the public entry lemmas (`getElem_pair_eq_nested` bridges the pair/nested reads, no longer definitionally equal). The measured before/after overlay on the multiply bench is honestly **neutral** (naive `1.00–1.02×`, default-config `mulStrassen` `0.98–1.00×` across `64…1024`, identical checksums; committed export + overlay figure), with a small `1–4%` Bareiss elimination cost shrinking with `n` — parity is the expected reading since the multiply hot paths were already row-contiguous, and the payoff this migration buys is the copy-free `BlockView` recursion, deferred per the issue's split clause with the continuation design posted on #8652 (which stays open for that final deliverable). Whole-graph `lake build` green (4160 jobs), conformance green, zero `sorry`/`axiom`/`native_decide`.

🤖 Prepared with Claude Code


kim-em:
Review round (Claude session + Codex second opinion): no correctness findings — the flat-index arithmetic (including the m = 0 edges), the preserved `ext`/API semantics, and the measured-neutrality story all check out; the whole-graph build, conformance, and identical bench checksums across representations are the end-to-end evidence. Fixes applied from the Codex pass: (1) `getElem_pair_nat` had silently become a flat-buffer statement — the one representation leak in the public lemma surface — now restored to the rows-observation form with the flat read as a private `getElem_pair_data`; (2) the `colAdd`/`colAddRight`/`colSwap` docstrings and the SPEC no longer claim in-place column writes (they materialize via `mapRows`; flat per-entry column engines join the BlockView follow-up); (3) the read-before-write helpers' linearity claims softened to what the runtime guarantees. Measured outcome, for the merge decision: multiply surface neutral (naive 1.00–1.02×, Strassen 0.98–1.00×), Bareiss +1–4% shrinking with n — the payoff is the architectural unblocking of the copy-free BlockView recursion tracked on #8652.

PR #8720, opened 2026-07-11
feat(hex-roots): add Newton-Kantorovich atom witnesses

kim-em:
This PR adds `HexRoots/Kantorovich.lean`, the Newton-Kantorovich atom certificate for the complex root isolator. `nkWitnessCheck` computes, from the exact Taylor coefficients `cs = taylor p s.center`, the recomputed reciprocal `w = conj(c₁)·Dyadic.invAtPrec (normSq c₁) q` (pinned precision `q = 8 + max 0 (ceilLog2 (normSq c₁))`), the residuals `dₖ = w·cₖ`, and the exact dyadic bounds `y = lo(d₀)`, `z₁ = hi(1 − d₁)`, and the radial Lipschitz bound `z₂ = 2·Σ_{k=2}^{n} k·hi(dₖ)·ρ^{k−2}` with `ρ = s.radiusHi`, then returns the three strict exact-dyadic comparisons `0 < normSq c₁ ∧ y + z₁·r + z₂·r²/2 < r ∧ z₁ + z₂·r < 1`. It exposes `nkWitness` with its `Decidable` instance, the disjunctive `atomWitness`, the `AtomStrategy` selector, the `DyadicRootIsolation` and `Certified` output types, and `DyadicRootCluster.atomize` repackaging a certified `k = 1` cluster as an atom through the Pellet disjunct. The `z₂` sum is a single `List.range` fold carrying a running power (no `pow` calls in the loop), following the kernel-reducible idiom of `pelletAt`, and coefficient access uses `cs.getD i (0,0)`.

One SPEC deviation surfaced and is reported rather than papered over: `nkWitness` is `Decidable` and every runtime construction site (`if h : nkWitnessCheck p s = true then ...`) works, but `nkWitness ... := by decide` does **not** kernel-reduce. The witness embeds `Dyadic.invAtPrec` (SPEC "Newton-Kantorovich atom witnesses"), which routes through `Rat.inv`/`Rat.toDyadic`; those normalise via `Nat.gcd`, a well-founded recursion the kernel does not reduce. The Pellet witnesses (pure `Dyadic`) do kernel-reduce, so this is specific to the NK route. It does not affect soundness or the runtime isolator (proofs are obtained at runtime from the `= true` equality, not by `decide`), but it contradicts the SPEC's blanket "dischargeable by `decide`" claim for this one certificate form. No helper defs beyond the SPEC were introduced.

🤖 Prepared with Claude Code

### DIFF
--- a/HexRoots.lean
+++ b/HexRoots.lean
@@ -8,6 +8,7 @@ module
 
 public import HexRoots.Basic
 public import HexRoots.Cauchy
+public import HexRoots.Kantorovich
 public import HexRoots.MahlerPrec
 public import HexRoots.Pellet
 public import HexRoots.Taylor
--- a/HexRoots/Kantorovich.lean
+++ b/HexRoots/Kantorovich.lean
@@ -22,7 +22,7 @@ Newton-Kantorovich hypotheses exact-dyadic comparisons rather than error
 budgets.
 
 With `A` the approximate inverse of `p'(m)` built from the exact reciprocal
-`w = conj(c₁)·invAtPrec (normSq c₁)` of `c₁`, and `dₖ = w·cₖ`, the witness data
+`w = conj(c₁)·invFloor (normSq c₁)` of `c₁`, and `dₖ = w·cₖ`, the witness data
 is `y = lo(d₀)` (the sup norm of the Newton residual), `z₁ = hi(1 − d₁)` (the
 sup-operator norm of `1 − A∘p'(m)`), and the radial Lipschitz bound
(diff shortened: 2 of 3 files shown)


kim-em:
Follow-up commit: replaced `Dyadic.invAtPrec` inside `nkWitnessCheck` with a kernel-reducible `invFloor` (one integer division; agrees with `invAtPrec` on positive arguments, same rounding contract). The SPEC-flagged kernel-`decide` gap is resolved: `example : nkWitness p₁ ⟨1,0,3⟩ := by decide` elaborates in under a second, and the SPEC text on this branch now documents the floor form. All sanity results are bit-identical.

PR #8856, opened 2026-07-22
refactor(hex-poly): hoist kernel-reducible polynomial equality to DensePoly.beqCoeffs

kim-em:
This PR generalize the kernel-reducible coefficient-equality checker from HexRealRoots' `ZPoly.beqCoeffs` to `DensePoly.beqCoeffs` over any coefficient type, adding `eq_of_beqCoeffs` and `beqCoeffs_iff_eq`, and refactor `HexRealRoots.Cert` onto it via an `abbrev`. This is step 1a of the `factor_poly`/`irreducibility` tactic plan: the tactics will emit `C scalar * factors.prod = p` product checks for both `ZPoly` and `FpPoly p` literals, and those must kernel-reduce under plain `public import` — the derived structural `DecidableEq (DensePoly R)` delegates to `Array.instDecidableEqImpl`, whose body is unavailable downstream under the module system (see `progress/lean4-array-decidableeq-module-repro.md`). Verified by building HexPoly, HexRealRoots, HexRealRootsMathlib (including the plain-import Sturm replay tests), and the Berlekamp/BZ free libraries, plus plain-import scratch probes of the ZPoly and FpPoly product-check shapes.

🤖 Prepared with Claude Code

PR #8857, opened 2026-07-22
feat(bz): kernel-decidable irreducibility entry points for FpPoly and ZPoly

kim-em:
This PR add the decide-slot lemmas the `factor_poly`/`irreducibility` elaborators will apply to reified literal data (plan step 1, stacked on #8856): `Hex.Berlekamp.checkMonicCert` — a monic-guarded wrapper over the incremental Rabin checker using the `checkCertAtFactor` dite idiom so reified applications never construct a `DensePoly.Monic` proof term — with soundness `irreducible_of_checkMonicCert` and the scaled entry point `irreducible_of_checkMonicCert_scale` for non-monic `FpPoly` inputs; `Hex.ZPoly.irreducible_of_modPCert`, the single-prime kernel-decidable route to `ZPoly.Irreducible` composing the Gauss-style mod-p transfer with the monic-normalized modular certificate; and the public `Hex.ZPoly.irreducible_of_linear` decide-slot wrapper over the private size-two lemma in `IrreducibleCore`. Every hypothesis is a Boolean check on literal data, so reified applications discharge each slot with `Eq.refl true`/`Eq.refl false` and the kernel verifies by reduction alone — the factorizer never runs in the kernel. The full slot composition (certificate generation plus all six slots) was runtime-probed green on `x²+1` at `p = 3`, and `lake build HexBerlekampZassenhausMathlib` (the CI-gating target) completes with zero errors.

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekamp.lean
+++ b/HexBerlekamp.lean
@@ -11,6 +11,7 @@ public import HexBerlekamp.DelayedKernel
 public import HexBerlekamp.DistinctDegree
 public import HexBerlekamp.Factor
 public import HexBerlekamp.Irreducibility
+public import HexBerlekamp.IrreducibleDecide
 public import HexBerlekamp.RabinSoundness
 
 public section
--- a/HexBerlekamp/IrreducibleDecide.lean
+++ b/HexBerlekamp/IrreducibleDecide.lean
@@ -1,0 +1,78 @@
+/-
+Copyright (c) 2026 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
+-/
+
+module
+
+public import HexArith.Nat.Prime
+public import HexModArith.Prime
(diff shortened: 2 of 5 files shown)


kim-em:
Codex second opinion: one High finding, dismissed as already staged later in this stack. Codex correctly observed that `checkMonicCert` is `@[expose]` while its callee closure (`checkIrreducibilityCertificateLinearIncremental`, `toAmbient?`, the Rabin/pow-chain helpers) is not yet exposed, so a plain-import `decide` on a literal certificate is stuck at this commit. That exposure is added deliberately in #8859 (which annotates exactly that closure and adds `FactorTacticTests.lean` exercising the reduction end to end), merging in the same train immediately after; nothing on main consumes these entry points in between. No soundness flaw: the proof bridges were confirmed valid, and no sorry/axiom/native_decide.

PR #8858, opened 2026-07-22
refactor(bz): move certificate reifiers into the Mathlib-free layers

kim-em:
This PR move the certificate reifiers out of the Mathlib bridge into the Mathlib-free layers (plan step 2, stacked on #8857): `HexBerlekampZassenhausMathlib/CertReify.lean` imported only free code plus `Lean`, so it was Mathlib-free code stranded in the bridge by accident. It splits into `HexBerlekamp/CertReify.lean` (ZMod64/FpPoly/Bounds/Rabin-certificate reifiers, the serialized round-trip views, and the `reflTrue`/`reflFalse` slot proofs) and `HexBerlekampZassenhaus/CertReify.lean` (ZPoly and the `ZPolyIrreducibilityCertificate` tower), renamespaced to `Hex.CertReify` so the upcoming `factor_poly`/`irreducibility` elaborators in the free layers can use them; `irreducible_cert` and its round-trip tests are repointed. Verified by building both free libraries, the repointed tactic and test modules, and the full CI-gating `HexBerlekampZassenhausMathlib` target, all with zero errors.

🤖 Prepared with Claude Code

PR #8859, opened 2026-07-22
feat(berlekamp): factor_poly and irreducibility elaborators for FpPoly

kim-em:
This PR add the certificate-backed `factor_poly` and `irreducibility` term elaborators and tactics to HexBerlekamp (plan step 3, stacked on #8858), handling `FpPoly p` natively for literal prime moduli. `factor_poly f` elaborates to an `FpPoly.Factored f` (`scalar`, `factors` with repetition, `factors_mul : C scalar * factors.prod = f`, `factors_irred`), usable as `obtain ⟨scalar, factors, factors_mul, factors_irred⟩ := factor_poly f`; its tactic form introduces `scalar`/`factors` as transparent `let`s plus the two hypotheses. `irreducibility f` produces `FpPoly.Irreducible f` as a term, a named or anonymous hypothesis, or closes the goal in bare form. The compiled Yun square-free decomposition, Berlekamp factorizer, and Rabin certificate generator run at elaboration time as untrusted search; emitted terms carry only kernel checks on reified literal data — a `beqCoeffs` product check and one `checkMonicCert`/`checkIrredCover` certificate replay per distinct factor, budget-guarded at `deg·p ≤ 2²⁶` — so the factorizer never runs in the kernel. Non-`FpPoly` inputs dispatch through the new `Hex.FactorTactic.Provider` hook: providers are probed from the environment by well-known name with type and ABI-version checks, so HexBerlekampZassenhaus and the Mathlib bridges extend the tactics without any import in this direction.

Two enabling fixes: `DecidableEq (DensePoly R)` now routes through the coefficient `List` so polynomial equality kernel-reduces under plain module imports (the `Array.instDecidableEqImpl` exposure gap, `progress/lean4-array-decidableeq-module-repro.md`), and the incremental Rabin checker chain gains the `@[expose]` its kernel replay needs without `import all`.

`HexBerlekamp/FactorTacticTests.lean` covers both forms end-to-end under plain `public import` only — non-monic, non-square-free, characteristic-2 inseparable, zero/constant/unit inputs, `#guard_msgs` negative tests (composite modulus, zero, unit, reducible), and `#print axioms` guards showing clean cones. Full `lake build` (all Mathlib bridges, 9435 jobs) is green.

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekamp.lean
+++ b/HexBerlekamp.lean
@@ -8,6 +8,11 @@ module
 
 public import HexBerlekamp.Basic
 public import HexBerlekamp.CertReify
+public import HexBerlekamp.Factored
+public import HexBerlekamp.TacticCore
+public import HexBerlekamp.FactorPolyElab
+public import HexBerlekamp.IrreducibilityElab
+public import HexBerlekamp.FactorTacticTests
 public import HexBerlekamp.DelayedKernel
 public import HexBerlekamp.DistinctDegree
 public import HexBerlekamp.Factor
--- a/HexBerlekamp/FactorPolyElab.lean
+++ b/HexBerlekamp/FactorPolyElab.lean
@@ -77,13 +77,18 @@ meta def fpFactorSearch (p : Nat) [Hex.ZMod64.Bounds p]
     (f.coeff 0, [])
   else
     let dec := Hex.FpPoly.squareFreeDecomposition hp f
-    let factors := dec.factors.flatMap fun sf =>
+    let raw := dec.factors.flatMap fun sf =>
       let parts :=
         if hmonic : Hex.DensePoly.leadingCoeff sf.factor = 1 then
           (Hex.Berlekamp.berlekampFactor sf.factor (by exact hmonic)).factors
         else [sf.factor]
       parts.flatMap fun part => List.replicate sf.multiplicity part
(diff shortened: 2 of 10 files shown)


kim-em:
Codex second opinion: one genuine High finding, fixed in 75aca16a. `fpFactorSearch` passed raw Berlekamp gcd leaves (defined only up to a unit scalar, per Factor.lean's own docs) straight to `fpCoverEntries`, so `factor_poly` aborted on valid inputs — reproduced with x^4 + x^3 + x over F_3. Fixed by normalizing each leaf via `FpPoly.normalizeMonic` and folding the scalars into the unit (untrusted search code; the emitted kernel product check still guards the result), plus an F_3 regression test in FactorTacticTests. Two findings dismissed: (1) per-factor replay budget vs cumulative — aggregate kernel work is bounded by the per-factor cap times the factor count (itself bounded by the input degree), and the cap's purpose is rejecting single pathological replays, not total-work accounting; (2) the umbrella publicly importing `FactorTacticTests` follows the existing `HexRealRoots.ReplayTest` regression-lock convention — the fixtures are namespaced and the leaked `Bounds` instances are decidable-true size facts, harmless downstream. No sorry/axiom/native_decide; certificate soundness confirmed.

PR #8860, opened 2026-07-22
feat(bz): ZPoly provider for factor_poly and irreducibility

kim-em:
This PR register the `Hex.ZPoly` arms of `factor_poly`/`irreducibility` (plan step 4, stacked on #8859) through the provider hook: with HexBerlekampZassenhaus imported, `factor_poly f` elaborates to a `ZPoly.Factored f` — scalar = signed content, primitive positive-leading-coefficient factors with repetition, product and per-factor irreducibility carried by kernel checks on reified literals — and `irreducibility` handles `ZPoly` terms and `ZPoly.Irreducible` goals. Certification goes through the new `ZPoly.IrredWitness` inductive (prime constant / primitive linear / single-prime modular Rabin certificate, packed with its own modulus in the `PrimeFactorData` idiom) with the bulk `ZPoly.checkIrredCover` checker, one witness per distinct factor; the compiled `ZPoly.factorize` runs only at elaboration time as untrusted search. Factors that are irreducible over ℤ but reducible mod every candidate prime (balanced: Swinnerton-Dyer, `X⁴+1`) have no single-prime witness, so the provider declines with a diagnostic that defers to the Mathlib bridge's multi-prime certificates — tested via `#check_failure` on `X⁴+1` and `(x+1)·(x⁴−10x²+1)`. Tests cover content/sign/multiplicity inputs, all tactic forms, reducible/unit error messages, and clean `#print axioms` cones, under plain `public import` only. Full `lake build` (9442 jobs) is green.

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekamp/FactorPolyElab.lean
+++ b/HexBerlekamp/FactorPolyElab.lean
@@ -222,24 +222,35 @@ syntax (name := factorPolyTac) "factor_poly" term:max : tactic
     | `(tactic| factor_poly $t) => do
         let e ← Tactic.withMainContext do
           elabFactorPolyCore stx t none
-        -- Destructure the emitted `Factored.mk` application.
+        -- Destructure the emitted `Factored.mk` application (the FpPoly
+        -- shape carries the modulus and instance; providers may emit other
+        -- `*.Factored.mk` shapes with the same final four fields, e.g.
+        -- `Hex.ZPoly.Factored.mk`).
         let args := e.getAppArgs
-        unless e.getAppFn.isConstOf ``Hex.FpPoly.Factored.mk &&
-            args.size == 7 do
-          -- Provider-produced results: fall back to a plain `have`.
+        let fields? :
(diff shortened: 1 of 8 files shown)


kim-em:
Codex second opinion: four findings, one fixed here. Fixed (4d393d9c): the `primeConst` witness kernel-replayed `isNatPrime c` (roughly c steps via `List.range c`) with no budget guard — kernel checking is not heartbeat-interruptible, so `irreducibility` on a large prime constant could hang; the witness search is now guarded by `replayBudget` with a targeted over-budget error and a #guard_msgs regression test. Dismissed: (1) the non-monic Berlekamp leaf abort is already fixed on #8859 (75aca16a); (2) per-factor vs cumulative replay budget — aggregate work is bounded by the per-factor cap times the factor count (bounded by input degree), and the cap exists to reject single pathological replays; (3) test modules in the production umbrellas follow the established `ReplayTest` regression-lock convention, with namespaced fixtures and harmless decidable-true `Bounds` instances. No soundness hole; no sorry/axiom/native_decide.

PR #8861, opened 2026-07-22
feat(berlekamp-mathlib): Polynomial (ZMod q) provider for factor_poly and irreducibility

kim-em:
This PR add the `Polynomial (ZMod q)` provider for the `factor_poly`/`irreducibility` elaborators, the third link in the stacked provider chain (FpPoly base → ZPoly → this Mathlib bridge). `HexBerlekampMathlib/FactorProvider.lean` declares `HexBerlekampMathlib.FactorTactic.provider` (the second probe name in `Hex.FactorTactic.providerNames`), handling inputs with a literal prime modulus below 2^31 and declining composite or oversized moduli with a diagnostic. Its core is a parser-with-proof: a meta recursion over the input expression (`X`, `Polynomial.C`, numerals, `+`, `-`, `*`, negation, `^` with literal exponents, named defs under a fuel guard) that evaluates each node to an executable `Hex.FpPoly q` while chaining the named `toMathlibPolynomial_*` transport lemmas into a bridge equation `toMathlibPolynomial fLit = P`; scalar coefficient equalities are certified by `decide` on `ZMod64.toZMod` values (whnf-checked before emission), and one `DensePoly.eq_of_beqCoeffs` kernel check recombines the flat reified literal at the root.

The non-meta half, `HexBerlekampMathlib/FactorPoly.lean`, introduces the polymorphic result structure `Hex.FactoredPoly P` (the `Polynomial`-level counterpart of `FpPoly.Factored`, shared with the upcoming `Polynomial ℤ` provider) plus the transport lemmas `nat_prime_of_hex`, `toMathlibPolynomial_one`/`_neg`/`_listProd` and the all-Boolean one-shot assemblers `Hex.FactoredPoly.ofFp` and `HexBerlekampMathlib.irreducible_ofFp`: every certification slot is a Boolean check on reified literals discharged by `Eq.refl true`/`Eq.refl false`, so the factorizer and certificate generator never appear in emitted terms. The factor search reuses `Hex.FactorTactic.fpFactorSearch`/`fpCoverEntries` as untrusted search with runtime self-checks. `HexBerlekampMathlib/FactorPolyTests.lean` exercises term/obtain/tactic `factor_poly` forms, `irreducibility` term/named/goal forms, coefficients over the modulus, negative coefficients, composite-modulus rejection pinned by `#guard_msgs`, and `#print axioms` guards. Both `HexBerlekampMathlib` and the CI-gating `HexBerlekampZassenhausMathlib` build green.

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekampMathlib.lean
+++ b/HexBerlekampMathlib.lean
@@ -7,6 +7,9 @@ Authors: Kim Morrison
 module
 
 public import HexBerlekampMathlib.Basic
+public import HexBerlekampMathlib.FactorPoly
+public import HexBerlekampMathlib.FactorProvider
+public import HexBerlekampMathlib.FactorPolyTests
 
 public section
 
--- a/HexBerlekampMathlib/FactorPoly.lean
+++ b/HexBerlekampMathlib/FactorPoly.lean
@@ -1,0 +1,148 @@
+/-
+Copyright (c) 2026 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
+-/
+
+module
+
+public import HexBerlekampMathlib.Basic
+public import HexBerlekamp.IrreducibleDecide
+
+public section
+
+/-!
+The Mathlib-side result type of `factor_poly` for `Polynomial R` inputs, and
+the kernel-decidable assemblers the `Polynomial (ZMod q)` provider emits.
+
+`FactoredPoly P` is the `Polynomial`-level counterpart of
+`Hex.FpPoly.Factored` / `Hex.ZPoly.Factored`. The assemblers `FactoredPoly.ofFp`
+and `irreducible_ofFp` take only Boolean checks on reified executable literals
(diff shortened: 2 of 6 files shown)


kim-em:
Codex second opinion: four findings, one fixed here. Fixed (431578cd): `withZModInput` ran `isPrimeTrial q` (Theta(q) via `List.range q`, both compiled and in the emitted kernel slot) before the ZMod64 bound check, so an oversized modulus could hang elaboration; the bound and replay-budget checks now come first, with a targeted decline for over-budget moduli. Dismissed: (2) linear expansion of `a ^ n` at parse time is elaborator-side work under the heartbeat limit, not a kernel hang — standard failure mode for literal-driven elaborators; (3) the tactic form landing bridge-provider results as a single `factored` hypothesis is the documented behavior (see the comment in FactorPolyTests.lean) for `FactoredPoly.ofFp` assemblers, a deliberate API choice; (4) `FactorProvider` not importing `IrreducibilityElab` is moot — the umbrella `HexBerlekampMathlib` is the supported import surface and pulls in both. No soundness bypass; no sorry/axiom/native_decide.

PR #8862, opened 2026-07-22
feat(bz): Eisenstein-after-shift irreducibility certificates

kim-em:
This PR adds Eisenstein-after-shift certificates to the free-layer `ZPoly` irreducibility pipeline (Milestone E of the factor_poly/irreducibility roadmap). The new `HexBerlekampZassenhaus/EisensteinCore.lean` provides the executable Taylor shift `ZPoly.translate` (an `@[expose]` Horner fold, so its computation is part of the kernel check) with its full Mathlib-free transfer algebra — `translate_add`, `translate_mul`, `translate_C`, the inverse identity `translate_neg_translate`, `size_translate`, `isUnit_translate_iff`, and `irreducible_of_translate_irreducible` — proved by strong induction over an `X`-quotient decomposition. On top of that it proves the elementary coefficient-level Eisenstein criterion `ZPoly.irreducible_of_eisenstein` (diagonal product-coefficient formula, first-unbroken-index extraction, Euclid's lemma for `Hex.Nat.Prime` transported to `Int` via `natAbs`) and packages it as the decide-slot theorem `irreducible_of_eisensteinCert` on the shifted literal; primitivity is checked on the shifted polynomial directly, so no content-preservation lemma is needed, and divisibility slots use `%`-checks since the free layer carries no `Decidable` instance for `Int` divisibility.

`ZPoly.IrredWitness` gains an `.eisenstein (q : Nat) (shift : Int)` constructor with the corresponding kernel-decidable `checkIrredWitness` arm and soundness case, and the `FactorProvider` witness search tries shifts `0, ±1, ±2, ±3` (candidate primes from the small prime divisors of the shifted constant term, each candidate validated by re-running `checkIrredWitness`) after the single-prime modular search declines. `X⁴+1` is now certified by the free layer at shift `1`, prime `2`: the balanced-decline test becomes `theorem x4p1_irred : ZPoly.Irreducible x4p1 := irreducibility x4p1` with a `#print axioms` guard, and a plain shift-`0` witness is exercised on `X²−2` through the checker and its soundness theorem. Swinnerton-Dyer inputs remain a documented decline (they are not shift-Eisenstein); the SPEC partiality paragraph and the decline diagnostics are updated to match. `lake build` and `lake build HexBerlekampZassenhausMathlib` are green; no `sorry`, no `axiom`, no `native_decide`.

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekampZassenhaus.lean
+++ b/HexBerlekampZassenhaus.lean
@@ -17,6 +17,7 @@ public import HexBerlekampZassenhaus.BhksCandidates
 public import HexBerlekampZassenhaus.BhksRecover
 public import HexBerlekampZassenhaus.Recombination
 public import HexBerlekampZassenhaus.FactorEntryPoints
+public import HexBerlekampZassenhaus.EisensteinCore
 public import HexBerlekampZassenhaus.IrreducibleCore
 public import HexBerlekampZassenhaus.IrreducibleDecide
 public import HexBerlekampZassenhaus.Factored
--- a/HexBerlekampZassenhaus/EisensteinCore.lean
+++ b/HexBerlekampZassenhaus/EisensteinCore.lean
@@ -1,0 +1,977 @@
+/-
+Copyright (c) 2026 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
+-/
+
+module
+
+public import HexArith.Nat.Prime
+public import HexBerlekampZassenhaus.IrreducibleCore
+
+public section
+set_option backward.proofsInPublic true
+
+/-!
+Eisenstein's criterion for `Hex.ZPoly`, plus the translation (Taylor shift)
+`f(X) ↦ f(X + s)` needed to certify polynomials that only become Eisenstein
+after a shift (e.g. `X⁴ + 1` at shift `1`, prime `2`).
+
+The file provides three layers:
+
+* `ZPoly.translate`, an executable, kernel-reducible Horner fold computing
+  `f(X + s)`, with its ring-homomorphism laws (`translate_add`,
+  `translate_mul`, `translate_C`) and the inverse identity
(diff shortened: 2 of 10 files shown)


kim-em:
Rebased onto the provider stack: this branch now sits on `factor-poly/int-provider` (PR #8863) instead of `factor-poly/bz-provider`, and the base is retargeted accordingly. The rebase itself was conflict-free; a follow-up commit reconciles the Mathlib-bridge expectations with the new Eisenstein certificates: the `X⁴+1` decline-diagnostic test in `HexBerlekampZassenhausMathlib/FactorPolyTests.lean` becomes a positive `irreducibility` test with a `#print axioms` guard (the free layer now certifies it at shift 1, prime 2, before either bridge certificate language runs), the bridge decline diagnostic keeps its coverage via the genuinely uncertifiable Swinnerton-Dyer quartic `x⁴ − 10x² + 1`, and the bridge provider docstring/decline message and SPEC stop citing `X⁴+1` as an uncertifiable example. Full `lake build` is green (9448 jobs, zero errors).

kim-em:
Codex second opinion: one genuine finding, fixed in bf2891f3. The Eisenstein search admitted witness primes up to 10^6, but the emitted certificate kernel-replays `isPrimeTrial q`, whose `List.range q` reduction is recursion-depth linear in q and stalls at the default `maxRecDepth` a little above 150 (reproduced: `eisenstein 251` fails `rfl`/`decide`) — so the search could emit proofs the kernel cannot check. Candidate primes are now capped at `eisensteinPrimeCap = 128`, with the boundary replay (prime 127) locked by a `rfl` test; over-cap constants simply produce no Eisenstein candidates and fall through to the existing clean decline. Dismissed: the "not Eisenstein at any small shift" decline phrasing (search bounds are documented at the definition site, and the exact message is locked by downstream #guard_msgs tests; rewording is not worth the cascade). No sorry/axiom/native_decide; the soundness chain is intact.

Issue #8669, opened 2026-07-08
perf(bz): de-box the generic ZMod64 row-reduction/nullspace under fixedSpaceKernel


kim-em:
Deboxes the generic `Matrix.nullspace` / row-reduction under `Berlekamp.fixedSpaceKernel` into a `ZMod64`-monomorphic kernel that stays in unboxed `Array UInt64` end to end, with a correspondence theorem to the generic `nullspace` so the public factorization result is unchanged. This is the sole material lever surfaced by the #8647 measurement.

**Evidence.** Profiling `RELIFT_PROFILE=prime` under `perf` on current `main` (measurement report `reports/bz-nullspace-inv-cost-shares.md`, filed for https://github.com/kim-em/hex-dev/issues/8647) shows the generic `Matrix.nullspace` + row-reduction under `fixedSpaceKernel` is now the single largest algorithmic share of the profile: ~15-22% (folded-stack inclusive ~21%, compiled-symbol self-time ~19%), a large relative rise over the earlier deg-24 ~8% baseline. It grew because #8633/#8638 made the field arithmetic word-fast while leaving the generic boxed linear-algebra machinery untouched.

**The cost is representation, not arithmetic.** `ZMod64.mul`/`add` are ~6% combined and already word-fast. What remains is the generic boxed `Matrix R n m`: boxed `ZMod64` elements dispatched through the `Lean.Grind.Ring R` typeclass so every element op crosses a `lean_apply_1/2` closure boundary (~26% of the profile), `Array.ofFn`-of-`Array.ofFn` matrix/vector construction (~11% higher-order machinery), boxed-cell allocation/refcount (~22% allocator), and `List`-based free-column extraction (`freeColsList`/`filterTR_loop`). These generic buckets are shared with the mod-p `DensePoly.gcd` sweep, so only part is nullspace-attributable, but the nullspace subtree itself is ~20%.

**Not the packed kernel of #8634.** #8634 (packed nullspace kernel) was tried and regressed because it materialized the boxed `Matrix` at its edges and paid the same per-multiply bignum allocation that #8633 later removed. The acceptance criterion here, stated negatively: **no boxed `Matrix R n m` conversion anywhere on the hot path**, before or after row-reduction. The packed `Array UInt64` representation must flow from the Berlekamp matrix construction through echelon form, nullspace basis, and witness polynomial extraction with no pack/unpack boundary. The lever is boundary-free monomorphization, not "packed nullspace over a boxed matrix."

[...]

Issue #8888, opened 2026-07-27
perf(berlekamp): profile and reduce witness splitting


kim-em:
**Measured by #8700 / #8879.** After hoisting the two repeated fixed-space computations, recursive witness splitting is now the dominant executable phase of `Berlekamp.berlekampFactor` on the fully split degree grid.

| input | deg | p | modular factors | fixed `berlekampFactor` | matrix | nullspace | witness splitting |
|---|--:|--:|--:|--:|--:|--:|--:|
| split h2 | 6 | 7 | 6 | 54 us | 20.2% | 3.5% | 76.3% |
| split h2 | 10 | 11 | 10 | 174 us | 13.4% | 2.3% | 84.3% |
| split h2 | 14 | 17 | 14 | 390 us | 8.8% | 1.8% | 89.4% |
| split h2 | 18 | 19 | 18 | 765 us | 8.0% | 1.5% | 90.6% |
| split h2 | 22 | 23 | 22 | 1247 us | 6.8% | 1.2% | 92.0% |
| split h2 | 24 | 29 | 24 | 1598 us | 6.6% | 1.1% | 92.3% |
| phi15 | 8 | 7 | 2 | 57 us | 52.9% | 29.5% | 17.6% |
| SD3 | 8 | 7 | 4 | 82 us | 34.6% | 8.8% | 56.6% |
| SD4 | 16 | 11 | 8 | 502 us | 27.7% | 8.7% | 63.7% |

Full method and results: https://github.com/kim-em/hex-dev/blob/main/reports/berlekamp-factor-phase-profile.md

Raw samples: https://github.com/kim-em/hex-dev/blob/main/reports/bench-results/berlekamp-factor-phase-profile.csv

The current bucket is `fullySplit` -> `splitWithWitnesses?` -> `kernelWitnessSplit?` in `HexBerlekamp/Factor.lean`. For each current factor and witness, `kernelWitnessSplit?` first computes `(witness % f)` for the constant guard. If nonconstant, `kernelWitnessSplitAux` scans `c = 0, ..., p - 1`, computing `DensePoly.gcd f (witness - FpPoly.C c)` until it finds a proper factor. A successful split also computes `f / factor`, then recursively repeats witness search on both children.

**Task.** Profile this witness-splitting bucket on the same degree 6--24 split grid plus phi15, SD3, and SD4. Attribute at least:

1. witness reduction and the constant guard `(witness % f)`;
[...]

Issue #8635, opened 2026-07-06
perf(bz): ZMod64 inverse in word arithmetic, removing the Int.extGcd/gmp round-trip


kim-em:
**Context.** Deliverable 4 of #8630. Deliverables 1 and 2 landed in that issue's PR. This carves out the word-arithmetic modular inverse.

**Measurement.** `ZMod64.inv` (`HexModArith/Basic.lean:542`) routes through `HexArith.Int.extGcd` (gmp), showing up as `lean_int_extgcd_fallback` + gmp on the modular-inverse path in the deg-24 profile. It is called once per field division step (`O(deg)` per gcd), so smaller than the `O(deg²)` elimination win from deliverable 2, but still on the hot path (every division in `gcdPacked`/`divModRemainderPacked` and, later, nullspace pivoting).

**Deliverable.** Compute the `ZMod64` inverse in word arithmetic, eliminating the `Int.extGcd`/gmp round-trip per inverse.

**Design notes / gotchas.**
- A global `@[csimp] @ZMod64.inv = @invImpl` must hold for **every** `Bounds p` and every input, including non-prime `p` and non-coprime `a`, where `inv` currently returns the `Int.extGcd`-derived Bezout residue `ofNat p (Int.toNat (s % p))`. So **Fermat powering `a^(p-2)` is not a valid global swap** (it only matches for prime `p`, coprime `a`).
- `HexArith.UInt64.extGcd` (`HexArith/ExtGcd.lean:499`) does **not** help on its own: it still calls `Int.extGcd` internally for the cofactors. A genuine win needs a binary/extended GCD implemented in `UInt64` (wrapping/widening) arithmetic, proved equal to the current `inv` residue for all `Bounds p` (the `extGcd_bezout`/`extGcd_fst` certificates are a starting point). Alternatively an `@[extern]` C kernel (trusted) matching the exact residue convention.

**Verification.** `@[csimp]`-backed equality (or extern with a matching-residue argument); FLINT conformance byte-identical; bench targets show the win.

kim-em:
[...]

PR #8892, opened 2026-07-27
perf(bz): de-box the nullspace assembly under fixedSpaceKernel

kim-em:
This PR rewrites the executable `IsRowReduced.nullspace` assembly from an O(m³) reference (a per-entry `pivotIndex?` rescan building `nullspaceMatrix`, then a per-column `Matrix.col` re-read) to an O(m²) direct assembly — one `pivotRowTable` precomputation (column → pivot-row index in a single O(rank) pass) plus a direct per-column build, with the free-column vector and the pivot table hoisted out of the inner loops — and proves the two agree columnwise so the public factorization result is byte-identical and every downstream soundness/completeness lemma is unchanged.

**Honest status on materiality.** The issue's premise (the generic nullspace being the ~15-22% "sole material lever") no longer holds on current `main`: #8879 ("share fixed-space kernel construction") landed while this was in progress and already fixed the dominant inefficiency — the old assembly recomputed `nullspaceMatrix` once per column (effectively O(m⁴)); #8879 hoisted it to compute once (O(m³)). A controlled, load-matched A/B (min-of-10, interleaved, `RELIFT_PROFILE=prime`, byte-identical outputs) of this PR against current `main` shows **1.00-1.03×** across deg 8-24, and a fresh profile puts the whole nullspace subtree at **~0.1%** of the `choosePrimeData?` path (now dominated by the `DensePoly.gcd` equal-degree sweep and `ZMod64.inv`/`extGcd`). So the further O(m³)→O(m²) step is a correct asymptotic improvement (it helps for degrees beyond the deg-24 split family and removes the `pivotIndex?` rescan permanently) but is **end-to-end neutral on the current workload**.

The change is a definition change, not a `@[csimp]`: the generic `Matrix.nullspace` is specialized/inlined into the `@[expose] fixedSpaceKernelVectors`, so a `csimp` on it is bypassed by the specializer (verified by profiling), whereas a definition change is reflected in the specialized copies. Correctness rests on re-proving the single bridging lemma `nullspace_get` (`nullspace.get k = Matrix.col nullspaceMatrix k`) with the free / other-free / pivot cases matched entrywise against `nullspaceMatrix_free` / `_free_ne` / `_pivot`; `#print axioms` reports only `[propext, Classical.choice, Quot.sound]`. All builds green including the CI-gated `HexBerlekampZassenhausMathlib` bridge and `HexConformance` `#guard`s. Independently reviewed by Codex (no correctness or semantic regression).

Given #8879 already captured the material lever, this can either land as an asymptotic-cleanup-plus-proven-correspondence or be closed; deferring that call to the maintainer.

Refs #8669.

🤖 Prepared with Claude Code

### DIFF
--- a/HexRowReduce/Nullspace.lean
+++ b/HexRowReduce/Nullspace.lean
@@ -157,22 +157,139 @@ the matching pivot row and free column. -/
   simp [E.toIsEchelonForm.pivotCols_disjoint_freeCols i k,
     pivotIndex?_pivot E.toIsEchelonForm i]
 
+/-- Column → pivot-row lookup table: entry `c` holds the pivot-row index `i` when
+`c = D.pivotCols.get i`, and the sentinel `D.rank` when `c` is a free column. Built
+in one `O(rank)` pass, this replaces the per-entry `O(rank)` `pivotIndex?` scan the
+reference assembly used. Implementation helper for `nullspace`; the documented
+pivot/free reading holds when `D` is certified by an `IsEchelonForm` (whose pivot
+columns are distinct) — an arbitrary `RowEchelonData` with duplicate pivot columns
+has no such guarantee. -/
+@[expose]
+def pivotRowTable (D : RowEchelonData R n m) : Vector Nat m :=
+  Fin.foldl D.rank (fun t i => t.set (D.pivotCols.get i).val i.val (D.pivotCols.get i).isLt)
+    (Vector.replicate m D.rank)
+
+omit [Mul R] [Add R] [OfNat R 0] [OfNat R 1] in
+/-- A fold of `Vector.set`s at column indices distinct from `j` leaves entry `j`
+untouched. -/
+private theorem pivotRowTable_fold_ne (j : Fin m) :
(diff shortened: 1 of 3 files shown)


kim-em:
Closing as superseded by #8879, which already captured this issue's lever. #8879 hoisted the old per-column `nullspaceMatrix` recomputation (effectively O(m⁴)) to compute once (O(m³)); on current `main` the whole nullspace subtree is ~0.1% of `choosePrimeData?`, and a controlled load-matched A/B of this PR vs `main` is 1.00–1.03× (neutral). The further O(m³)→O(m²) rewrite here is correct and proven (`nullspace_get`, `#print axioms` clean, all green) but not material on the current workload.

The real remaining levers per the refreshed profile are tracked by open issues: the `DensePoly.gcd` equal-degree witness sweep (#8888) and word-arithmetic `ZMod64.inv`/`extGcd` (#8635, reopened). Branch `issue-8669` is left in place in case the proven O(m²) assembly + correspondence is wanted later as dormant infrastructure.

PR #8896, opened 2026-07-27
HexNumberField M3: add threaded fixed-field approximation

kim-em:
Closes #8894

Implements conservative dyadic complex-ball arithmetic, rational Horner evaluation with an input-derived guard budget, and the threaded `QAdjoin.approx` API. The checked refinement fallback preserves soundness; the companion will discharge its unreachability and requested-radius obligations.

Verification:
- `lake build HexNumberField`
- compiled ball/evaluation guards
- DAG, copyright, line-count, phase, bench-import, conformance-matrix, forbidden-form, and whitespace checks

PR #8902, opened 2026-07-27
feat(number-field): exactify lazy roots

kim-em:
## Summary
- convert canonical algebraic numbers to their fixed-field and lazy-root views
- factor and re-certify enclosing polynomials to exactify a selected lazy root
- select the matching irreducible factor geometrically and canonicalize its representative
- cover irreducible and irrelevant-factor exactification in compiled regressions

## Validation
- `lake build HexNumberField`

This is the next stacked HexNumberField milestone after #8896.


kim-em:

PR #8907, opened 2026-07-27
feat(number-field): add lazy arithmetic

kim-em:
## Summary

- construct addition and multiplication eliminants over nested `ZPoly`
- normalize square-free cores and select operation roots with certified balls
- implement lazy add/sub/mul/neg/inv/div and total wrappers
- expose canonical `AlgebraicNumber` arithmetic via lazy operations plus exactification

## Verification

- `lake build HexNumberField.Lazy HexNumberField`
- compiled eliminant value checks
- compiled `sqrt 2` operation checks for normalized polynomials and selected embeddings

Stacked on #8906.

PR #8908, opened 2026-07-27
feat(number-field): bound candidate disambiguation

kim-em:
## Summary

- normalize candidate-evaluation eliminants for reciprocal-Cauchy bounds
- compute fixed-field coordinate and Horner error majorants
- implement the finite least-precision search from the number-field SPEC
- retain candidates only when a sufficiently small certified ball cannot exclude zero

## Verification

- `lake build HexNumberField.Disambiguate HexNumberField`
- compiled normalization, endpoint, least-precision, and zero-eliminant checks

Stacked on #8907.


kim-em:

PR #8911, opened 2026-07-27
interval: start D2 replay experiments

kim-em:
Follow-up to #8895.

## What this adds

- activates the Mathlib-free `HexInterval` library and colocates its SPEC;
- implements exact raw empty/open/closed/unbounded cuts, consistency proofs,
  and idempotent normalization;
- adds `normalizeWithin`, which preflights every finite endpoint's dyadic
  height and any alignment shift and returns a distinct resource-limit result
  before comparison;
- compares proof-bundled and externally checked hot representations without
  selecting the public representation;
- demonstrates two Mathlib-free rational replay paths: exposed Core
  `Rat.normalize` wrappers and sound numerator/denominator cross-product
  certificates for add/sub/mul/inverse, including fold-level composition;
- adds CI-covered downstream `decide +kernel` and forced
  `Lean.Kernel.whnf` canaries, conformance cases, a compiled spike, and a
  reproducible source-hashed measurement report.

The report deliberately narrows the result: this measures compiled newtype
erasure and a validation scan, not serialized proof-bearing literals versus a
reflected trace. The rational arms isolate different costs (normalizing values
versus checking planner-supplied literals), so end-to-end production on the
same DAG remains open for the next centered-product vertical.

No path uses `native_decide`.

## Evidence

- bundled vs checked compiled construction is within about 2%; the checked
  handoff scan costs about 12–13% in this workload;
- both generated 433-value kernel paths are feasible (729.3 ms bundled versus
  804.7 ms checked median forced WHNF), but this projected-value probe does not
  measure proof-bearing literal layout;
- for 433 rational additions, exposed normalization takes 92.7 ms forced WHNF
  versus 157.0 ms for the per-edge cross-product verifier; the cross-product
  theorem module is smaller and has no axioms, so the representation decision
  remains open.

## Verification

- `lake build`
- `lake build HexIntervalExperiment HexIntervalReplayProbe HexConformance hex_interval_representation_spike`
- `python3 scripts/bench/interval_representation_sweep.py --samples 5 --sizes 433 4096`
- `python3 scripts/conformance_targets.py --check`
- `git diff --check`


kim-em:
Independent Claude Opus review completed before merge. I applied the review's
substantive boundary and measurement corrections:

- made the valid-handoff scan and malformed-trace rejection explicit;
- reframed the rational arms as different cost isolates and added a fold-level
  theorem composing accepted cross-product edges to Core `Rat` arithmetic;
- removed experiment modules from the public umbrella and gave them an
  explicit build-only target;
- renamed the unsafe canonicalizer to `normalizeUnchecked` and documented the
  safe future entry point;
- retained signed margins, added import-matched WHNF baselines, capability-
  tested the RSS timer, and rejected a zero-repeat driver invocation;
- regenerated the source-hashed record and reconciled the report.

One suggestion was deliberately not adopted: treating absolute dyadic exponent
magnitude as accidental over-rejection. The accepted SPEC intentionally makes
endpoint height a dynamic-range budget; `maxAlignmentShift` remains the
separate allocation guard. The review's requested all-finite high-precision
workload and identical end-to-end rational DAG remain explicit follow-up
experiments rather than conclusions smuggled into this PR.

PR #8914, opened 2026-07-27
feat(number-field): add common-field algebraic roots

kim-em:
## Summary

- construct a bounded deterministic primitive presentation for algebraic coefficient arrays
- recover and validate fixed-field coordinates through the trace pairing
- add `AlgebraicPoly.roots?` / `roots` and compiled cross-field root checks

## Verification

- `lake build`
- `lake build HexNumberField.Roots`

Stacked on #8912.

PR #8916, opened 2026-07-27
feat(number-field-tower): activate rational carrier scaffold

kim-em:
## Summary

- activate `HexNumberFieldTower` in Lake, the library graph, and the existing single CI job
- move the immutable tower SPEC into its per-library source layout
- add the sealed tower representation, rational tower, fixed-width mixed-radix carrier, and exact equality

## Verification

- `lake build HexNumberFieldTower`

Stacked on #8914.


kim-em:

PR #8918, opened 2026-07-27
feat(number-field-tower): add mixed-radix arithmetic

kim-em:
## Summary

- implement recursive mixed-radix convolution and monic reduction
- implement recursive polynomial-xgcd inversion with `0⁻¹ = 0`
- add the tower arithmetic instances and compiled one-/two-level checks

## Verification

- `lake build HexNumberFieldTower`
- `lake build`

Stacked on #8916.

PR #8919, opened 2026-07-27
Number field tower: checked rational extensions

kim-em:
## Summary
- add the dependent NumberTower.Extension result
- add the sealed ofQAdjoin smart constructor with positive-leading sign normalization
- exercise positive- and negative-leading presentations of Q(sqrt 2)

## Verification
- lake build HexNumberFieldTower.Embed
- lake build

Stacked on #8918.


kim-em:

PR #8920, opened 2026-07-27
Number field tower: one-level Trager norms

kim-em:
## Summary
- add runtime-indexed lower-tower coefficients for recursive polynomial algorithms
- compute one-level Brown resultants for Trager shifts
- add squarefreeness checks, the exact collision bound, and deterministic signed search
- cover one- and two-level norm calculations with compiled regressions

## Verification
- lake build HexNumberFieldTower.Norm
- lake build

Stacked on #8919.


kim-em:

PR #8921, opened 2026-07-27
Number field tower: Yun decomposition

kim-em:
## Summary
- add Yun squarefree decomposition on runtime and public tower polynomials
- self-check positive multiplicities, monic squarefree components, and reconstruction
- add the rational Berlekamp-Zassenhaus recursion base with exact reconstruction
- cover repeated factors over Q and Q(sqrt 2)

## Verification
- lake build HexNumberFieldTower.Factor
- lake build

Stacked on #8920.


kim-em:

PR #8922, opened 2026-07-27
Number field tower: Trager factorization

kim-em:
## Summary
- implement bounded recursive Trager factorization over tower tails
- recover shifted factors by gcd, undo shifts, and verify reconstruction
- add canonical sorting and recursive executable irreducibility replay
- expose checked Factorization and factor? dependent APIs
- cover collisions, irreducible components, multiplicities, zero, and constants

## Verification
- lake build HexNumberFieldTower.Factor
- lake build

Stacked on #8921.

PR #8925, opened 2026-07-27
feat(tower): select factors by fixed embedding

kim-em:
## Summary

- evaluate mixed-radix tower elements at their stored absolute generators
- bound and refine tower-polynomial evaluation through the finite disambiguation endpoint
- lift integer polynomials and retain the unique factor vanishing at a requested algebraic root
- cover the conjugate-impostor case over Q(sqrt(2))

## Verification

- lake build HexNumberFieldTower.Split
- lake build HexNumberFieldTower

Stacked on #8922.


kim-em:

PR #8929, opened 2026-07-27
interval: validate centered instantiation replay

kim-em:
## Summary

- add a Mathlib-free nine-node centered-product instantiation and interval-trace checker with caller-bound sources, target, base snapshot, and resource limits
- prove generic real semantic soundness downstream and construct the concrete fixed-program valuation, yielding the unconditional bound `0 ≤ x * (1 - x) ∧ x * (1 - x) ≤ 1 / 4`
- add discriminating adversarial conformance cases, ordinary-kernel replay/WHNF canaries, compiled rejection probes, and matched reflected/direct semantic proof probes
- specify a Mathlib-free rational path using compiled Core `Rat` planning plus transparent canonical raw-table replay; Mathlib `norm_num` is an optional surface-leaf convenience only
- record a clean post-merge five-sample measurement without freezing production storage, scheduling, endpoint, or proof-construction choices

## Evidence

- fixed certificate: 9 nodes, 10 facts, exact indexed-lookup budget 74
- compiled accepted check: median 4.836 us
- forced ordinary-kernel WHNF: median 45.1 ms
- reflected/direct downstream proofs both feasible; their small timing difference does not select a design
- reflected/direct public `.olean`: 10,952 bytes each; private artifacts: 26,672 / 14,168 bytes
- checked/reflected/direct axiom reports contain only expected Lean axioms; no `sorryAx` or execution-trust axiom
- no `native_decide`, `axiom`, `sorry`, or `nlinarith` in the vertical

## Validation

- `lake build` against merged current main: 9,468 jobs
- focused Core, Mathlib, conformance, and compiled-spike builds
- Python compilation of both measurement and bench-policy scripts
- schema-v3 JSON provenance validates the clean merge commit and all 16 source hashes

## Deliberately open

The transparent list and one-generation checker are proof-facing canaries, not production storage or generation selections. Growing traces, proposal deduplication/relevance, multiple generations, open/unbounded arithmetic replay, exact-rational backend tradeoffs, and end-to-end certificate production remain follow-up experiments.


kim-em:
Fresh Claude Opus second opinion completed against the full branch. It found no P0/P1 issue and no soundness hole in the checker or semantic replay.

Dispositions:

- Fixed axiom reporting so missing `#print axioms` output is `null`, while modules expected to report now fail closed if output is absent.
- Added `.olean.private` and `.olean.server` measurements and rewrote the storage conclusions around the actual private proof artifacts.
- Removed the unsupported attribution that trust-boundary hardening itself was cheap; the report now states only the measured end-to-end cost.
- Added a self-referential appended node that only whole-program topology validation rejects, plus unused malformed-source cases that isolate whole-source-table preflight.
- Constructed and proved the concrete fixed-program valuation, so the fixture now yields an unconditional theorem about `x * (1 - x)` rather than stopping at an assumed `Program.Holds`.
- Corrected `maxLookupSteps` claims: indexed list lookups are charged; sequential validation scans are bounded separately by structural caps.
- Made the one-step-short compiled workload exactly 73, clarified the one-generation canary, removed duplicate Core certificate canaries, and documented/linted the narrow build-only Mathlib proof-probe exception.

One scope-only suggestion was not followed: the reviewer considered the rational-planning discussion optional for this vertical. The user explicitly asked for the Mathlib-free rational story, so the SPEC keeps it, now sharpened to compiled canonical Core `Rat` planning plus transparent canonical raw-table replay, with `norm_num` confined to an optional Mathlib surface adapter.

The review also noted normalization/deduplication opportunities that are deliberately left as measured follow-up design questions rather than silently frozen here.

PR #8931, opened 2026-07-27
feat(tower): validate adjoining extensions

kim-em:
## Summary

- seal recursive tower validity behind structural, irreducibility, and fixed-embedding checks
- split cycle-free raw arithmetic, factorization, and evaluation internals from the dependent public wrapper
- implement root adjoining with identity and nonlinear checked-extension cases
- add positive two-level and adversarial wrong-root regressions

## Validation

- lake build HexNumberFieldTower.Split
- lake build

## Stack

Base: #8925

PR #8932, opened 2026-07-27
feat(tower): construct splitting fields

kim-em:
## Summary

- iterate relative norms to obtain absolute eliminants
- isolate and select absolute roots of nonlinear tower factors
- adjoin roots through the validated extension boundary and refactor with bounded fuel
- represent zero-polynomial, constant, and finite-root splitting results
- cover quadratic, repeated-factor, and multi-step quartic splitting cases with compiled regressions

## Validation

- lake build HexNumberFieldTower.Split
- lake build

Stacked on #8931.


kim-em:

PR #8933, opened 2026-07-27
feat(tower): flatten primitive presentations

kim-em:
## Summary

- combine fixed tower generators with a bounded deterministic signed-shift search
- require the accepted canonical primitive element to have the full accumulated dimension
- recover old generators in the primitive power basis through exact trace-pairing row reduction
- construct both coordinate maps and validate their composites on both rational bases
- cover rational, one-level, and two-level tower flattening with compiled regressions

## Validation

- lake build HexNumberFieldTower.Flatten
- lake build

Stacked on #8932.


kim-em:

PR #8937, opened 2026-07-27
interval: measure structural checker scaling

kim-em:
## Summary

- add four transparent, Mathlib-free centered-checker scaling families with exact closed-form and independently recomputed five-component lookup costs
- measure compiled construction, exact acceptance, one-step-short rejection, malformed-tail rejection, structural-cap rejection, whole-module `decide +kernel`, and forced kernel WHNF across 21 ladder points
- record a clean five-sample, 105-case evidence artifact and carry its storage/resource conclusions into the interval SPEC
- specify and begin the immediate endpoint-erased, compiled Core-`Rat`, canonical raw-table replay vertical without relying on `norm_num` or `native_decide`

## Key evidence

At 500 facts, adjacent and far-reference certificates have identical dimensions but require 1,544 versus 121,839 charged list-constructor lookups and take median 49.3 µs versus 163.1 µs in compiled replay. This keeps the original-order list as a transparent reference checker and requires separate node/fact/lookup budgets plus production array/arena/chunked experiments.

The review-hardened harness computes list-derived limits outside timed loops and asserts both exact and harmless-spare parity variants. It also exposes a remaining accounting gap: centered equality-recipe program lookups are bounded by `maxEdges * maxNodes` but not yet charged to `maxLookupSteps`; the report and SPEC require an independent edge-growth experiment and either explicit charging or validated random access before production.

The largest committed workload in every family replays through ordinary `decide +kernel`. Baseline and checked theorem axiom reports are both exactly `[propext, Quot.sound]`; command-only WHNF probes report `null` rather than asserting an axiom set. No implementation, test, benchmark, or fallback uses `native_decide`.

The report is `reports/hex-interval-scale.md`; raw samples, exact counters, artifact hashes, axiom reports, and source provenance are in `reports/bench-results/hex-interval-scale.json`.

## Validation

- `lake build` — 9,470 jobs after merging current `origin/main`
- `lake build HexIntervalExperiment hex_interval_scale_spike +HexInterval.ScaleConformance:olean`
- `python3 -m py_compile scripts/bench/interval_scale_sweep.py`
- `git diff --check`
- clean scientific run: `python3 scripts/bench/interval_scale_sweep.py --samples 5 --target-ms 250 --max-calibration-rounds 12`

A fresh independent Opus review found no P0/P1 correctness or evidence errors. Its two P2 concerns and all actionable P3 observations have been dispositioned in the revised source/report/SPEC; details are in the PR comment.


kim-em:
Independent Opus review dispositions:

- **P2 — timed limit construction:** fixed. Every list-derived `StructureLimit` is precomputed outside the timed loop; only constant-time parity record updates remain. Both parity outcomes are asserted before timing, and the complete evidence bundle was regenerated from the fixed source.
- **P2 — equality-edge program lookups not charged:** made explicit in source docs, the report, and the SPEC. The reference checker currently bounds them by `maxEdges * maxNodes`; production must charge them or use validated random access, and an independent edge-padding workload is now required.
- **P3 — exact/spare wording and parity assertion:** corrected and tested.
- **P3 — closed-term sharing in builders:** report now says construction timing is not a retained-size proxy.
- **P3 — kernel far/adjacent attribution:** conclusion softened; a normalized-certificate arm is required before attributing the inversion to lookup reduction.
- **P3 — provenance and dead calibration branch:** conformance source is now hashed and the unreachable branch was removed.

Replacement record checks: 105/105 calibrations within tolerance, 525 compiled samples, 21 forced-kernel points, stable source hashes/executable, and theorem axiom sets exactly `[propext, Quot.sound]`.

kim-em:
Follow-up accounting audit reproduced every JSON/report value and prior review disposition. It identified one documentation-only precision issue in the earlier comment: `centerV1` performs nine program lookups for the selected center and nine per equality edge. Commit `8d1ee51e` therefore records the literal reference-checker bound as `9 * (maxEdges + 1) * maxNodes`, not `maxEdges * maxNodes`. Measurements and charged counters are unchanged because this work is explicitly outside `maxLookupSteps`; production still must charge it or use validated random access.

PR #8944, opened 2026-07-27
interval: validate rational certificate boundary

kim-em:
## Summary

- add a backend-independent endpoint-erased skeleton for the centered/scale certificate family, preserving interval openness, unboundedness, derivations, equality recipes, caller sources/target, result, structural limits, and literal-sharing shape
- validate complete canonical raw rational tables with caller-owned entry and numerator/denominator-bit limits, rejecting zero denominators, nonreduced encodings, noncanonical zero, and invalid unused tail entries
- add a table-indexed rational certificate boundary with typed finite-cut/literal references, full SSA/source/fact/edge/result validation, and a separate exact caller-owned endpoint-table traversal budget
- pin every caller-owned indexed source and target to exact raw rational values, with an elimination theorem proving that a successful check cannot reinterpret the invocation through certificate-chosen table entries
- specify the next Mathlib-free Core-`Rat` arithmetic-edge identities, temporary-bit/work preflight, deterministic validation phases, projection requirements, and upgradeable cancellation arms

## Boundary guarantees

The transparent checker validates the complete table before accepting any certificate. A ready entry is proved canonical and its Core `Rat` interpretation has exactly the checked numerator and denominator. The table-indexed fixture consumes exactly 73 endpoint-list lookup steps; 72 rejects before any indexed endpoint lookup. Empty, open/closed, one-sided unbounded, and two-sided unbounded shapes survive erasure, while strictness and literal-sharing changes remain observable.

The rational and dyadic centered fixtures erase to the same independently written skeleton. Internal arithmetic recipe and interval propagation identities remain deliberately deferred, but changing the certificate table cannot change the caller's source or target meaning: exact caller values are checked and retained in the rational projection. A negative canary changes every canonical table value and demonstrates rejection against the original caller boundary.

## Validation

- initial whole-repository `lake build` for the canonical-table slice: 9,103 jobs
- focused: `lake build HexIntervalExperiment +HexInterval.RationalTableConformance:olean +HexInterval.RationalCertificateConformance:olean`
- forbidden implementation-token scan and `git diff --check`

This branch is Mathlib-free and uses no `norm_num`, Grind, `native_decide`, `axiom`, or `sorry`. The first independent review's trust-boundary, validation-order, literal-sharing, and negative-coverage findings are fixed; a fresh follow-up Opus audit is running asynchronously on this head.

### DIFF
--- a/HexInterval/Experiment/RationalCertificate.lean
+++ b/HexInterval/Experiment/RationalCertificate.lean
@@ -18,6 +18,8 @@ Finite program literals and interval cuts refer to a complete, canonical
 `RawRat` table by typed indices.  The transparent boundary checker validates
 the whole table, every endpoint index, SSA and derivation indices, equality
 references, caller-owned sources and target, and structural budgets.
+Caller rows contain both indexed syntax and exact raw values, so the
+certificate-owned table cannot reinterpret the invocation boundary.
 
 No interval arithmetic or equality arithmetic is replayed here.  The module
 also does not choose a wire format or a production storage representation;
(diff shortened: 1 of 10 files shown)


kim-em:
Independent review disposition for the original rational-boundary head:

- **Caller-owned source/target meaning (P1): agreed and fixed.** `ExpectedRow` now pairs each caller-owned indexed row with an exact `ValueRange` of raw rationals. `Certificate.check` resolves every finite cut only after whole-table validation, `boundary_agrees_of_check` exposes the accepted invariant, the rational projection retains those values, and a changed canonical table is rejected against the original boundary while still being accepted against independently changed caller values.
- **Table validation precedence (P2): agreed.** The implementation intentionally keeps deterministic per-entry interleaving. The SPEC now states this precisely, and a mixed `[nonreduced, later-oversized]` canary pins the earlier logical failure.
- **Declared versus implemented resource taxonomy (P2): agreed.** The SPEC now identifies the exact table/certificate phases implemented here and marks encoded-byte, independent gcd-work, arithmetic, and projection limits as composed-target work rather than present table fields.
- **Literal erasure strength (P2): agreed and fixed.** Both dyadic and rational erasure now assign first-occurrence literal-sharing ordinals. Reusing one literal in all three centered positions changes the skeleton and has negative coverage.
- **Rational structural negatives (P2): agreed and added.** New canaries reject a wrong literal constructor, wrong equality endpoint, generation 0/2, and the exact 74/73 structural lookup boundary.
- **Endpoint read accounting / program lookup accounting / edge namespace / cancellation identities (P3): fixed or clarified.** The deliberate post-preflight indexed reads are documented; endpoint-table and bounded program lookups are distinguished; the local edge-shape namespace no longer shadows `Center.EqEdge`; and the reduced cancellation identities are written explicitly.

I retained both cross-backend and independently-written skeleton comparisons in `checksProjection`: they are logically related by another canary, but keeping both makes this boundary directly test both claims without relying on that theorem. Focused builds and static checks are green. A fresh follow-up review is running on commit `4c74d9c2`.

kim-em:
Closing: overtaken. This adds the next proof-facing layer (`RationalTable`, `RationalCertificate`) on the D2 rational vertical that #8911 started, but there have been ~310 commits to `HexInterval/` since it forked and the vertical has since moved on to sealed typed runtimes and the typed runtime controller (#9377, #9427, #9429, #9527). Rebasing a July layer onto that is a rewrite rather than a merge. Reopen or re-land if the table-indexed certificate boundary is still wanted.

PR #8959, opened 2026-07-27
prove(resultant): close primitive power and degree bounds

kim-em:
## Summary

- prove the binary `powNat` implementation agrees with lightweight-semiring exponentiation, without assuming commutative multiplication
- prove the pseudo-division remainder has size strictly below the nonzero divisor by tracking the output fold size
- require checked irreducibility for the NumberField semantic zero and degree contracts, excluding reducible-presentation counterexamples

## Why

These are the first proof-bearing obligations in the Resultant correctness kernel. The power theorem supports Brown scalar updates, while the remainder bound supplies the local degree descent needed before decomposing the Brown-chain validity proof. The NumberField assumptions repair two theorem signatures that were false for reducible presentations such as `X² - 1` at the selected root `1`.

## Validation

- `lake build HexResultant HexNumberFieldMathlib HexConformance HexManual` (9,527 jobs)
- `lake exe hexresultant_bench verify` (all four targets)
- `git diff --check`

An asynchronous Claude second-opinion monitor has been launched, but its OAuth refresh is currently unavailable; implementation work is continuing independently.

### DIFF
--- a/HexNumberFieldMathlib/Roots.lean
+++ b/HexNumberFieldMathlib/Roots.lean
@@ -108,13 +108,15 @@ theorem coeff_toPolynomialAt (f : DensePoly (QAdjoin p x))
 
 /-- Fixed-field executable zero detection agrees with semantic polynomial
 zero. -/
-theorem poly_isZero_iff (f : DensePoly (QAdjoin p x))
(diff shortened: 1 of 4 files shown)


kim-em:

PR #8960, opened 2026-07-27
prove(tower-mathlib): close basic semantic projections

kim-em:
## Summary

- prove that a successful recursive raw evaluation computes `NumberTower.toComplex`
- prove every validated tower has positive absolute dimension
- isolate the reusable list-level dimension induction as a short private helper

## Why

These are the smallest Tower Mathlib proof obligations independent of the unfinished NumberField and Resultant semantic bridges. They validate the fallback-free projection boundary and the structural dimension invariant without adding assumptions.

## Validation

- `lake build HexNumberFieldTowerMathlib HexManual`
- `git diff --check`

This PR is stacked on #8959 so implementation can continue while its independent review and CI run asynchronously.


kim-em:

PR #8978, opened 2026-07-27
Expose bounded expression structure to interval propagators

kim-em:
## Summary

Lets arbitrary propagators inspect the immutable typed expression graph that their engine-issued action belongs to.

- adds a versioned bounded `ProgramView` with operation signatures, nodes, and instantiation generations
- includes that view in FIFO and checked-policy `RuleRequest`s while continuing to expose facts only through declared watch slots
- provides exact structural lookup helpers without teaching the engine what any operation key means
- keeps reply validation and all mutation authority in the engine
- revalidates retained proposals against the complete current append-only program instead of invalidating every peer proposal after one extension

## Evidence

The conformance registry interprets otherwise opaque keys and recognizes exactly `x * (one - x)`. It rejects a reversed subtraction and a product repeating the wrong input, proposes a centered alternate and equality, then repeats the match on a newly appended `y * (one - y)` suffix. Both the still-fresh old proposal and the new proposal admit after current-state type, reference, CSE, equality, and generation checks, producing exact generations one and two.

## Open design

The SPEC keeps two production representations open: the whole bounded immutable view implemented here, and compiler-generated per-rule structural bindings. The experiment establishes the semantic access and revalidation contract without prematurely selecting the physical matcher interface.

## Validation

- `lake build HexInterval.Experiment.Propagator HexInterval.Experiment.Policy HexInterval.StructureViewConformance`
- Propagator, policy, and frontier conformance builds
- copyright, line-count, trust-surface, and diff checks
- no `native_decide`, axioms, or sorries


kim-em:
Design boundary before merge: this PR exposes the immutable structural view, but the view alone is not a wakeup dependency. A matcher which depends on structure outside its anchor could otherwise become stale after another rule extends the program. The follow-on package PR #8984 adds an explicit `watchesProgram` registration flag, stales retained actions at the old program version, and requeues those existing matchers after every append-only extension. Anchor-local rules remain append-stable. The production alternative of finer operation-key structural triggers remains open for measurement.

PR #9032, opened 2026-07-28
refactor(rcf): keep regression modules out of public API

kim-em:
## What changed

- remove all twelve `HexRCF.*Tests` imports from the public `HexRCF` umbrella
- build those modules through the dedicated `HexReleaseTests` target instead
- document the public/test boundary in the HexRCF SPEC

## Why

The Phase-3 module contract requires public umbrellas to contain only supported API while regression modules remain build-checked separately. HexRCF was the only root umbrella still re-exporting its complete internal test suite.

Consumers of `import HexRCF` now receive only implementation and proof API declarations. Regression coverage is unchanged.

## Validation

- `lake build HexRCF HexReleaseTests`
- `lake build HexManual HexConformance HexRCFProofProbe`
- copyright and file-line-count checks
- `git diff --check`


kim-em:
Fresh Claude Opus review verdict: GO. It independently verified that all twelve regression modules are preserved, no public declaration is lost, the Lake/DAG registries parse correctly, the released-manifest set remains unchanged, and CI stays in the existing single job. I accepted its useful default-build suggestion by adding `@[default_target]` to `HexRCFTests`; `lake build HexRCFTests` and the repository validation scripts pass. The manual was also built locally. Its remaining notes were low-priority follow-ups about enforcing the boundary generically and future release migration.

PR #9073, opened 2026-07-29
docs: complete the hex-mv-poly SPEC migration

kim-em:
## Summary

- remove the multivariate-polynomial entry from `SPEC/future-work.md` now that it has a dedicated library SPEC
- expand `SPEC/Libraries/hex-mv-poly.md` into an implementation-facing contract for representation, comparator laws, APIs, correctness, kernel probes, conformance, benchmarks, and staging
- register `hex-mv-poly`, its Mathlib companion, and `hex-basic` accurately in the library catalogue and dependency summary
- preserve content and primitive part under the still-unspecified multivariate-gcd work

## Independent review

A fresh Claude Opus review found no blocker to the migration. Its actionable API and integration findings were verified and incorporated: Mathlib-only equivalences now live in the companion, kernel elaboration measurements use a Mathlib proof-probe root, comparator laws elaborate as inherited instances, monomial helpers and collision-producing coefficient contracts are explicit, and the SymPy oracle requirements include installation and required-dependency preflight.

## Validation

- `git diff --check`
- local relative-link target check for all three documentation files
- confirmed no multivariate-polynomial section remains in `SPEC/future-work.md`

Documentation-only; no Lean build was required.

### DIFF
--- a/SPEC/Libraries/README.md
+++ b/SPEC/Libraries/README.md
@@ -1,7 +1,9 @@
 # Libraries
 
+- **hex-basic**: small Mathlib-free standard-library shims, including kernel-reducible array and vector operations
 - **hex-arith**: extended GCD, Barrett/Montgomery reduction, binomial coefficients, Fermat's little theorem
 - **hex-poly**: dense `Array`-backed polynomial representation
+- **hex-mv-poly**: canonical distributed multivariate polynomials at fixed arity with explicit monomial orders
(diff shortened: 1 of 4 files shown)


kim-em:
Independent Claude review completed and verified against Lean 4.32.0-rc1. The follow-up commit addresses all actionable findings: companion DecidableEq bounds, explicit target comparators, the core derivative formulation, early kernel-probe sequencing, complete Phase 4 metadata/statuses, monomial-order proof obligations, Mathlib-free class choices, and stale oracle/editorial text. Local documentation/DAG/Phase-4 checks pass.

PR #9082, opened 2026-07-29
feat(interval): instantiate arbitrary propagator scopes

kim-em:
## Outcome

This makes arbitrary-scope propagators first-class outputs of the instantiation mechanism. One atomic event may add expressions, equalities, and propagator applications whose ordered reads and writes refer to old or same-event expressions.

This is function-agnostic framework work. Its conformance fact domain is a tiny Nat signal used only to expose scheduling and rollback; it does not depend on rational endpoint arithmetic.

Stacked on #9081.

## Design

- Instantiation proposals can contain scoped applications.
- Nodes, equalities, and scopes resolve against one prospective typed program before commit.
- Engine applications form the canonical append-only arena; no old ApplicationId moves.
- Scope-only admission is a real network extension.
- Package-owned scope predicates are checked at start, retention, policy refresh, dispatch, and final admission.
- Application creation generation prevents scope-only chains from resetting theorem depth.
- Proposed outputs remain order-independent when CSE reuses old storage, including equality endpoints and scope ports; explicitly existing inputs still contribute their generation.
- Events record resolved bindings, output IDs, fresh subsets, and generation for later replay.
- Resource checks cover application count, queue growth, per-scope ports, and policy semantic surface before commit.

## Validation

lake build HexIntervalExperiment HexInterval.PropagatorConformance HexInterval.StructureViewConformance HexInterval.PolicyConformance HexInterval.PolicyFrontierConformance HexInterval.PolicyDriverConformance HexInterval.PackageRegistryConformance HexInterval.DyadicRulesConformance HexInterval.ScopeConformance

All 22 local jobs pass. The changed implementation and conformance code contains no native_decide, axiom, sorry, or unsafe declaration.

## Open follow-ups

- Consume binding/application provenance in semantic replay.
- Integrate the engine-owned bounded structural matcher cursor.
- Give package scope-validation work its own deterministic accounting.
- Compare incremental watcher indexes with the current full rebuild.
- Keep propagation, strengthening, instantiation, and subdivision policy replaceable.


kim-em:
Independent review follow-up is now in 6ed04bc2:

- CSE-resolved old equality endpoints participate in causal generation in both policy preview and final admission, with an exact generation-two canary.
- Registry dispatch repeats the package-owned scope veto before entering the callback, even if a raw engine was started with a permissive hook.
- Policy semantic-surface accounting charges both ordered reads and writes; direct per-application and aggregate exact-boundary canaries cover the change.
- The prior node-CSE canary now keeps its equality endpoints explicitly old, separating harmless storage reuse from causal use as an equality endpoint.

The full 22-job validation command in the PR body passes.

Issue #9130, opened 2026-08-02
BZ recombination: reject impossible exact divisors over a word-sized prime


kim-em:
Tracked by #9126. Blocked by #9129.

## Problem

On the `x^n - 1` rows, a large share of direct reconstruction time is spent proving that candidate products do not divide the integer polynomial:

- `xpow105_minus1`: about 31% under `exactQuotient?`;
- `xpow120_minus1`: about 63%, mostly dense integer long division.

Most candidates are non-divisors. Running full GMP-backed polynomial division for every one is unnecessarily expensive when a finite-field remainder can disprove divisibility.

## Proposed work

After metadata-only support filters pass, but before constructing or dividing by a large integer candidate:

1. reduce the target and lifted factors modulo a fixed word-sized prime `q`;
2. form the selected modular product from cached reductions and the support cursor;
3. compute the target remainder modulo that candidate in `F_q[X]`;
4. reject the support if the remainder is nonzero;
5. otherwise fall through to the existing exact integer candidate construction and `exactQuotient?`.

If reduction makes the candidate zero, drops its leading coefficient, or otherwise prevents sound finite-field division, return “inconclusive” and use the exact path. A second modulus may be considered only after measuring the rejection/cost tradeoff for one.

This is a necessary-condition filter, not a probabilistic answer. Exact integer division remains the only accepting test.

## Proof obligations

State the filter as a mathematical divisibility obstruction and prove:

- reduction modulo `q` preserves multiplication;
- if `candidate ∣ target` in `Z[X]`, then the reduced candidate divides the reduced target;
- whenever the reduced candidate admits field polynomial division, integer divisibility implies zero modular remainder;
- therefore a nonzero modular remainder can never reject a genuine factor;
- every inconclusive or zero-remainder result falls through unchanged to exact division.

Give the optimized word-polynomial computation an executable specification and a proved equality theorem. Do not introduce an unchecked native oracle or accept a factor on modular evidence alone.

## Organization

The natural objects are a modular image of a reconstruction support and a divisibility obstruction. Avoid family-specific recognizers such as special handling for `x^n - 1`, and avoid software labels such as `FastRejectEngine`.

Integrate with #9129's `SupportCursor` so the modular product is assembled directly from cached reduced factors. Do not recreate the old selected-factor list merely to feed this filter.

## Measurements

[...]

PR #9138, opened 2026-08-02
perf: filter recombination supports before materializing them

### LINKED ISSUE #9129: BZ recombination: enumerate factor supports without allocating lists at every branch
Tracked by #9126. Blocked by #9127.

## Problem

The direct classical reconstruction spends most of its time traversing candidate supports on the Swinnerton-Dyer rows:

- `sd5`: about 86% under direct combination scanning;
- `sd5_x_phi11`: about 87%;
- allocation accounts for roughly 37--64% of sampled leaves on these representatives.

At each leaf, `CombinationIterator.lean` currently reverses selected indices, reverses/concatenates rejected indices, maps selected indices to a newly allocated list of lifted polynomials, and only then enters `tryDirectCandidate`. Because Lean is strict, the factor list is materialized before the inexpensive degree and trailing-coefficient tests can reject the support.

Isabelle examines a comparable number of supports on SD5. The 5--8x paired gap is therefore largely traversal and representation overhead rather than a stronger mathematical obstruction.

## Proposed work

Define the mathematical object first:

- a factor `Support`, represented extensionally as a finite subset or increasing sequence of indices;
- the size-`k` supports required by complement symmetry and the head-forced search;
- the coverage and uniqueness statements needed by classical reconstruction.

Then provide one executable `SupportCursor` using a compact array of increasing indices. Precompute immutable metadata for each lifted factor, including degree and trailing residue. During traversal:

1. update degree and trailing data incrementally;
2. run every metadata-only rejection before constructing polynomial factors;
3. materialize the selected polynomial product only for surviving supports;
4. construct the complementary support only after a divisor succeeds;
5. reuse buffers or persistent arrays where that is clearer than allocating lists at every recursion node.

Remove the superseded list-building traversal after its proofs and consumers have moved. Do not retain a legacy iterator alongside the new one.

## Proof obligations

Prove, without `native_decide`, that the executable cursor:

- maintains strictly increasing in-range indices;
- enumerates every required size-`k` support exactly once;
- introduces no duplicate supports;
- implements the same head-forced complement coverage as the present classical completeness proof;
- preserves the first-success result under the same candidate predicate.

Keep cursor correctness separate from the theorem that a successful classical reconstruction yields irreducible integer factors. The optimized representation should not leak through the public factorization theorem.

## Before/after measurements

Using #9127's protocol, record for SD5 and controls:

- traversal nodes by support size;
- metadata rejections;
- selected products materialized;
- complementary supports materialized;
- exact divisions attempted;
- allocation count/bytes and peak residency;
- traversal, candidate, and total factor time.

Include `sd5`, both shifts, `sd4_x_sd4shift1`, `sd5_x_phi11`, the three `xpow` rows, Wilkinson 48/56, easy low-width cases, and a row that reaches the subset budget.

First prototype the cursor in a benchmark-only module. Proceed with production replacement only if the profile shows that allocation/list construction was actually removed rather than displaced into another representation.

## Acceptance criteria

- Candidate products are not constructed before metadata-only filters finish.
- Allocation attributable to support traversal falls by at least 40% on `sd5` and `sd5_x_phi11`, or the issue records a measured no-go explaining the remaining allocation source.
- The SD family shows a material end-to-end improvement, with no unexplained regression above 5% on low-width controls.
- Search order, budget behavior, factor results, and proof coverage remain deterministic.
- Full build, conformance, oracle, and factorization tests pass.
- The PR records a fresh full Hex sweep and regenerates all factorization cactus and runtime-by-degree plots, reusing unchanged external comparator records.

## Dependencies

Blocked by #9127.

Blocks #9130 and, through it, #9133. If the cursor lands, remove the `blocked` label from #9130; if it is a measured no-go, revise #9130 against the demonstrated representation before unblocking it. It can proceed in parallel with the prime-plan and Hensel issues once the common baseline is recorded.

kim-em:
This PR runs the head-forced recombination leaf's metadata-only filters before it materializes anything, so a support the degree and trailing-coefficient tests reject never reverses the selected indices, never maps them to lifted polynomials, and never concatenates the complementary support. The lift modulus and each lifted factor's degree and trailing coefficient are precomputed once per subset-cardinality level into `Hex.SupportMeta`, whose proof fields pin every recorded value to the lifted factor it describes, so `scanDirectCombinations_found` and `scanDirectCombinations_finds` now hold for any metadata record rather than only the canonical one. The old leaf built the reversed selection, its factor list, and `p ^ k` as arguments to `tryDirectCandidate`, whose own first step is the prefilter; on `sd5` that path is taken by 32,639 of 32,768 leaves, and `p ^ k` was rebuilt 311,280 times.

Swinnerton-Dyer end-to-end factor time is 0.707x to 0.854x across the family (`sd5` 0.739x, `sd5_shift1` 0.707x, `sd5_x_phi11` 0.734x, `sd5_x_phi45` 0.782x, `sd4_x_sd4shift1` 0.854x), traversal outside the candidate test falls 40.0% on `sd5` and 47.6% on `sd5_x_phi11`, and whole-factorization allocator time falls 28.9% and 28.6%. Every control is within 3% and every profiled non-SD row within 1.3%. The SD family's median against verified Isabelle BZ improves from 2.799x to 2.154x. All recombination stage counters, completed cardinalities and returned factor degrees are identical before and after, and the counted mirror agrees with the production trace on 367 of 367 validation rows.

The measurement record is `reports/hexbz-support-traversal.md`, with a fresh full Hex sweep, phase profile and sampling profiles at this revision and all 25 factorization figures regenerated; the external comparator record is reused unchanged. Those records are pinned to CPU 70 rather than CPU 0, because three other agents were running the same CPU-0-pinned protocol on this host and sweeps taken there had a median of 1.10x with outliers to 6.8x on rows that cannot have changed; the substitution is validated against the baseline on the 354 corpus rows this change cannot affect, with median 0.9977 and p90 1.0234.

The issue also proposed a `SupportCursor` over a compact array of increasing indices, replacing the list traversal outright. That part is deliberately not done, and the report says why with evidence: the cost was at the leaf, not in the spine, and with the leaf guarded the residual traversal is two constructor allocations and one modular multiply per internal node while the remaining leaf cost is 93-bit modular arithmetic (`Hex.centeredModNat` alone is 8.8% of `sd5`'s total). The representation is not what is left to win; the modular arithmetic is, which is #9130.

Closes #9129

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekampZassenhaus/Classical/Candidate.lean
+++ b/HexBerlekampZassenhaus/Classical/Candidate.lean
@@ -83,9 +83,9 @@ def directCandidatePrefilter
     directTrailingPrefilter coreLc target modulus trailingResidue
 
 /-- The candidate computation itself, for a selection the cached prefilters have
-already accepted.  A caller that has run the prefilter — the streaming
-traversal does, so that it can decide whether to build `selected` at all — calls
-this rather than paying for the prefilter twice. -/
+already accepted.  The streaming traversal runs the prefilter itself, so that it
+can decide whether to build `selected` at all, and then calls this rather than
+paying for the prefilter twice. -/
 @[expose]
 def directCandidateAfterPrefilter
     (coreLc : Int) (target : ZPoly) (modulus : Nat) (selected : List ZPoly) :
--- a/HexBerlekampZassenhaus/Classical/Candidate.lean
+++ b/HexBerlekampZassenhaus/Classical/Candidate.lean
@@ -82,6 +82,20 @@ def directCandidatePrefilter
   directDegreePrefilter coreLc target degreeSum &&
     directTrailingPrefilter coreLc target modulus trailingResidue
 
+/-- The candidate computation itself, for a selection the cached prefilters have
+already accepted.  A caller that has run the prefilter — the streaming
+traversal does, so that it can decide whether to build `selected` at all — calls
+this rather than paying for the prefilter twice. -/
+@[expose]
+def directCandidateAfterPrefilter
+    (coreLc : Int) (target : ZPoly) (modulus : Nat) (selected : List ZPoly) :
+    Option (ZPoly × ZPoly) :=
+  let candidate := directCandidate coreLc modulus selected
+  if shouldRecordPolynomialFactor candidate then
+    (exactQuotient? target candidate).map fun quotient => (candidate, quotient)
+  else
+    none
+
 /-- Evaluate the candidate computation after the cached prefilters. -/
 @[expose]
 def tryDirectCandidate
@@ -89,11 +103,7 @@ def tryDirectCandidate
     (selected : List ZPoly) (degreeSum : Nat) (trailingResidue : Int) :
     Option (ZPoly × ZPoly) :=
   if directCandidatePrefilter coreLc target modulus degreeSum trailingResidue then
-    let candidate := directCandidate coreLc modulus selected
-    if shouldRecordPolynomialFactor candidate then
-      (exactQuotient? target candidate).map fun quotient => (candidate, quotient)
-    else
-      none
+    directCandidateAfterPrefilter coreLc target modulus selected
   else
     none
 
--- a/HexBerlekampZassenhaus/Classical/CombinationIterator.lean
+++ b/HexBerlekampZassenhaus/Classical/CombinationIterator.lean
@@ -19,6 +19,14 @@ selected and rejected prefixes in reverse, plus the cheap candidate statistics,
 and stops at the first exact divisor.  The proved classical search uses the
 head-forced iterator; the proposal tier uses the unforced low-cardinality
 iterator.
+
+Two properties keep a rejected support free of allocation.  `SupportMeta`
+records the lift modulus and each lifted factor's degree and trailing
+coefficient once, so a traversal step is an array read and one modular multiply
+rather than a fresh `p ^ k` and a fresh `Option`.  And every leaf runs its
+metadata-only filters before it reverses the selected indices, maps them to
+lifted polynomials, or concatenates the complementary support -- the last of
+which is built only once an exact divisor is in hand.
 -/
 
 namespace Hex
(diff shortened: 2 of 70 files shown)


kim-em:
Second-opinion review (Codex) found no correctness regression. It confirmed that `Hex.directLeaf_eq` is strong enough — both sides return `tried = 1` on every reached leaf and agree on every `DirectSplit` field, for arbitrary arguments rather than only traversal-reachable states — and that generalizing the iterator theorems over an arbitrary `Hex.SupportMeta` introduces no vacuity, since every traversal index has type `Fin basis.liftedFactors.size` and the proof fields fix the value read at each one.

It raised four points. Three are fixed in 635854b7:

- A surviving leaf evaluated `directCandidatePrefilter` twice, once in `directLeaf`'s guard and again as `tryDirectCandidate`'s first step. That is wasted work on a performance change, and it also meant the counted bench mirror, which runs the filters once, was no longer an operational mirror. `directCandidateAfterPrefilter` is now the shared post-filter body and both call it.
- `SupportMeta` gained `degrees_size` and `trails_size`. Without them a too-short array satisfies the elementwise `getD` equations whenever the omitted degree and trailing coefficient happen to be zero, so the docstring claimed more than the fields did.
- Two docstrings overclaimed: `LiftData` only requires `p > 0` and does not encode primality or the recovery bound, and a rejected leaf is not allocation-free — it still pays for the filters' own modular arithmetic. What it no longer pays for is the support representation.

The fourth is declined and now documented in the report: hoisting the metadata from per-level to per-search would have to thread it through `findDirectHead` and `searchDirectAux`, whose signatures the completeness proofs consume, to save 16 constructions against 32,768 leaves on `sd5`. A suggested rename to `DirectSupportMeta` is also declined — it pushed several `simp only` lists past the file's line width for no gain, and the project's naming guidance prefers the shorter form.

All measurements were re-taken at 635854b7 because the prefilter fix changes runtime. Swinnerton-Dyer end-to-end factor time is 0.721x to 0.853x, traversal outside the candidate test falls 40.2% on `sd5` and 46.0% on `sd5_x_phi11`, whole-factorization allocator time falls 30.6% and 30.2%, and every non-SD row is within 4.1%.

Issue #9142, opened 2026-08-02
BZ arithmetic: measure a subquadratic dense polynomial product for the Hensel lift


kim-em:
Tracked by #9126.

## Problem

After #9131 the Hensel lift's residual cost is dominated by a single kernel: `Hex.DensePoly.mulImpl`, the schoolbook dense polynomial product. Its inclusive share of total factor time, against the lift's own inclusive share, from `hexbz-factor-sampling-profiles-c34ffbbb-chungus2.json`:

| instance | `mulImpl` | `henselLiftFactors` | `mulImpl` as share of lift |
|---|---:|---:|---:|
| `cyclo_phi179` | 32.27% | 72.13% | **~45%** |
| `cyclo_phi385` | 15.57% | 35.67% | **~44%** |
| `wilkinson_56` | 22.88% | 55.62% | **~41%** |
| `xpow105_minus1` | 17.16% | 17.20% | shared with recombination |
| `xpow120_minus1` | 12.93% | 3.71% | mostly recombination |
| `xpow48_minus1` | 12.13% | 4.19% | mostly recombination |
| `sd5` | 6.04% | 1.62% | not a factor |

This satisfies the trigger #9131 reserved for its own Phase 3 -- "at least 25% of material lift time to polynomial multiplication" -- on the three Hensel-bound rows, and #9131 closed without acting on it, so the item is currently unowned.

Two scoping facts to keep the work honest:

- **Swinnerton-Dyer is not a target.** `sd5` spends 6.04% in `mulImpl`; its cost is traversal (`scanDirectCombinations` 85.58% against `tryDirectCandidate` 30.13%), which is #9129's. Do not score this work on SD rows.
- **The `x^n - 1` rows are a secondary, partial target.** Their `mulImpl` share is split between the lift and recombination's product materialization, so only part of it is reachable from a faster lift.

## The shape to optimize

`cyclo_phi179` lifts two factors of degree 89 to `3^113`. The recursion steps at moduli `3^1, 3^2, 3^4, 3^8, 3^15` -- all inside the `m * m < 2^64` word guard -- and then at `3^29` and `3^57`, which are not. The two bignum steps carry coefficients of about 92 and 181 bits and account for essentially all of the multiplication cost.

Per bignum step, `quadraticHenselStepBignum` performs **nine** full polynomial products and two monic divisions; the factor-only last step, `quadraticHenselFactorsBignum`, performs four products and one division. Every one goes through `DensePoly.mul`.

[...]

Issue #9128, opened 2026-08-02
BZ prime plans: scout modular degree patterns before full splitting


kim-em:
Tracked by #9126. Blocked by #9127.

## Problem

`HexBerlekampZassenhaus/Modular/PrimePlan.lean` currently performs full Berlekamp factorization at the first good prime and, whenever the modular width exceeds 8, at exactly two more good primes. It selects lexicographically by complete subset cost, reachable degrees, lift precision, and prime.

The combined-cactus cases show that the fixed rule is wrong in both directions:

- `xpow105_minus1` benefits materially: widths 30, 33, and 14 make the third prime worth finding;
- `sd5` sees 16 quadratic factors at all three primes, so the extra factorizations buy no combinatorial reduction;
- Wilkinson 56 remains 56 linear factors and changes chiefly in lift precision;
- `xpow120_minus1` sees widths 39, 64, and 42, leaving the first plan best;
- `cyclo_phi64_x_phi105` starts at width 10 and then pays for worse widths 14 and 12; 84% of its measured time is prime planning.

Simply reverting to the first good prime would lose the important `xpow105_minus1` win. We need a bounded downstream-cost decision.

## Proposed investigation

Prototype a mathematical `PrimeDegreePattern` obtained without fully splitting every rejected prime. The existing executable distinct-degree factorization in `HexBerlekamp/DistinctDegree.lean` is the first candidate: a degree-`d` bucket of polynomial degree `m` predicts `m / d` irreducible factors of degree `d`.

Measure before committing to this design. Distinct-degree factorization is useful only if scouting plus one selected full factorization is cheaper than the present repeated full splitting on the relevant rows.

Use the pattern to estimate:

1. the cost of another modular scout or full split;
2. selected prime and original-coordinate Hensel precision;
3. Hensel work from factor degrees and tree shape;
4. expected recombination nodes under the actual support enumeration;
5. expected exact-divisibility work.

Scan a small, deterministic, explicitly bounded set of good primes. Run full Berlekamp splitting only for the selected prime. If a scout is incomplete or unavailable, fall back to the existing verified full factorization path.

The scout is a performance heuristic, not a correctness oracle: the selected prime's actual factors and all existing certificates remain the source of mathematical correctness.

## Design requirements

- Use mathematical objects such as `PrimeDegreePattern` and `PrimePlan`; do not introduce `ScoutEngine`, `Provider`, `Dispatch`, or family-labelled policy objects.
[...]

PR #9148, opened 2026-08-03
feat: measure a packed finite-field matrix for Berlekamp's fixed space

### LINKED ISSUE #9132: Berlekamp factorization: measure a packed finite-field matrix representation
Tracked by #9126. Blocked by #9127 and #9128.

## Hypothesis

Current combined-cactus profiles provide a new candidate for specialization:

- `cyclo_phi385` spends about 43% in modular prime factorization, with roughly 40% under matrix row reduction/nullspace;
- `cyclo_phi64_x_phi105` shows roughly 50% under row reduction across its current prime walk;
- allocation is a major leaf category in both cases.

However, #9128 should first remove unnecessary full factorizations of rejected primes. The matrix share must be remeasured afterward. This issue has a valid no-go outcome if row reduction no longer dominates.

The earlier #9105 measurement on Legendre P30/P38 did not find one modular kernel above its 25% specialization threshold. That result was correct for those rows; it does not answer the newly measured cyclotomic cases.

## Measurement gate

After #9128 lands, attribute separately:

1. Berlekamp matrix column construction;
2. conversion to the generic matrix representation;
3. pivot search;
4. row swapping, scaling, and row addition;
5. rank/nullspace-basis construction;
6. conversion of basis vectors back to polynomials;
7. allocation bytes and peak residency.

Run on `cyclo_phi385`, `cyclo_phi64_x_phi105`, `cyclo_phi128_x_phi165`, `cyclo_phi179`, Wilkinson 40/48/56, Legendre P30/P38, Conway controls, and small matrices.

Proceed to production implementation only if contiguous finite-field matrix operations remain at least 25% of total factor time on a material row, or at least 40% of its selected-prime Berlekamp time, after planning overhead is removed.

## Candidate design

Define a `FiniteFieldMatrix p rows cols` with contiguous row-major word storage and an interpretation as the existing mathematical `Matrix (Fin rows) (Fin cols) (ZMod64 p)`.

Provide only the operations required by Berlekamp's fixed space:

- entry access;
- row swap;
- row scale by a nonzero field element;
- add a scalar multiple of one row to another;
- echelon reduction, rank, and nullspace basis.

The optimized operations may update a private mutable buffer, but the public computational result should be a mathematical matrix/basis with a simple invariant. Avoid a parallel hierarchy of generic and “fast” Berlekamp algorithms.

## Proof obligations

Prove:

- packed entry access agrees with the interpreted matrix;
- each elementary row operation preserves the expected row equivalence;
- the returned rank agrees with `Matrix.rowReduce_rank` on the interpreted matrix, or directly satisfies the rank characterization required by Berlekamp irreducibility;
- every returned nullspace vector is sound and the basis spans the full kernel;
- the polynomial witnesses supplied to `fullySplit` are the same mathematical fixed-space witnesses required by the existing soundness and completeness proofs.

Prefer a single correspondence theorem at the `fixedSpaceKernel` boundary so the large Berlekamp proof development continues to reason about ordinary matrices.

## Measurements after implementation

Repeat the phase table and full factor sweep. Report:

- matrix dimensions and selected prime;
- generic versus packed row-reduction/nullspace time;
- conversion and allocation cost;
- total Berlekamp and total integer-factorization time;
- every testcase regression above 5%.

## Acceptance criteria

- The measurement gate is met before a production specialization lands; otherwise close with the no-go report.
- The packed representation has proved correspondence, with no unchecked oracle, `axiom`, `sorry`, or `native_decide`.
- There is one public Berlekamp factorization algorithm and one fixed-space mathematical API.
- The selected cyclotomic rows improve materially without degrading small matrices or Legendre controls.
- Full row-reduction, Berlekamp, conformance, factorization, and oracle tests pass.
- Any implementation PR records a fresh full Hex sweep and regenerates all factorization cactus and runtime-by-degree plots, reusing unchanged external comparator records.

## Dependencies

Blocked by #9127 for the common attribution protocol and #9128 because prime planning changes how many matrices are constructed.

Its implementation or measured no-go feeds the final acceptance run in #9126. It does not block the support-cursor, modular-divisibility, or Hensel work.

kim-em:
This PR remeasures the Berlekamp fixed-space kernel after the prime-planning
change in #9128 and reports a **go** on the measurement threshold in #9132.

On `cyclo_phi385` the row reduction and nullspace basis at the selected prime are
46.5% of total factor time and 94.6% of that prime's Berlekamp split; on
`cyclo_phi128_x_phi165`, 44.6% and 89.3%; on `cyclo_phi64_x_phi105`, 29.2% and
79.2%; `cyclo_phi179` is 15.7% of total but 77.9% of its split. 9 of 24 rows
clear the 25%-of-total bar and 11 of 24 clear the 40%-of-split bar. Inside that
share the cost is one stage: on `cyclo_phi385` row addition is 302.049 ms of the
306.315 ms Gauss-Jordan run, while pivot search is 6.506 us, the nullspace basis
1.987 ms, and the conversion of basis vectors back to polynomials 7.290 us.

A prototype packed representation, one contiguous row-major buffer with the
modulus in a machine-word local, cuts that row reduction from 312.206 ms to
15.606 ms, a 17.0x speedup including the cost of packing, and projects a
1.73x end-to-end improvement on `cyclo_phi385`, 1.68x on `cyclo_phi128_x_phi165`,
1.36x on `cyclo_phi64_x_phi105`, and 1.14x on `cyclo_phi179`, with nothing
regressing.

Add `--entry kernelProfile` to `hexbz_factor_service`, which walks one kernel
computation stage by stage through the production functions and prices four
packed variants against it, and extend `scripts/bench/factor_phase_profile.py` to
sweep it. Every packed result is checked entry for entry against
`Hex.Matrix.nullspace` on all 24 instances before its time is reported, and the
counted Gauss-Jordan mirror is checked against the production
`Hex.Matrix.rowReduce`, matching it to a median wall ratio of 1.019; the driver
exits non-zero on any disagreement.

The numerator counts the row reduction and basis alone. Berlekamp matrix
construction is a polynomial computation that a packed matrix does not
accelerate, and it is 2.9% of the kernel on `cyclo_phi385` but 85% on
Wilkinson 56, so counting it would inflate exactly the work under discussion.
The in-cascade kernel is estimated as the in-cascade Berlekamp split times the
kernel's share of it, that share being a ratio between two spans of one
kernel-section execution; differencing spans across the two sections is not
robust, and on a contended run put the kernel at 146% of the split containing it.

The measurements narrow the design from the issue's sketch: build the packed
matrix directly from the column polynomials rather than converting a generic one,
store `UInt32` entries (the modulus is already capped below `2^31`, and an
immediate `UInt32` stops the inner loop allocating, which is worth 2.41x), skip
the `n x n` transform that the nullspace never reads (1.98x on its own), and
specialize the whole reduction loop rather than a chosen subset, because swap and
scale become dominant once row addition is accelerated.

Four measurement hazards are recorded in the report because they would silently
corrupt a re-measurement: compiler let-floating moving a stage's cost into
whichever later stage first demands its value, common-subexpression elimination
defeating a scalar forcing barrier so that the second and later matrix rebuilds
measured about 130 ns, aliasing defeating the in-place buffer update that
`Hex.Matrix`'s elementary operations rely on, and `Vector` operations
re-demanding the length index `m - rowReduce_rank M` and so running a second full
row reduction.

An independent review by Codex found three defects in the first version of this
measurement, all confirmed and fixed here: the packed loops skipped zero source
entries that the generic path does not skip (worth 1.69x, now priced as a
separate variant), the rebuild barrier did not survive common-subexpression
elimination, and the threshold numerator counted matrix construction. The report
carries a section stating what the measurements do not establish, including that
the two sections supplying numerator and denominator are not paired observations
and that the corpus moduli are far below the `2^31` bound.

This does not close #9132: the threshold is met, so the production specialization
it authorizes is still owed, and is filed as #9150.

🤖 Prepared with Claude Code

### DIFF
--- a/bench/HexBench/BerlekampKernel.lean
+++ b/bench/HexBench/BerlekampKernel.lean
@@ -282,6 +282,10 @@ inductive PackedKind where
   | halfWord
   /-- `UInt32` words, hardware remainder instead of a Barrett reciprocal. -/
   | halfWordDiv
+  /-- `UInt32` words and a Barrett reciprocal, additionally skipping the
+  multiply where the source-row entry is already zero. Prices an algorithmic
+  optimization the generic path does not have. -/
+  | halfWordSkipZero
 deriving Repr, DecidableEq, BEq
 
 /-- Barrett reciprocal `floor(2^64 / p)` for a modulus below `2^32`. -/
@@ -346,8 +350,15 @@ structure PackedCounters where
   rank : Nat := 0
 
 /-- Packed Gauss-Jordan over `UInt32` storage, returning the echelon buffer and
-the pivot columns. Full-width row additions, matching the generic `rowAdd`. -/
-private def reduce32 (useBarrett : Bool) (p : UInt32) (n m : Nat)
+the pivot columns.
+
+Row additions are full width, matching the generic `Hex.Matrix.rowAdd`, which
+evaluates `x + c * rsrc[k]` at every entry. `skipZero` additionally skips the
+multiply and the store where the source entry is already zero -- an algorithmic
+optimization the generic path does not have, priced separately so it is not
+counted as a gain from the representation. Both variants skip a whole row
+addition when its coefficient is zero, which the generic path also does. -/
+private def reduce32 (useBarrett skipZero : Bool) (p : UInt32) (n m : Nat)
     (buf : Array UInt32) : Array UInt32 × Array Nat × Nat :=
   Id.run do
     let p64 := p.toUInt64
@@ -380,11 +391,10 @@ private def reduce32 (useBarrett : Bool) (p : UInt32) (n m : Nat)
       if inv32 != 1 then
         for k in [0:m] do
           let x := a[row * m + k]!
-          if x != 0 then
-            let prod :=
-              if useBarrett then barrett p64 pinv (x.toUInt64 * inv)
-              else (x.toUInt64 * inv) % p64
-            a := a.set! (row * m + k) prod.toUInt32
+          let prod :=
+            if useBarrett then barrett p64 pinv (x.toUInt64 * inv)
+            else (x.toUInt64 * inv) % p64
+          a := a.set! (row * m + k) prod.toUInt32
       -- eliminate the column
       for j in [0:n] do
         if j != row then
@@ -395,7 +405,7 @@ private def reduce32 (useBarrett : Bool) (p : UInt32) (n m : Nat)
             let neg64 := neg.toUInt64
             for k in [0:m] do
               let s := a[row * m + k]!
-              if s != 0 then
+              if !skipZero || s != 0 then
                 let prod :=
                   if useBarrett then barrett p64 pinv (s.toUInt64 * neg64)
                   else (s.toUInt64 * neg64) % p64
@@ -436,8 +446,7 @@ private def reduce64 (p : UInt64) (n m : Nat) (buf : Array UInt64) :
       if inv != 1 then
         for k in [0:m] do
           let x := a[row * m + k]!
-          if x != 0 then
-            a := a.set! (row * m + k) (barrett p pinv (x * inv))
+          a := a.set! (row * m + k) (barrett p pinv (x * inv))
       for j in [0:n] do
         if j != row then
           let c := a[j * m + col]!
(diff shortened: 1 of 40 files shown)


kim-em:
Heads-up from the agent on #9147: your PR body is also being written to https://github.com/kim-em/hex-dev/pull/9147, which is a different PR (the Kronecker Hensel-product change, already merged). Two attempts to restore #9147's own description were overwritten within seconds, so something in this session's tooling is holding a stale PR number — most likely one captured before #9148 was created, since #9147 was opened moments earlier.

Nothing of yours is lost: #9148's body is correct. But the overwrite stripped #9147's `Closes #9142` line, so that issue did not auto-close (I closed it manually). Worth checking the PR number your publish step resolves before it edits again.

PR #9158, opened 2026-08-03
perf: pack the Berlekamp fixed-space matrix into UInt32 words

### LINKED ISSUE #9150: Berlekamp factorization: implement the packed finite-field matrix
Follows the measured **go** in #9132. The measurement is landed in
https://github.com/kim-em/hex-dev/pull/9148 "feat: measure a packed finite-field
matrix for Berlekamp's fixed space"; the evidence is
[reports/hexbz-packed-fp-matrix.md](https://github.com/kim-em/hex-dev/blob/main/reports/hexbz-packed-fp-matrix.md)
reading `reports/bench-results/hexbz-phase-profile-0e1601a0-chungus2.json`.

No `depends-on`: the measurement this follows has landed on `main`. #9132 stays
open only to track that this specialization is owed, so depending on it would
deadlock both.

## Current state

`Hex.Berlekamp.fixedSpaceKernelVectors` is `Hex.Matrix.nullspace (fixedSpaceMatrix f hmonic)`.
On `cyclo_phi385` (240x240 at `p = 3`) the row reduction and nullspace basis are
46.1% of total factor time and 94.3% of the selected prime's Berlekamp split,
and row addition alone is 98.7% of the Gauss-Jordan run. A measured prototype
gets the reduction from 311.525 ms to 15.761 ms, 16.8x including packing, and
projects 1.72x end to end on that row, 1.67x on `cyclo_phi128_x_phi165`, 1.33x on
`cyclo_phi64_x_phi105`, and 1.13x on `cyclo_phi179`.

The prototype lives in `bench/HexBench/BerlekampKernel.lean` (`reduce32`,
`reduce64`, `basisOfBuffer32`) and is checked entry for entry against
`Hex.Matrix.nullspace` on 24 instances. It is a benchmark, not a design: it
carries no proofs and uses `Array.get!`/`Array.set!`.

## Deliverables

A packed finite-field matrix behind the `fixedSpaceKernelVectors` boundary, with
proved correspondence and no `axiom`, `sorry`, `native_decide`, or unchecked
oracle, and one public Berlekamp factorization algorithm.

The measurements fix five design points; do not re-litigate them without new
evidence.

1. **Build the packed buffer directly from the Berlekamp column polynomials.**
   Do not convert an already-built `Hex.Matrix`. The conversion stage is 1.256 ms
   and the pack step 2.819 ms on `cyclo_phi385`, and going direct is also what
   removes the only measured overhead on rank-0 rows.
2. **`UInt32` entries.** `ZMod64.Bounds` caps the modulus at `2^31`. This is
   worth 2.42x, and the reason is boxing, not storage density: `lean_box_uint32`
   is a tagged immediate on a 64-bit runtime while `lean_box_uint64` allocates.
   `Array` holds one machine-word slot per element either way.
3. **No transform.** `Hex.Matrix.rowReduce` maintains an `n x n` transform in
   lockstep that `nullspace` never reads; dropping it is 2.00x with no change of
   representation.
4. **Specialize the whole reduction loop** -- pivot search, swap, scale,
   elimination, basis readback. Swap and scale are 3.926 ms against a packed
   reduction of 15.761 ms, so anything left generic dominates afterwards.
5. **No Barrett context** unless new evidence says otherwise. It measures 1.01x
   against a hardware remainder, on a corpus whose primes are 3 to 79. Leaving it
   out also keeps the reciprocal's proof obligations out of the correspondence.

The zero-entry skip (1.70x on `cyclo_phi385`) is orthogonal and should be decided
separately, on evidence that includes dense matrices where it buys nothing.

## Proof obligations

- packed entry access agrees with the interpreted `Matrix (ZMod64 p) n n`;
- each elementary row operation on the buffer interprets to the corresponding
  `HexMatrix.Elementary` operation;
- pivot search agrees with the reference;
- the returned rank agrees with `Hex.Matrix.rowReduce_rank` on the interpreted
  matrix;
- every returned nullspace vector is sound and the basis spans the full kernel;
- the polynomial witnesses reaching `fullySplit` are the same mathematical
  fixed-space witnesses the existing soundness and completeness proofs require.

Land these as **one** correspondence theorem at the `fixedSpaceKernelVectors`
boundary, so the Berlekamp development keeps reasoning about ordinary matrices,
and select the fast path with a `@[csimp]`-proved equality rather than
`@[implemented_by]`.

## Traps the measurement hit

Any re-measurement will hit these. All four are documented in the report.

- **Let-floating.** A pure `let` first used later is moved to that use, so a
  stage's cost lands in whichever later stage first demands it.
- **CSE through a scalar forcing barrier.** Pushing a projection into an
  `IO.Ref` is not enough when the same pure expression appears twice; give each
  force a runtime-varying operand.
- **Aliasing.** `Hex.Matrix`'s elementary operations update the flat buffer in
  place only while it is uniquely referenced. A shared matrix costs 2.742 s
  against 304 ms.
- **The length index.** `Vector` operations take the length as a runtime
  argument, and for this kernel it is `m - Hex.Matrix.rowReduce_rank M`;
  demanding it runs a second full row reduction.

## Verification

- `lake build`, `lake build HexConformance`, row-reduction and Berlekamp tests,
  conformance fixtures, and the external oracles.
- `python3 scripts/bench/factor_phase_profile.py` with the kernel section, whose
  packed variants must still agree with `Hex.Matrix.nullspace`.
- A fresh full Hex sweep and regenerated factorization cactus and
  runtime-by-degree plots, reusing the unchanged external comparator records.
- Report every testcase regression above 5%, and re-measure the five design
  factors above against the integrated implementation rather than quoting the
  prototype's.

kim-em:
This PR runs Berlekamp's fixed-space kernel on a packed `UInt32` matrix behind an unchanged `fixedSpaceKernelVectors` boundary, selected in compiled code by a `@[csimp]`-proved equality so the Berlekamp soundness and completeness proofs keep reasoning about ordinary `Matrix (ZMod64 p)`. On `cyclo_phi385` the kernel goes from 327.823 ms to 53.886 ms, 6.08x, and the whole factorization from 434.641 ms to 167.124 ms, 2.60x; no row of the 24-row representative set regresses above 5% on either span. Follows the measured go in #9132 and the evidence in `reports/hexbz-packed-fp-matrix.md`.

`HexBerlekamp/PackedKernel.lean` holds the development. The packed matrix is a `Hex.Matrix UInt32 n m`, not a bare `Array`: reusing `HexMatrix`'s flat row-major backing means `Matrix.rowSwap`, `Matrix.modifyEntries` and their entrywise read lemmas are already proved, and the in-place update on a uniquely referenced buffer is the one the generic path already relies on. `Packed.Rep A E` says a packed word and a residue have the same `toNat`, which carries reducedness for free because residues are canonical. `Packed.reduceLoop_sim` is one induction relating the packed loop to `Matrix.rowReduceLoop`, with the reference transform carried on the right of the relation and never inspected -- which is what makes dropping it sound. `Packed.fixedSpace` fills the buffer straight from `berlekampColumnPolys`, so no `Matrix (ZMod64 p)` is built on the fast path, and `Packed.kernelArray_eq` is the single correspondence theorem. No `axiom`, `sorry`, or `native_decide`.

One part of the directive's premise did not survive contact. `fixedSpaceKernelVectors` took an opaque `[Lean.Grind.Field (ZMod64 p)]`, and `Lean.Grind.Field` bundles `Add`, `Mul` and `Inv` as structure fields, so `rowReduceLoop`'s arithmetic is whatever dictionary the caller supplies. A packed word only stands for a residue under the canonical `ZMod64` arithmetic, so no packed implementation is correct for every dictionary and the `@[csimp]` as specified would be false. The Berlekamp kernel path is therefore now parameterised by `[ZMod64.PrimeModulus p]`; every call site already had a primality witness in scope and was already building `zmod64FieldOfPrime` from it, and `PrimeModulus` is a `Prop` class, so this strengthens an existing hypothesis without passing anything new at runtime. `fixedSpaceKernel` also had to move below the `@[csimp]`, since csimp only rewrites code generated after the attribute is declared.

`reports/hexbz-packed-fp-matrix-integration.md` measures the landed implementation rather than the prototype, over all 24 representative rows, and states two limits plainly: the implementation is 2.6x slower than the benchmark prototype on the rows where the reduction dominates (`Matrix.rowAdd` copies the source row, the pivot columns are a boxed list, the basis readback mirrors the reference's `pivotIndex?` scan), and design points 2 and 5 -- `UInt32` versus `UInt64` entries, Barrett versus the hardware remainder -- are not re-measured against this build, because separating them needs compiled variants nothing would ship.

Verified: full `lake build` and `lake build HexConformance` green with no new warnings; `hexberlekamp_emit_fixtures` and `hexbz_emit_fixtures` reproduce the committed fixtures byte for byte; the phase profile's own cross-checks agree with `Hex.Matrix.nullspace` and the production `factorTrace` on 368 of 368 instances; a fresh 392-instance sweep with the external comparator records reused unchanged, all 25 figures regenerated, and `check_factor_sweep_freshness.py` and `hexbz-cactus.py --check` passing.

Closes #9150

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekamp.lean
+++ b/HexBerlekamp.lean
@@ -6,7 +6,7 @@ Authors: Kim Morrison
 
 module
 
-public import HexBerlekamp.BerlekampMatrix
+public import HexBerlekamp.PackedKernel
 public import HexBerlekamp.CertificateSyntax
 public import HexBerlekamp.Factored
 public import HexBerlekamp.PolynomialTactic
--- a/HexBerlekamp/BerlekampMatrix.lean
+++ b/HexBerlekamp/BerlekampMatrix.lean
@@ -680,23 +680,10 @@ Berlekamp-local linear-algebra implementation.
 -/
 @[expose]
 def fixedSpaceKernelVectors (f : FpPoly p) (hmonic : DensePoly.Monic f)
-    [inst : Lean.Grind.Field (ZMod64 p)] :
+    [ZMod64.PrimeModulus p] :
     Vector (Vector (ZMod64 p) (basisSize f))
       (basisSize f - Matrix.rowReduce_rank (fixedSpaceMatrix f hmonic)) :=
   Matrix.nullspace (fixedSpaceMatrix f hmonic)
-
-/-- The fixed-space kernel basis converted back to polynomial representatives.
-
-Compute the vector basis once and map the conversion over it. Keep the basis
-outside any per-index body so matrix construction and row reduction remain
-shared; mapping also carries its dependent length without recomputing rank. -/
-@[expose]
-def fixedSpaceKernel (f : FpPoly p) (hmonic : DensePoly.Monic f)
-    [inst : Lean.Grind.Field (ZMod64 p)] :
-    Vector (FpPoly p)
-      (basisSize f - Matrix.rowReduce_rank (fixedSpaceMatrix f hmonic)) :=
-  let vectors := fixedSpaceKernelVectors f hmonic
-  vectors.map vectorToPoly
 
 /-- Vector-level executable Berlekamp kernel condition for the fixed-space
 matrix `Q_f - I`. -/
@@ -747,21 +734,6 @@ theorem fixedSpaceKernelVectors_sound (f : FpPoly p) (hmonic : DensePoly.Monic f
     IsFixedSpaceKernelVector f hmonic ((fixedSpaceKernelVectors f hmonic).get k) := by
   unfold IsFixedSpaceKernelVector fixedSpaceKernelVectors
   exact Matrix.nullspace_sound (fixedSpaceMatrix f hmonic) k
-
-/-- Every polynomial representative returned by `fixedSpaceKernel` satisfies the
-executable fixed-space kernel condition. -/
-theorem fixedSpaceKernel_sound (f : FpPoly p) (hmonic : DensePoly.Monic f)
-    [ZMod64.PrimeModulus p]
-    (k : Fin (basisSize f -
-      Matrix.rowReduce_rank (fixedSpaceMatrix f hmonic))) :
-    IsFixedSpaceKernelPolynomial f hmonic ((fixedSpaceKernel f hmonic).get k) := by
-  have hk : (fixedSpaceKernel f hmonic).get k =
-      vectorToPoly ((fixedSpaceKernelVectors f hmonic).get k) := by
-    unfold fixedSpaceKernel
-    exact Vector.getElem_map vectorToPoly k.isLt
-  unfold IsFixedSpaceKernelPolynomial
-  rw [hk, coeffVector_vectorToPoly]
-  exact fixedSpaceKernelVectors_sound f hmonic k
 
 /-- Every vector satisfying the executable fixed-space kernel condition is a
 linear combination of the public nullspace-basis matrix for `Q_f - I`. -/
--- a/HexBerlekamp/Factor.lean
+++ b/HexBerlekamp/Factor.lean
@@ -965,7 +965,6 @@ splitting preserves `factorProduct` without using square-freeness, so the
 product equality holds for every monic input. -/
 theorem factorProduct_berlekampFactor
     [ZMod64.PrimeModulus p]
-    [ZMod64.PrimeModulus p]
     (f : FpPoly p) (hmonic : DensePoly.Monic f) :
     factorProduct (berlekampFactor f hmonic).factors = f :=
   factorProduct_fullySplit ((fixedSpaceKernel f hmonic).toList) (f.size + 1) f
@@ -976,7 +975,6 @@ product of the returned factors for any monic input.
 -/
 theorem prod_berlekampFactor
     (f : FpPoly p) (hmonic : DensePoly.Monic f)
-    [ZMod64.PrimeModulus p]
     [ZMod64.PrimeModulus p] :
     (berlekampFactor f hmonic).product = f := by
   simp only [Factorization.product_def]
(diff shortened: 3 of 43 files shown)


kim-em:
Second opinion (Codex) came back with no correctness defect in the packed kernel or the simulation proof, and independently confirmed both structural claims: that a `@[csimp]` at this boundary is false for an arbitrary `Lean.Grind.Field (ZMod64 p)` dictionary (it noted the same construction I had in mind -- transport the canonical field across a permutation of the residues), and that `@[csimp]` does not retroactively regenerate earlier definitions, which it checked by reading the generated C.

Its three findings are all taken, in `a5e0ebe4`:

- eight `Factor.lean` declarations carried a duplicated `[ZMod64.PrimeModulus p]` binder, left by the mechanical replacement of `[Field] [PrimeModulus]`;
- `primeModulusOfNatPrime` was a proposition-valued `def` that only renamed `ZMod64.primeModulusOfPrime`; it is gone and the three call sites use that directly;
- `berlekampRankTest` and its spec never reach the fixed-space kernel and gain nothing from the packed path, so their opaque `[Lean.Grind.Field (ZMod64 p)]` binder is restored. The narrowing now covers only what the packed fast path needs.

Codex also observed that moving `fixedSpaceKernel` below the rule was not the only option -- a second `@[csimp]` on `fixedSpaceKernel` itself, declared later, would work too. I kept the move: one rule at one boundary is the shape the issue asks for.

Re-measured at the reviewed revision (`65c8caad`), so the report's tables, hashes and revision all refer to the code as it now stands: `cyclo_phi385` kernel 327.823 ms → 54.057 ms (6.06x), total 434.641 ms → 165.854 ms (2.62x), still no regression above 5% on any of the 24 rows.

One correction to the report while re-measuring: the earlier draft read a 5.6% improvement out of the full-sweep solved-instance total. That does not survive a second run -- two sweeps of the same code differ by 6.4% between themselves, and the new total (20.847 s) is indistinguishable from the baseline's 20.684 s. The sweep is an anti-regression check across all 392 instances and the freshness record the figures are drawn from, not a measurement of the gain; the report now says so.

Issue #9127, opened 2026-08-02
BZ cactus elbow: record a phase-attributed baseline and counterfactual prime plans


kim-em:
Tracked by #9126.

## Problem

The combined cactus crossover spans solved ranks 125--140. Sampling has identified several plausible costs, but implementation work needs a durable, reproducible attribution record rather than conclusions copied from one profiler session.

The current Hex record is `reports/bench-results/hexbz-factor-sweep-1de5a116-hex-chungus2.json`; the current comparator record is `reports/bench-results/hexbz-factor-sweep-aa68c920-chungus2.json`. The cactus curve sorts each system's times independently, so the diagnostic must preserve both paired testcase data and cumulative-rank data.

## Work

Add a benchmark-only diagnostic, or zero-cost-when-disabled tracing surface, that records:

1. each good-prime candidate, its factor-degree pattern, and whether it was selected;
2. modular-image/good-prime, distinct-degree, Berlekamp matrix, row-reduction, and splitting time where applicable;
3. Hensel precision, tree shape, word/bignum work, and reduction time;
4. recombination nodes, cheap-filter rejections, candidate products materialized, exact divisions attempted, and successful divisors;
5. allocation counts/bytes when the platform tooling supports them;
6. total factor time and the corresponding cumulative cactus rank.

For bounded prime walks, also record counterfactual downstream measurements for stopping at each candidate. It is acceptable to collect expensive counterfactuals only on the representative set, provided the estimator is validated on a wider sample.

Run on at least:

- `sd5`, its shifts, `sd4_x_sd4shift1`, and `sd5_x_phi11`;
- `xpow48_minus1`, `xpow105_minus1`, and `xpow120_minus1`;
- `cyclo_phi179`, `cyclo_phi64_x_phi105`, `cyclo_phi128_x_phi165`, and `cyclo_phi385`;
- Wilkinson degrees 40, 48, and 56;
- Chebyshev/Legendre, easy cyclotomic, sparse, and random controls.

Capture symbolized sampling profiles for `sd5`, `sd5_x_phi11`, `xpow120_minus1`, `cyclo_phi179`, `cyclo_phi64_x_phi105`, `cyclo_phi385`, and Wilkinson 56.

## Required artifact

Commit a report containing:

- exact revision, Lean toolchain, host, CPU pinning, warmup/repetition protocol, and corpus hash;
- the phase and counterfactual tables;
- exact ranks 118--144 for Hex and the paired Isabelle times;
[...]

PR #9171, opened 2026-08-04
perf: price BZ prime scouting against attainable downstream savings

### LINKED ISSUE #9156: BZ prime plans: price scouting against attainable downstream savings
## Objective

Replace the fixed prime-scouting threshold with a deterministic, shape-based policy that only pays for another modular factorization when its attainable Hensel/recombination saving justifies the cost.

This is a general planner improvement, not a benchmark-name recognizer.

## Why this is now the leading planning question

`chooseDirectPrimePlanCore` currently uses a fixed `scoutWidth = 8` and `scoutFuel = 2`. Candidate plans are compared by a lexicographic downstream score, but the score does not price the work already spent obtaining the scout or the value of obtaining one more observation.

Fresh current-main counterfactuals show both failure modes:

- `cyclo_phi64_x_phi105`: the first retained prime, `p = 11`, has width 10 and costs about 8.77 ms downstream. The planner nevertheless scouts `p = 13` and `p = 17`; their downstream costs are about 243.52 ms and 37.40 ms. The first prime was already the best, while scouting accounts for about 46% of the 57.69 ms sample.
- `sd5`: `p = 19`, `p = 23`, and selected `p = 29` all have width 16 and downstream costs 63.454, 63.232, and 63.496 ms. The extra observations do not buy a meaningful downstream improvement.
- `xpow105`: scouting is essential. The first prime (`p = 11`, width 30) costs about 417 ms downstream, `p = 13` is worse, and the selected `p = 17` plan (width 14) costs about 17.26 ms downstream.
- `legendre_P30`: the first prime (`p = 61`, width 15) costs about 29.7 ms downstream; `p = 67` (width 7) costs about 1.42 ms, while a further `p = 71` plan (width 4) costs about 0.50 ms. A fixed stop threshold can stop before a still-valuable observation.

Thus neither “always scout” nor simply raising `scoutWidth` is sound as a performance policy.

## Investigation first

Before changing production selection:

1. Record, for every considered prime, the cost of probing/scouting, the resulting degree pattern and width, the plan selected by each simulated policy, and counterfactual downstream Hensel/recombination time.
2. Run the full current corpus under the repository timing protocol on an idle selected core, with enough repetitions to separate policy effects from noise.
3. Compare at least:
   - the current fixed width/fuel rule;
   - stop after the first acceptable prime;
   - a bounded value-of-information rule;
   - simple degree/field-size cost models as controls.
4. Keep `xpow105` and mixed/non-split cases as mandatory negative controls: an early-stop policy that loses the large useful scout is not acceptable.

## Candidate design

Introduce one explicit planner decision that compares:

- an estimate or measured bound for obtaining the next degree pattern/full split; and
- a conservative bound on the downstream saving that a better width/pattern can still attain.

The rule must be deterministic and depend only on mathematical/input shape and observations already made. Benchmark names, corpus identities, and family recognizers are forbidden. Keep one planner implementation and one proof-facing plan object.

Prime choice does not change factorization correctness, but all existing `DirectPrimePlan` facts and retained-prime certificates must remain proved after the refactor.

## Acceptance

Proceed from measurement to production only if a single auditable rule:

- reduces prime-walk time by at least 20% and total time by at least 10% on `cyclo_phi64_x_phi105`;
- preserves the order-of-magnitude scouting benefit on `xpow105`;
- does not materially regress the full corpus or reduce solved coverage;
- improves, or at least does not regress, the combined-cactus elbow; and
- passes a fresh all-systems performance sweep and regenerates every cactus plot in the same PR.

If no such policy survives the controls, publish the measurements and close this as a measured no-go rather than landing a family-specific heuristic.

## Dependencies

- Follows the planner cleanup in #9128.
- Can be investigated in parallel with #9150 and #9133.
- Should be resolved before judging the final acceptance criterion in #9126.
- If this leaves a material `powModMonicAux` cost only on cases where scouting is demonstrably valuable, open a separate kernel issue from the post-change profile rather than folding a second optimization into this PR.

kim-em:
This PR replaces the direct prime planner's fixed `scoutWidth = 8` stop threshold with `Hex.scoutPays`, one deterministic comparison between the recombination work the plan in hand may still have to do and the scouts plus Berlekamp split the rest of the walk may spend. Both sides are estimated from shape already observed -- the input degree, the primes involved, and the degree multisets of the plans in hand -- and both carry a factor of the modular degree squared, which cancels, leaving `min(2^(w-1), defaultSubsetBudget) * W > bitLen q * (scoutRoundCost * fuel * d + splitColumnCost * n)`. `scoutPays` is the walk's only stopping decision: it governs the first good prime and every scouted candidate alike, and no benchmark name, corpus identity or family recognizer enters it.

Both sides are worst cases, so passing means the walk *could* pay for itself rather than that it will. That makes this a bounded affordability gate, weaker than a value-of-information rule, and the report says so rather than overselling it.

Running both binaries over the whole corpus in counterbalanced AB/BA blocks finds seven instances whose walk the rule changes, and all seven improve: `cyclo_phi105_x_phi128` 0.409x, `cyclo_phi64_x_phi105` 0.414x, `sd4_x_phi17` 0.441x, `legendre_P18` 0.598x, `legendre_P20` 0.667x, `cyclo_phi24_x_phi35` 0.753x and `cyclo_phi275` 0.792x, against a load control of 0.985x to 1.003x. Solved coverage is the same 377 of 392 instances in both arms of every block with none lost or gained, no instance above 1.10x costs more than 2 ms, and the combined-cactus cumulative curve improves at every rank from 118 to 145. The mandatory `xpow105_minus1` control keeps its plan exactly and with it an order of magnitude: the `first` policy costs 415.564 ms there against 43.949 ms.

Every one of the eight decisions the walk takes on those seven rows is confirmed by measurement, against the best *attainable* net gain -- the incumbent's measured downstream less the cheapest downstream any candidate still reachable within the remaining fuel offers, less every scout and split needed to reach it. Three of the seven have no better prime reachable at all, so the observation the threshold bought could not have paid at any price; `cyclo_phi275` is the clearest, with three candidates all of width 10 and largest factor degree 20 whose downstreams are 613.5, 639 and 613 ms. Of the 39 decisions on the issue #9127 representative set, 23 are unconfirmed and every one is an *acceptance* the fixed threshold also made, at incumbent widths from 14 to 56. So the change removes seven observations that could not pay and introduces none.

The two cost ratios are calibrated medians and four records of the same source disagree about them (10.2/5.2, 12.9/4.3, 13.4/4.4, 13.3/4.4), so they are not precise. What is claimed instead is that 113 of 120 ratio pairs over `scoutRoundCost` 6 to 20 and `splitColumnCost` 2 to 9 reproduce the shipped pair's whole replayed walk on all 30 recorded instances, with every exception at or below the smallest measured `splitColumnCost` and every one worse than shipped. `scripts/bench/prime_policy_replay.py` also replays the eight policies the issue asks to be controlled against, prints every `scoutPays` decision against what it was worth, and confirms on 24 of 24 rows that the replayed rule reproduces the prime the measured binary selected.

`DirectPrimePlan` is unchanged and both facts about it stay proved, one induction case shorter each. Capping the recombination estimate needed `defaultSubsetBudget` visible to both the engine that spends it and the planner that prices it, so it moved from `Classical/Search.lean` up to `FactorizationResult.lean`. Conformance pins both directions of the decision, the modulus width it reads and the budget cap. All 25 cactus figures are regenerated, `scripts/plots/hexbz-cactus.py --check` passes byte for byte, and the sweep is fresh at this revision from a clean tree.

The branch was rebased onto #9172 part-way through, which speeds up completely split finite-field inputs; every number above was re-measured afterwards with the before arm built from that commit, so the comparison isolates this change.

Two measurement instruments came out of this and are committed: `scripts/bench/factor_sweep_paired.py`, because a single-pass corpus sweep is not a before/after instrument on a shared host (it put the long Swinnerton-Dyer rows at 1.25x to 1.31x where the paired driver puts them inside its control band), and counterbalanced blocks with paired within-block ratios in `scripts/bench/prime_plan_paired.py`.

`reports/hexbz-prime-plan-pricing.md` carries the measurements, the derivation, the policy comparison, and a section on what the work does not establish.

Closes #9156

🤖 Prepared with Claude Code


kim-em:
Ran a second opinion (OpenAI Codex) on the model and the measurement methodology. It found eight issues, five of them material, and I have acted on all of them. Recording the exchange here because two of its findings were about claims in the report that were simply false.

**Corrected in the model.** `liftWords` estimated the Hensel modulus width as `precision * bitLen p`, which overshoots by up to a bit per power; three committed rows came out a machine word too wide (`xpow105_minus1` at `p = 17, k = 26` was modelled as three words where `17^26` occupies two). It now forms the modulus and takes its bit length. Separately, the recombination estimate counted the complete head-forced subset search, but the direct engine abandons at `defaultSubsetBudget = 262144`, so beyond width 19 the estimate claimed work production never performs; it is now capped there, which needed the budget constant moved above both the engine and the planner in the import graph. Neither changes a decision on the recorded corpus, but the inequality now means what it says.

**Corrected in the claims.** Three statements were wrong and are gone:

- *"any common rescaling of the two constants leaves every decision unchanged."* False — the inequality counts a recombination candidate as one, so scaling only the modular side does change decisions.
- *"120 of 120 ratio pairs give the identical decision."* That was measured on the 24-row representative record and only on the walk's *first* decision. Over both records (29 instances) and the full replayed walk it is **116 of 120**. The four exceptions all sit at `splitColumnCost = 2`, below half the measured value, and all four are worse than the shipped pair; the replay now prints which rows move and what the move costs.
- *"no function of the degree multiset can separate Wilkinson from `sd5`."* Too strong. What is true is that no degree-only rule can determine search *termination* exactly. `maxDegree = 1` and `w = n` are real signals and an expected-cost model could use them; the report now says that.

**Also corrected.** The incumbent's largest factor degree is described as a proxy for the next candidate's scout depth, not a bound — the records contain counterexamples where the next candidate is 7x deeper (`legendre_P30`, 2 to 14). The rule is described as a bounded **affordability** gate rather than a value-of-information rule: it compares an upper bound on benefit with an upper bound on contingent cost, so passing means the walk *could* pay for itself, not that it will. The counterfactual's `downstreamNanos` is documented as the direct route only, excluding proposal, lattice and trial fallback.

**Measurement protocol.** The paired driver ran before-then-after in every block, confounding arm with position, and summarised a ratio of independently pooled per-arm medians. It now runs counterbalanced AB/BA blocks and reports the median of *within-block* ratios. Re-measuring now; the report's drift narrative goes with it.

Codex agreed the six large paired wins (0.41x to 0.79x against controls near 1.0) are almost certainly real and much larger than any of these flaws. Its one suggestion I did not take is replacing the scout-depth proxy with the safe bound `n/2`: on `cyclo_phi24_x_phi35`, the row where scouting genuinely pays 14.7 ms for a 2.6 ms observation, that bound makes the decision a 1.07x near-tie, so it trades a documented proxy for a real risk of losing wins. The report says so rather than claiming the proxy is sound.

PR #9172, opened 2026-08-04
perf: extract completely split finite-field inputs by their roots

### LINKED ISSUE #9157: Berlekamp factorization: extract fully split inputs by their roots
## Objective

Add a proved, general fast path for squarefree finite-field polynomials that split completely into linear factors: enumerate their roots and extract the corresponding monic linear factors instead of constructing a Berlekamp basis and recursively splitting with gcds.

The path must be selected from the finite-field polynomial itself. It must not recognize Wilkinson, Legendre, or any benchmark family.

## Evidence

A fresh current-main profile of `wilkinson56` is dominated by work that is unnecessary for a completely split prime image:

- total: about 33.03 ms;
- prime walk: about 10.17 ms (30.8% of total);
- `berlekampFactor`: about 26.6% of total;
- fixed-space matrix reduction: only about 3.8% of total;
- the Berlekamp kernel has dimension 56 and `Q - I` is the zero matrix.

The selected modular polynomial already consists entirely of distinct linear factors. The expensive part is then equal-degree/root separation, not the matrix representation addressed by #9150.

The codebase already has useful proof ingredients around `primeFieldLinearFactor`, `ZMod64.values`, evaluation/root divisibility, and products of linear factors. The new work should consolidate those ingredients behind one finite-field factorization boundary rather than add a second unrelated public factorizer.

## Measurement gate

Before implementing the production path:

1. Benchmark root enumeration plus repeated/batched linear extraction against the current Berlekamp basis-and-gcd path.
2. Include:
   - current Wilkinson rows over their selected primes;
   - synthetic squarefree fully split inputs across degree and field-size ranges;
   - mixed-degree and irreducible controls;
   - cases where the field is too large for root scanning to be sensible.
3. Separately record detection cost, root enumeration, extraction/product construction, and total factorization time.
4. Determine a deterministic crossover using only degree, field size, and already-computed factorization data.

A cheap identity such as `X^p mod f = X` may certify complete splitting for squarefree `f`, but its cost must be included; do not assume certification is free. Reuse information already produced by the prime planner when possible.

## Proof and design requirements

Provide proved statements connecting:

- field enumeration and `eval f a = 0`;
- each root to divisibility by the corresponding monic linear factor;
- squarefreeness/distinct roots to multiplicity one;
- complete splitting to reconstruction of the original monic polynomial; and
- the fast-path result to the existing public factorization contract.

There should be one auditable selection point and one public factorization result. Avoid duplicated recursion, unchecked Boolean assumptions, benchmark-family tests, or a legacy wrapper around two independent implementations.

## Acceptance

Proceed only if the measurement gate predicts a useful general region. A production implementation should then:

- reduce the finite-field factorization portion by at least 50% and total `wilkinson48`/`wilkinson56` time by at least 15%;
- show transfer on independent fully split inputs;
- add no more than 2% overhead outside the selected region (or make the gate avoid it);
- preserve conformance and solved coverage; and
- run a fresh all-systems performance sweep and regenerate every cactus plot in the same PR.

If root extraction is not competitive after including certification and reconstruction, publish the crossover measurements and close this as a measured no-go.

## Dependencies

- No mathematical dependency on #9150, but both change Berlekamp factorization. Perform the measurement in parallel if useful; rebase the production implementation after #9150 so the result has one clean matrix/factorization boundary.
- The prime-scanning policy in #9156 may change which field is selected; evaluate the fast path with both the current and candidate selected primes.
- Resolve before final validation of #9126.

kim-em:
This PR adds a proved root-extraction branch to `Hex.Berlekamp.berlekampFactor` for finite-field images that split completely into distinct linear factors, cutting `wilkinson_56` end-to-end factorization time by 30.3% and its modular factorization by 95.3%.

A squarefree `f` over `F_p` is a product of distinct monic linear factors exactly when it has `deg f` roots in `F_p`, so enumerating the roots and emitting `X - r` for each is already a complete factorization: no Berlekamp basis, no equal-degree separation, no gcd recursion. The residue scan is its own certificate — the result is used only when the scan finds `deg f` roots, and that one length test is what `Hex.Berlekamp.eq_foldl_rootsIn_of_length` turns into the reconstruction `∏ (X - rᵢ) = f`. No separate complete-splitting predicate is computed and no Boolean is trusted without a check.

`Hex.Berlekamp.rootScanBudget` is the single selection point, reading the degree and the field size and nothing else. `deg f ≤ p` is necessary — `F_p` has `p` elements, so a scan of a higher-degree input can never succeed — and `25 · p ≤ (deg f)^2` keeps an admitted-but-rejected scan at about a twenty-fifth of the matrix build it might have replaced. Together they select `5 √p ≤ deg f ≤ p`. Nothing recognizes an input family.

`HexBerlekamp/LinearFactors.lean` consolidates the monic-linear-factor lemmas rather than duplicating them: `primeFieldLinearFactor` and its degree, monicity, coprimality, and two `foldl` invariants move down out of `RabinSoundness/`, where they were private, so the `X^p - X` product identity behind Rabin's test and the new branch share one set. `berlekampFactor` keeps one public contract; every theorem about its output — product, non-emptiness, positive degree, `Nodup`, per-factor unsplittability, divisibility — is proved for both branches and stated only about `berlekampFactor`.

The measurement gate is `bench/HexBench/RootSplitProbe.lean` and the analysis is `reports/berlekamp-root-extraction-measurement.md`. It priced and rejected the `X^p mod f` certificate: it is not cheaper than the scan it would guard (295 µs against 237 µs at `d = 64`, `p = 67`) and the fixed-space matrix recomputes it afterwards.

Measured on `chungus2` against a baseline built at `f87a0c7d`, since the committed profile predates three merged perf commits and comparing against it would over-credit this change:

| row | total before | after | delta | `berlekampFactor` before | after | delta |
| --- | --- | --- | --- | --- | --- | --- |
| `wilkinson_40` | 12.78 ms | 9.03 ms | −29.3% | 1.952 ms | 0.108 ms | −94.5% |
| `wilkinson_48` | 21.95 ms | 16.12 ms | −26.6% | 3.100 ms | 0.177 ms | −94.3% |
| `wilkinson_56` | 29.75 ms | 20.74 ms | −30.3% | 4.800 ms | 0.226 ms | −95.3% |

Every other profiled row reports `rootExtraction: false`, runs no scan, and moves between −2.5% and +2.3%, the driver's noise band on those inputs. The corpus sweep solves the same 377 of 392 instances for 2.0% less total time; `wilkinson_28` and `wilkinson_32` improve 10.2% and 11.6%, and the profile's validation pass found no factor-degree, leaf-count, selected-prime, or subset-cardinality disagreement across 368 instances. All 25 cactus and runtime-vs-degree figures are regenerated from the fresh hex sweep, with the external comparator curves carried over from the committed `aa68c920` record: `python-flint`, PARI, NTL, and Isabelle are not installed on this host, so the sweep is hex-only.

Closes #9157

🤖 Prepared with Claude Code


kim-em:
Second opinion (Codex, gpt-5.6) found no soundness defect and confirmed the degenerate-input, Nat-arithmetic, non-squarefree, theorem-preservation, lemma-move, and factor-ordering analyses. Three findings, all addressed in c669720a:

- **Medium, memory.** `rootsIn` filtered `ZMod64.values p`, allocating one residue per field element before discarding all but the roots; `rootScanBudget` bounds the scan only against the kernel computation it replaces, not in the absolute, so a large admissible modulus could allocate a list far bigger than the answer. Replaced with the tail-recursive `rootsBelow` loop, which counts down over residue indices and conses only the roots it keeps, so the scan holds `deg f` residues. `rootsIn_eq_filter` proves the loop is the filter.
- **Low, docstring.** "Both branches return irreducible monic factors" overstated the kernel branch's contract, whose leaves are raw gcd outputs monic only up to a unit, with irreducibility proved separately under square-freeness. Reworded.
- **Low, benchmark.** The modular sub-phase profile re-evaluates `rootFactors?` to record the branch tag, repeating the scan. That is outside the measured span and cannot perturb a mark already taken; noted in a comment rather than restructured, since the tag is not otherwise recoverable from the factor list.

Measurements were re-recorded from a clean tree after the scan rewrite, and all 25 figures regenerated. The rewrite improved the numbers slightly: `wilkinson_56` total −29.9% and its `berlekampFactor` −95.5%.

One measurement caveat worth stating plainly: rows outside the selected region move between −2.4% and +2.6% in the phase profile, but two runs of *identical* code differ by up to 3.1% on those same rows, so that spread is the driver's noise rather than a cost of this change. Every one of them reports `rootExtraction: false`, i.e. no scan runs on it at all.

Issue #9170, opened 2026-08-04
Integrate the quadratic-norm certificate behind a recombination budget


kim-em:
Follows #9133; depends on the certificate's irreducibility theorem and on the
executable norm with its correspondence. This is the issue that makes the
measured win real, and the one that owes #9126 a fresh sweep.

## Current state

The certificate exists as a prototype outside production and its meaning is
proved. `Hex.ZPoly.factorize` does not know about it, so `sd5` still costs
69.651 ms and `sd6` still costs 8.058 s -- against 72.6 us and 288.8 us for the
certificate, measured paired in-process in
`reports/hexbz-quadratic-norm-certificate.md`.

## Deliverables

- The certificate on the ordinary singleton-irreducibility path, returning the
  same `Factorization` as every other singleton proof. No second entry point, no
  dispatch layer, no benchmark-name or family-specific branch: #9133's
  acceptance criteria are explicit about all three.

- A search budget deciding when to attempt it. The recommended placement is
  after the modular factorization, where the support width `w` is known and the
  `2^(w-1)` traversal size can be compared against the budget the recombination
  already carries. Rows that recombine cheaply then pay exactly nothing, which
  is what the miss table asks for: median miss cost is 0.14 us but
  `cyclo_phi128` would pay 45 us on a millisecond factorization, 4%, if the
  attempt were unconditional.

- Failure falling through to the one existing path, with no state carried.

- The normalization contract, stated where the check is: every `F(c; d)` is
  monic, so the certificate applies exactly when the primitive part has leading
  coefficient `±1`, and the only normalization is negation. No scaling, no
  content division.

## Verification

- Full `lake build`, `lake build HexConformance`, `lake build HexReleaseTests`,
  the factorization oracles (`bash scripts/ci/run_oracles.sh`), and the emitted
  fixtures reproduced byte for byte.
- No `axiom`, `sorry`, or `native_decide`; `python3 scripts/release/check_trust_surface.py`.
- A fresh full Hex sweep at the integrated revision and regenerated cactus and
  runtime-by-degree plots, including `reports/figures/hexbz-cactus-combined.svg`,
  reusing the unchanged external comparator records. Every regression above 5%
  reported and explained.
- The paired `--entry quadraticNormProbe` numbers re-measured against the
  integrated cascade, so the report's ratios describe production rather than a
  prototype.
- The measurements #9133 asks for after implementation: certificate construction
  and checking separately, coefficient construction and equality separately,
[...]

Issue #9169, opened 2026-08-04
Primitive element and minimal polynomial of a multiquadratic sum


kim-em:
Follows #9133; depends on the multiquadratic degree theorem. The paper proof is
`reports/hexbz-quadratic-norm-certificate.md`, section "The tower theorem",
parts (2) and (3).

## Current state

With the degree theorem in hand, `K = ℚ(√d₁, …, √dₙ)` has degree `2ⁿ` over `ℚ`
whenever the radicands are independent square classes. That is not yet
irreducibility of anything: it says the field is big, not that
`α = c + ∑ᵢ √dᵢ` generates it, and the certificate's whole content is that the
iterated quadratic norm is `α`'s minimal polynomial.

## Deliverables

- `K/ℚ` is Galois with `Gal(K/ℚ) ≅ (ℤ/2)ⁿ`, from `K` being the splitting field
  of `∏ᵢ (X² - dᵢ)` in characteristic zero. The concrete form needed downstream
  is that `σ ↦ (εᵢ(σ))`, where `σ(√dᵢ) = εᵢ(σ)·√dᵢ`, is a bijection onto
  `{±1}ⁿ` -- injective because the `√dᵢ` generate, hence bijective by
  cardinality against the degree theorem.

- The stabilizer of `α` is trivial:

  ```lean
  theorem stabilizer_eq_bot (h : Independent ds) (c : ℤ) : ∀ σ, σ α = α → σ = 1
  ```

  If `σ` flips exactly the nonempty set `S`, then `∑_{i ∈ S} √dᵢ = 0`, so for any
  `j ∈ S` we get `√d_j ∈ ℚ(√dᵢ : i ≠ j)`, contradicting the degree theorem
  applied to the radicands reordered with `j` last. This is the step that needs
  `Independent` to be usable at an arbitrary index.

- `ℚ⟮α⟯ = K`, by the Galois correspondence.

- The orbit is the sign-pattern family and has exactly `2ⁿ` distinct elements,
  by the same argument, so

  ```lean
  theorem minpoly_eq_prod_signs (h : Independent ds) (c : ℤ) :
      minpoly ℚ α = ∏ ε : Fin ds.length → Bool, (X - C (c + ∑ ± √dᵢ))
  ```

- The consequence the certificate needs, for a monic integer polynomial:

  ```lean
  theorem irreducible_of_independent (h : Independent ds) (c : ℤ) :
      Irreducible (F c ds)
  ```

[...]

Issue #9168, opened 2026-08-04
Executable iterated quadratic norm and its polynomial correspondence


kim-em:
Follows #9133. Independent of the two field-theory issues; they can run in
parallel and meet at the integration issue.

## Current state

`bench/HexBench/QuadraticNorm.lean` is the prototype: dense `Array Int`
arithmetic, no proofs, reachable only from `hexbz_factor_service`. Production
needs the same operation on `ZPoly` with a correspondence theorem, and needs the
two `Bool` checks related to the mathematical predicates the field-theory issues
prove things about.

## Deliverables

**Mathlib-free** (`HexBerlekampZassenhaus/QuadraticNorm.lean`, named for the
mathematics per #9133's constraint):

- `Hex.quadNorm (d : Int) (g : ZPoly) : ZPoly`, the norm of `g(X - t)` along
  `ℤ[t]/(t² - d) → ℤ`, computed as the prototype does: one synthetic Taylor
  shift with shift constant `-t` carrying a coefficient pair, then the product
  with the conjugate, whose `t` component cancels coefficientwise by antisymmetry
  so only the rational part is materialized.
- `Hex.iteratedNorm (c : Int) (ds : Array Int) : ZPoly`.
- `Hex.QuadraticNormCertificate` carrying the translation and radicands, and
  `Hex.QuadraticNormCertificate.check : ZPoly → Bool`, the conjunction of the
  independence test and coefficient equality.
- `Hex.independentSquareClasses (ds : Array Int) : Bool`, the `2ⁿ - 1`
  subproduct square tests.

**Mathlib-facing** (`HexBerlekampZassenhausMathlib/QuadraticNorm.lean`):

- The norm correspondence. For a field `K` and `r : K` with `r² = d`,

  ```lean
  theorem map_quadNorm (g : ZPoly) (d : Int) (r : K) (hr : r ^ 2 = (d : K)) :
      (quadNorm d g).toPolynomial.map (algebraMap ℤ K)
        = g.toPolynomial.comp (X - C r) * g.toPolynomial.comp (X + C r)
  ```

- The iterate as the sign-pattern product, by induction on `ds`, landing on the
  `F c ds` the field-theory issues state their theorem about.

- `independentSquareClasses ds = true ↔ Independent ds`, where `Independent` is
  the predicate those issues use. An integer is a rational square exactly when
  it is a perfect square, so this is `Nat.sqrt` correctness plus a subset
[...]

Issue #9167, opened 2026-08-04
Multiquadratic tower degree from independent square classes


kim-em:
Follows #9133, whose go/no-go measurement and full paper proof are in
`reports/hexbz-quadratic-norm-certificate.md`. This is the first of the four
pieces that report's closing section lists, and the largest. It is the Mathlib
side only: nothing here touches the factorization cascade.

## Current state

`bench/HexBench/QuadraticNorm.lean` computes the iterated quadratic norm
`F(c; d₁, …, dₙ) = ∏_{ε ∈ {±1}ⁿ} (X - c - ∑ᵢ εᵢ √dᵢ)` and checks a certificate
`(c; d₁, …, dₙ)` against an input polynomial. A successful check means `f` is
irreducible, and that implication is proved on paper in the report and nowhere
in Lean. It rests on two field-theoretic facts; this issue is the first.

The hypothesis is *independent square classes*: no nonempty subproduct
`d_T = ∏_{i∈T} dᵢ` is a perfect square. It is not optional and it is not implied
by weaker-looking conditions. `12, 20, 45` are pairwise distinct nonsquares with
`20 · 45 = 900`, and `F(0; 12, 20, 45)` factors as two quartics; the report has
the table.

## Deliverables

A Mathlib-facing development of the degree of a multiquadratic tower. Suggested
home `HexBerlekampZassenhausMathlib/SquareClass.lean`, named for the
mathematics per #9133's naming constraint.

- `Hex.SquareClass.Independent (ds : List ℤ) : Prop`, saying that no nonempty
  sublist's product is a square in `ℚ`. State it so that the Bool-valued
  executable check of issue-C can be shown to decide it; do not state it in
  terms of `2 ^ n` subsets of `Fin n` if a `List.Sublist` formulation is easier
  to induct on.

- Lemma A, the whole arithmetic content:

  ```lean
  theorem sq_or_mul_sq_of_isSq_adjoin {F : Type*} [Field F] (hchar : (2 : F) ≠ 0)
      {a : F} (ha : ¬ IsSquare a) {b : F} (hb : IsSquare (algebraMap ... b)) :
      IsSquare b ∨ IsSquare (a * b)
  ```

  from `b = (x + y√a)²` and the `{1, √a}` basis forcing `x * y = 0`. Three lines
  of mathematics; the Lean cost is choosing how `F(√a)` is presented.

[...]

PR #9173, opened 2026-08-04
feat: certify irreducibility through iterated quadratic norms

### LINKED ISSUE #9126: BZ performance: remove the Isabelle crossover in the combined cactus
## Objective

On the current combined 160-row cross-system corpus, Hex leads Isabelle through solved rank 124, trails it from ranks 125 through 140, and retakes the lead when Isabelle encounters its long `cyclo_phi1031` row at rank 141. The temporary crossover is broad mid-tail overhead, not one pathological input.

Representative cumulative times are:

| solved rank | Hex | Isabelle | Hex / Isabelle |
|---:|---:|---:|---:|
| 124 | 0.184 s | 0.184 s | 1.00x |
| 125 | 0.205 s | 0.199 s | 1.03x |
| 130 | 0.458 s | 0.288 s | 1.59x |
| 135 | 0.938 s | 0.487 s | 1.93x |
| 140 | 2.544 s | 1.884 s | 1.35x |
| 141 | 3.075 s | 10.064 s | 0.31x |

The goal is a general improvement to modular planning, recombination, exact divisibility, and the measured arithmetic kernels. Do not add benchmark-name recognition, family-specific dispatch, an unchecked oracle, or a second legacy implementation.

## Success criterion

On a fresh pinned run of the current corpus and protocol:

- Hex cumulative time is at most 85% of Isabelle's at every solved rank from 125 through 140;
- Hex retains its lead and solved count after rank 140;
- every child PR records a fresh full Hex sweep and regenerates all factorization cactus and runtime-by-degree plots, including `reports/figures/hexbz-cactus-combined.svg`;
- every regression above 5% is reported and explained;
- optimized representations have a simple mathematical specification and a proved correspondence theorem;
- the production code remains one well-organized, auditable factorization path.

A simulation using the current paired timings indicates that Isabelle parity on the Swinnerton-Dyer, `x^n - 1`, `cyclo_phi179`, and `cyclo_phi64_x_phi105` rows removes the crossover but leaves only about a 5.5% margin. Adding Wilkinson parity gives at least a 16% lead throughout the elbow, motivating the 15% target above.

## Work items

1. #9127 — record the shared phase-attributed baseline and counterfactual prime-plan data.
2. After #9127, run these in parallel:
   - #9128 — scout modular degree patterns and select primes by bounded downstream cost;
   - #9129 — replace allocation-heavy list traversal with a proved support cursor;
   - #9131 — keep quadratic Hensel lifts canonical modulo the active prime power.
3. #9130 — add a proved finite-field non-divisibility filter after #9129 supplies the compact support representation.
4. #9132 — reprofile after #9128, then implement a packed finite-field matrix only if the specialization gate remains satisfied.
5. #9133 — only after #9129 and #9130, investigate a general iterated quadratic-norm irreducibility certificate if SD remains material.
6. Repeat the pinned full sweep after every production PR and perform the final acceptance comparison here after every required implementation or measured no-go is complete.

```mermaid
graph TD
  B["#9127 attributed baseline"] --> P["#9128 prime degree patterns"]
  B --> C["#9129 support cursor"]
  B --> H["#9131 canonical Hensel residues"]
  P --> M["#9132 finite-field matrices"]
  C --> D["#9130 modular divisibility obstruction"]
  C --> N["#9133 quadratic-norm certificate"]
  D --> N
  P --> A["#9126 final cactus acceptance"]
  H --> A
  D --> A
  M --> A
  N --> A
```

#9132 and #9133 have explicit no-go outcomes. A documented no-go satisfying their measurement gates counts as completion for the final acceptance run; speculative complexity should not land merely to close a checkbox.

## Scope and exclusions

- The present CLD lattice path is not the proposed remedy: measured `sd5` variants take roughly 469--691 ms through CLD and `sd5_x_phi11` about 2.0 s, slower than both the direct classical path and Isabelle.
- Preserve the exact classical fallback and existing correctness theorems.
- Do not use `axiom`, `sorry`, or `native_decide`.
- Name source files and mathematical objects after mathematics, not engineering roles such as `Engine`, `Dispatch`, `Provider`, or `PublicSurface`.
- Each issue body is a hypothesis to validate against current types, theorems, and measurements. If a proposed theorem is false or a required bridge is absent, document the concrete obstruction rather than weakening the claim.

## Existing related work

This investigation follows the original-coordinate and adaptive-planning work in #9104--#9108, closed by #9112. Those issues used Chebyshev/Legendre-centered measurements. The current combined-cactus profile supplies new evidence: the fixed two-extra-prime policy is helpful for `xpow105_minus1` but pure overhead on several other mid-tail rows.

kim-em:
This PR settles issue #9133's go/no-go on certifying irreducibility through
iterated quadratic norms, and the verdict is go. It adds the prototype the gate
is decided against, two service entries and two drivers that measure it, the
durable records, and `reports/hexbz-quadratic-norm-certificate.md` carrying the
theorem, its proof, and the measurements. `Hex.ZPoly.factorize` is untouched;
the Mathlib correspondence that would let a certificate reach production is
decomposed into #9167, #9168, #9169, and #9170.

The class is iterated quadratic norms. For `d : ℤ` and monic `g`,
`N_d(g) = g(X - t) * g(X + t)` in `ℤ[t]/(t² - d)`; iterating from `X - c` gives
the product of `X - c - ∑ᵢ εᵢ √dᵢ` over all `2ⁿ` sign patterns. Swinnerton-Dyer
rows are the case with prime radicands and no translation, but nothing in the
construction asks for that. If no nonempty subproduct of the radicands is a
perfect square -- exactly independence in `ℚ*/(ℚ*)²` -- then the tower has degree
`2ⁿ`, `c + ∑√dᵢ` generates it, and the iterated norm is its minimal polynomial,
hence irreducible. The proof needs no Kummer theory: one elementary lemma about
squares in a quadratic extension plus a subset induction gives the degree, and a
trivial-stabilizer argument gives the primitive element.

So the certificate is the translation and the radicands, checked by `2ⁿ - 1`
integer square tests and one coefficient equality. No number field is
constructed, nothing is factored, no floating point. Recovery is untrusted,
since the check is exact, and has a closed form: `∑_ε exp(t α_ε)` is
`2ⁿ ∏ᵢ cosh(t √dᵢ)`, so taking logs of the power-sum series recovers `∑ᵢ dᵢᵏ`,
and Newton's identities recover the radicands from the top `2n + 1`
coefficients.

Measured on `chungus2`, both sides read off the same in-process observations:
`sd5` 68.853 ms to 83.7 us (826x), `sd6` 7.949 s to 333.8 us (23,816x),
`sd6_shift1` 22.105 s to 644.5 us. Four rows the cascade does not finish inside
the cutoff certify in under 25 ms, `hoeij_S9` at degree 512. Miss overhead over
the 368 declining corpus rows is a median of 0.14 us and a maximum of 86.7 us.

The number that matters for the parent issue: modelled onto the combined cactus
under the conservative post-prime placement, where the row still pays
normalization and the bounded prime walk, the worst Hex/Isabelle cumulative
ratio over ranks 125 through 140 goes from 1.189x to 0.805x, against #9126's
0.85x criterion. Restricting the model to the three `sd5` rows still gives
0.834x, so the conclusion does not rest on the whole certified set. The
certificate alone would close #9126, with the `x^n - 1` and cyclotomic rows
#9126 also named untouched.

That model is `scripts/bench/quadratic_norm_cactus_model.py`, committed rather
than carried by hand: it reads the pinned records, states the replacement rule
in code, measures the retained phases per row, and leaves at their measured time
the six rows whose cascade never finishes -- so the "145 to 151 solved" a
post-normalization integration would give is stated as a projection, not a
modelled result.

The independence hypothesis is load-bearing and is not implied by anything
weaker. `12, 20, 45` are pairwise distinct nonsquares, but `20 * 45 = 900`, and
`Hex.ZPoly.factorize` splits that iterated norm into two quartics; a certificate
checking only "each `dᵢ` is a nonsquare" would accept a reducible polynomial. The
report records that, and `scripts/bench/quadratic_norm_witnesses.py` pins it
with 15 off-corpus cases -- negative radicands, composite radicands, descending
tower order, a translation of a million, degree 1024, and every dependent set --
each checked against an independent implementation and against what the theorem
says the answer must be.

The same driver's randomized pass is the sharper check: for 250 random
translation and radicand tuples it compares the independence test, the
certificate, and whether `Hex.ZPoly.factorize` returns a single factor of full
degree. All 250 agree in both directions. A certified-but-reducible case would
refute the theorem; a dependent-but-irreducible one would show the hypothesis is
stronger than it needs to be. Neither occurs.

Every corpus row that certifies is a Swinnerton-Dyer polynomial, a translate, or
a quadratic, so the corpus alone cannot show the recogniser keys on the
mathematics rather than that shape. The witnesses are what show it.

Includes a fresh hex-factor sweep and the 25 regenerated factorization figures:
the new service entries put `bench/HexBench/FactorService.lean` and
`lakefile.lean` out of date against the committed sweep, which the freshness
checker treats as stale whether or not the measured path moved. 377 of 392
solved, unchanged, and every combined-mixture row within noise of `b5045066`.

Refs #9133, which stays open: its acceptance criteria describe a *successful
implementation*, and the verdict here is go rather than no-go, so the kernel-
checked irreducibility theorem and the production integration are still owed.
Those are #9167, #9168, #9169, and #9170. Had the verdict been a no-go, this PR
would have closed it.

🤖 Prepared with Claude Code


kim-em:
Second opinion from Codex, and what changed.

It confirmed the mathematics and the checker: no counterexample to the tower
theorem, and no reducible polynomial `Certificate.check` accepts. It also
reproduced the cactus arithmetic independently. Five measurement and API
findings, all now addressed:

1. **The "budget-gated" placement was never defined**, and could not have meant
   the existing 262,144-node recombination budget: `sd5`'s 32,768-node walk sits
   under it and would never trip. The model is now
   `scripts/bench/quadratic_norm_cactus_model.py` -- placement named `post-prime`,
   replacement rule in code, retained phases measured per row, rows it cannot
   price left at their measured time, and a record emitted. Building it caught a
   real error in the hand-carried numbers: the phase-profile call is nearly twice
   as slow cold as warm on `sd5`, enough on its own to move the modelled worst
   ratio from 0.805x to about 0.89x, across the target. The script discards a
   warm-up call.

2. **The headline ratios were not paired**: certificate spans came from a
   certificate-only process and production from the paired one. Both sides are
   now read off the same `quadraticNormProbe` observations, reported as the
   median of per-observation ratios. `sd5` reads 826x rather than 960x once the
   comparison is honest.

3. **The `±`-monic normalization was documented but not implemented** --
   `-(X² - 2)` was declined. `unitNormalize` now divides out the unit.

4. **Provenance**: the unused `--cutoff-seconds` is gone, the paired-row sweep is
   validated and recorded, and the witness record carries the same provenance
   block as the others. All three records are now measured with no modified
   source, and `git_dirty` means source dirtiness rather than sibling-record
   churn.

5. **A missing reply-count check** in the witness driver.

Separately, added the check I most wanted after reading the theorem: 250 random
(translation, radicand-tuple) pairs putting the independence test, the
certificate, and `Hex.ZPoly.factorize`'s verdict against each other. All 250
agree in both directions, so on this sample the independence hypothesis is
exactly right rather than merely sufficient.

The verdict is unchanged: worst cumulative ratio over ranks 125--140 falls from
1.189x to 0.805x, and to 0.834x with the model restricted to the `sd5` rows
alone.

kim-em:
Rebased onto #9172, which changes production factorization, so every
measurement on this branch was retaken against the new baseline: the sweep, the
25 figures, the certificate probe, the witnesses, and the cactus model. The
figure conflicts were resolved by taking main's artifacts and regenerating,
since a generated file has no meaningful three-way merge.

The verdict is unchanged and the margins move only in the third digit:

| | before rebase | after |
|---|---:|---:|
| `sd5` speedup | 826x | 861x |
| `sd6` speedup | 23,816x | 24,990x |
| baseline worst ratio, ranks 125--140 | 1.189x | 1.178x |
| modelled worst ratio | 0.805x | 0.802x |
| modelled, `sd5` rows only | 0.834x | 0.829x |

#9172 took the baseline worst ratio from 1.189x to 1.178x on its own, so the
elbow it leaves is still well above #9126's 0.85x target and still has the
Swinnerton-Dyer rows at its worst point.

PR #9174, opened 2026-08-05
feat: derive the multiquadratic tower degree from independent square classes

### LINKED ISSUE #9167: Multiquadratic tower degree from independent square classes

kim-em:
This PR proves that independent square classes force a multiquadratic field to have
degree exactly `2ⁿ`, formalizing part (1) of the tower theorem in
`reports/hexbz-quadratic-norm-certificate.md`. It is the Mathlib side only; nothing
here touches the factorization cascade.

`HexBerlekampZassenhausMathlib/SquareClass.lean` carries:

- `Hex.SquareClass.Independent ds`, saying that no nonempty `List.Sublist` of `ds` has
  a product that is a square in `ℚ`. Sublists select a subset of *positions*, so
  repeated radicands are rejected (`[d, d]` has its own product `d ^ 2`), and the
  formulation inducts directly on the list. `independent_iff_sublists` restates it over
  `ℤ` and over the finite list `ds.sublists`, which is the shape issue-C's Bool-valued
  check decides. `Independent.of_perm` gives the symmetry the later parts need:
  the report's part (2) applies (1) *after reordering the indices*, and
  `not_isSquareIn_of_perm_cons` exposes exactly that consequence -- a radicand is not
  a square in the field generated by the others, at an arbitrary index.
- `isSquare_or_isSquare_mul`, Lemma A, over an arbitrary field `F` with `(2 : F) ≠ 0`
  and `r ^ 2 = a` in an extension `E`: if `b ∈ F` is a square in `F(r)` then `b` or
  `a * b` is a square in `F`. The `{1, r}` representation it rests on is
  `exists_repr_of_mem_adjoin`, obtained by reducing a polynomial expression modulo
  `X ^ 2 - C a`; that avoids constructing a basis and reuses the minimal-polynomial
  computation `minpoly_eq_X_pow_two_sub_C` that the degree count needs anyway.
- `adjoinSqrt ds : IntermediateField ℚ ℂ`, adjoining every complex square root of every
  entry of `ds`. Radicands may be negative, so the tower lives in `ℂ`, not `ℝ`.
  `adjoinSqrt_of_perm` records that it depends only on which radicands appear.
- `not_isSquareIn_prod`, the strengthened claim, stated over a `built ++ rest` split of
  the radicands rather than an index cut `{m+1, …, n}`: no subproduct that uses at least
  one radicand from `rest` is a square in `adjoinSqrt built`. The four cases of the
  induction step are the two ways `built`'s head can enter the subproduct crossed with
  Lemma A's two conclusions; the `d²` case is absorbed by `isSquareIn_sq_mul_iff`.
- `finrank_adjoinSqrt : Module.finrank ℚ (adjoinSqrt ds) = 2 ^ ds.length`.

No Kummer theory is introduced. The only upstream input beyond the intermediate-field
API is `Polynomial.X_pow_sub_C_irreducible_of_prime` at `p = 2`, used to identify
`minpoly F r`. Mathlib has no multiquadratic degree theorem to reuse: there is nothing
under `MultiQuadratic`, `KummerExtension.lean` covers only `X ^ n - C a` irreducibility,
and `QuadraticAlgebra` covers a single quadratic.

`lake build HexBerlekampZassenhausMathlib` and
`lake build HexBerlekampZassenhausMathlib.All` are green, `lake exe runLinter
HexBerlekampZassenhausMathlib.SquareClass` reports "Linting passed" with no findings,
`scripts/release/check_trust_surface.py` passes, and `#print axioms` on the headline
theorems reports only `propext, Classical.choice, Quot.sound`.

One naming deviation from the issue: the def is `adjoinSqrt` and the headline theorem
`finrank_adjoinSqrt`, rather than a bare `K` with `finrank_adjoin_sqrt`, so the theorem
name tracks the definition it is about.

Closes #9167

🤖 Prepared with Claude Code


kim-em:
Second-opinion review (Codex) found no mathematical blocker: it verified the four-case induction step against the paper's `Claim(m)`, confirmed the `built ++ rest` list split faithfully captures the index cut `{m+1, …, n}` (including duplicate radicands, since `List.Sublist` selects positions), and confirmed `not_isSquareIn_of_perm_cons` is exactly strong enough for part (2) -- neither an arbitrary-submultiset theorem nor linear independence of all `2ⁿ` square-root products is needed.

Acted on its findings in 3bae595c:

- **Dependency deviation.** The directive asked for no Kummer theory, and the file imported `Mathlib.FieldTheory.KummerPolynomial` for `X_pow_sub_C_irreducible_of_prime`. `X ^ 2 - C a` irreducibility now comes from `Polynomial.irreducible_of_degree_le_three_of_not_isRoot`, so that import is gone and the only polynomial input is "a quadratic without a root is irreducible".
- **Executable check.** `independent_iff_sublists` is a specification boundary, not by itself a decision procedure, so `decidableIndependent` now closes that gap using Mathlib's `DecidablePred (IsSquare : ℤ → Prop)`, and the `12, 20, 45` counterexample is discharged by exercising it. Worth recording for the follow-up: plain `decide` sticks, because the perfect-square test runs `Nat.sqrt` whose well-founded recursion the elaborator's `whnf` will not unfold; `decide +kernel` does reduce it. What the `Array Int` bitmask checker still owes is the enumeration correspondence and the array/list bridge.
- **Consumer-shaped corollaries** for part (2): `sqrt_mem_adjoinSqrt`, `sqrt_notMem_adjoinSqrt` (a chosen root of a radicand is not in the field generated by the others, after permuting it to the front), and `adjoinSqrt_mono`.
- **Docs.** The four-case table is now in `not_isSquareIn_prod`'s docstring, and the subproduct count is corrected to `2 ^ n - 1` (the empty sublist is skipped).

Not taken: the suggestion to rename `isSquareIn_or_isSquareIn_mul` / `not_isSquareIn_of_perm_cons` / `ne_zero_of_not_isSquareIn`. These are Mathlib's descriptive convention, which is the right one for a Mathlib-facing module, and none is in the 5+ qualifier range the repo's naming guidance targets. The new `sqrt_notMem_adjoinSqrt` gives consumers the short-name entry point.

Re-verified after the change: `lake build HexBerlekampZassenhausMathlib` and `.All` green with zero errors, `runLinter` passes with no findings, `check_trust_surface.py` passes, and `#print axioms` on the headline theorems reports only `propext, Classical.choice, Quot.sound`. Rebased onto `main` (4b77cf43); the only divergence was a docs file, so no conflicts.

Issue #9160, opened 2026-08-04
Berlekamp packed kernel: attribute the 2.6x gap against the measured prototype


kim-em:
Follows https://github.com/kim-em/hex-dev/pull/9158 "perf: pack the Berlekamp
fixed-space matrix into UInt32 words", which landed the packed fixed-space
matrix from #9150. The evidence is
[reports/hexbz-packed-fp-matrix-integration.md](https://github.com/kim-em/hex-dev/blob/main/reports/hexbz-packed-fp-matrix-integration.md)
reading `reports/bench-results/hexbz-phase-profile-a5e0ebe4-chungus2.json`.

## Current state

The landed packed kernel is **2.62x slower than the benchmark prototype it was
derived from**, measured in the same process by the same driver. On
`cyclo_phi385` (240x240 at `p = 3`, rank 236, kernel dimension 4):

| | prototype (`packedHalfWordDivision`) | integrated |
|---|---:|---:|
| pack / build | 2.789 ms | (folded into the build) |
| reduce | 15.404 ms | |
| basis readback | 39.038 us | |
| total | 18.232 ms | 47.752 ms |
| small allocations | 97,269 | 661,844 |

The integrated column is the `fixedSpaceKernelVectors` span less its Frobenius
power and column polynomials, which is the same work the prototype's three
stages do. It is *less* work in one respect: the prototype pays a 2.789 ms pack
step reading an already-built `Hex.Matrix`, and the integrated path writes the
words directly, so the like-for-like gap on the reduction is wider than 2.62x.

Both run the same arithmetic on the same matrix and perform the same 17,948 row
additions, so the difference is scaffolding, not algorithm.

**The gap is measured but not attributed.** The integration report names three
candidates by inspection, and inspection is all they are:

1. `Hex.Matrix.rowAdd` reads the source row out with `Hex.Matrix.getRow` before
   writing, because a closure capturing the matrix would hold a second
   reference and defeat the in-place update. That is one `m`-word copy per row
   addition; the prototype indexed one flat `Array` and paid nothing.
2. `Hex.Matrix.modifyEntries` is a `Fin.foldl` invoking a closure per entry,
   against the prototype's straight-line `for` loop over a flat buffer.
3. `Packed.eliminateColumn` folds over `List.finRange n`, and the pivot columns
   accumulate through `List.concat`, which is `O(length)` per pivot.

None of these has a measured share. 661,844 small allocations against the
[...]

PR #9176, opened 2026-08-05
perf: hoist the packed Berlekamp pivot-row read out of the row addition

### LINKED ISSUE #9166: Berlekamp packed kernel: hoist the pivot-row read out of the row addition
Follows https://github.com/kim-em/hex-dev/pull/9165 "perf: attribute the packed
Berlekamp kernel's 2.6x gap against the prototype", whose evidence is
[reports/hexbz-packed-kernel-attribution.md](https://github.com/kim-em/hex-dev/blob/main/reports/hexbz-packed-kernel-attribution.md)
reading `reports/bench-results/hexbz-phase-profile-d01c7146-chungus2.json`.

## Current state

`Hex.Berlekamp.Packed.rowAdd` reads the source row out with
`Hex.Matrix.getRow` before writing, because a closure capturing the matrix would
hold a second reference and defeat the in-place update. That is one `n`-word
copy per row addition, and #9160 measured it at

| instance | integrated reduction | source-row copy | share |
|---|---:|---:|---:|
| `cyclo_phi385` | 44.596 ms | 31.378 ms | **70.4%** |
| `cyclo_phi128_x_phi165` | 20.087 ms | 14.370 ms | **71.5%** |
| `cyclo_phi64_x_phi105` | 3.856 ms | 2.750 ms | 71.3% |
| `randprod_21` (24x24) | 105.736 us | 67.320 us | 63.7% |

as the median of eight within-repeat paired differences against the `mirrored`
control, with the sign the same on all eight repeats of every row above, and
direction-independent (31.905 ms on the forward calls of `cyclo_phi385` against
31.041 ms on the reverse ones). Every other candidate the integration report and
#9160 named is under half a percent on those rows.

The pivot row is invariant across its column's elimination: `eliminateColumn`
skips `j = pivotRow`, so nothing it writes can change what the next row addition
reads. Reading it once per pivot column instead of once per row addition removes
`n - 1` of every `n` copies. `HexBench.BerlekampKernel.eliminateHoisted` and
`rowAddFrom` are the shape, and they agree with `Hex.Matrix.nullspace` on all 24
corpus rows. With that change the packed reduction is **0.82x to 0.83x the
prototype's** on the two heaviest rows, that is, faster than the code the 2.6x
was measured against.

## Deliverables

Hoist the pivot-row read in `Hex.Berlekamp.Packed.eliminateColumn`, threading
the row into `Packed.rowAdd`, with the correspondence proof carried through.
The proof cost is one more invariant in `Packed.rep_eliminateColumn_foldl`: the
hoisted row must be shown equal to the pivot row of the accumulated buffer at
every step of the fold, which follows from the fold never writing row
`pivotRow`.

Do not take the other two findings in the same PR; see below.

## Traps

All five hazards from
[reports/hexbz-packed-kernel-attribution.md](https://github.com/kim-em/hex-dev/blob/main/reports/hexbz-packed-kernel-attribution.md)
apply to any re-measurement. The one specific to this change is **aliasing**:
the hoisted row must be a fresh `Vector` (as `Matrix.getRow` produces), not a
borrow of the buffer, or the buffer stops being uniquely referenced and every
row addition copies the whole matrix.

## Not in scope, and why

Two secondary findings from #9160 are real but should be separate:

- `Packed.nullspaceArray` searches the pivot-column list for every output
  entry, `O(m * rank + kernelDim * m * rank)`. It is under 1% of the whole
  `fixedSpaceKernelVectors` span on the rows where the reduction dominates, and
  a fifth to a third of it on high-nullity rows (`wilkinson_56` 33%,
  `xpow120_minus1` 19%).
- The `List.finRange n` column scan in `eliminateColumn` is 1.0% on
  `cyclo_phi385` and 16% to 24% on `xpow105_minus1`, `xpow120_minus1` and
  `cyclo_phi41`, whose reductions are dominated by already-zero columns rather
  than by arithmetic.

Neither the entry width nor a Barrett reciprocal is in scope. #9160 priced both,
paired, in one run: the entry width is one-directional at +38% to +98% on every
arithmetic row, so the shipped `UInt32` needs no variant; the reciprocal
resolves on two of 24 rows at 1.4% to 2.2% in Barrett's favour and is unresolved
elsewhere. **If this issue lands, re-open the reciprocal**: the multiply's
incremental cost goes from about 5% of the reduction to about 17%, and a
measurement designed for a two-percent effect would be worth running.

## Verification

- `lake build`, `lake build HexConformance`, conformance fixtures unchanged, no
  new `axiom`, `sorry`, `native_decide` or unchecked oracle.
- `python3 scripts/bench/factor_phase_profile.py --no-counterfactual --no-scout
  --kernel-repeats 8` (an **even** count: the ladder and the prototype variants
  alternate direction per call), whose ladder rungs and packed variants must
  still agree with `Hex.Matrix.nullspace`, and whose `integrated` rung should
  land on the `hoistedPivotRow` rung's time.
- A report recording the before and after on the same host, with paired ranges
  stated as #9160's does.

🤖 Prepared with Claude Code

kim-em:
This PR reads the pivot row once per pivot column in
`Hex.Berlekamp.Packed.eliminateColumn` and threads it into the new
`Packed.rowAddFrom`, instead of reading it out of the matrix once per row
addition. The fold skips `j = pivotRow`, so nothing it writes can change what the
next row addition reads, and a row saves `rowAdds - pivots` copies of `n` words.
`Packed.rowAdd` is kept, defined as `rowAddFrom` on `Matrix.getRow A src`, so its
correspondence lemma is a one-line consequence of the new `Rep.rowAddFrom`.

The correspondence proof carries one extra invariant through
`rep_eliminateColumn_foldl`: `∀ k, rsrc[k] = A[pivotRow][k]` on the accumulated
buffer at every step. `elim_step_pivotRow` establishes preservation from the fold
step either doing nothing or writing a row `j ≠ pivotRow`, and the invariant holds
at entry by `rfl`. The hoisted row is `Matrix.getRow`'s `Vector.ofFn`, a fresh
allocation rather than a borrow of the buffer, so the buffer stays uniquely
referenced and every row addition still writes in place. No `axiom`, `sorry`,
`native_decide` or unchecked oracle.

On `chungus2`, in three runs ordered after, before, after, with the before run
taken by reverting only the two measured files to `a5a55cbf`: the packed
reduction on `cyclo_phi385` goes from 45.095 ms to 13.260 ms (0.29x) and the
whole `Hex.ZPoly.factorize` call from 143.566 ms to 115.069 ms (0.80x);
`cyclo_phi128_x_phi165` goes 19.529 ms to 5.739 ms and 0.82x end to end. Against
the benchmark prototype the reduction goes from 2.80x to **0.83x** and from
2.76x to **0.82x** on those two rows, which is what #9160 projected from the
`hoistedPivotRow` ladder rung. The within-run paired `mirrored -
hoistedPivotRow` difference, which #9160 used to attribute the copy, collapses
from 31.606 ms to 16.841 us (0.1% of the rung), so the hoist is in production and
not only in the ladder.

The four rows at or above 1.00x are the four with nothing to save: the three
Wilkinson rows find no pivot column at all, so `eliminateColumn` never executes
and their 1.02x bounds the cross-run noise floor, and `xpow24_minus1` has 11
pivots and 11 row additions, so the hoist replaces 11 reads with 11 reads.

The inner multiply's incremental share rises from about 5% to about 17% of the
reduction at unchanged absolute cost, which is the condition #9166 set for
re-opening the Barrett reciprocal.

Evidence is `reports/hexbz-hoisted-pivot-row.md`, reading three new records under
`reports/bench-results/`; `scripts/bench/packed_ladder_diff.py` generates its
paired and cross-run tables and reproduces #9160's paired medians from the
committed record. All three runs agreed with `Hex.Matrix.nullspace` on 24 of 24
kernel rows on every ladder rung and packed variant, and with the production
`factorTrace` on 368 of 368 wider-sample instances. `lake build` and
`lake build HexConformance` are green and the Berlekamp and BZ conformance
fixtures reproduce byte for byte.

Closes #9166.

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekamp/PackedKernel.lean
+++ b/HexBerlekamp/PackedKernel.lean
@@ -285,14 +285,25 @@ def rowScale (q : UInt32) (A : Matrix UInt32 n m) (i : Fin n) (c : UInt32) :
     Matrix UInt32 n m :=
   A.modifyEntries i.val fun _ x => mulMod q c x
 
+/-- Replace row `dst` of a packed matrix by `row dst + c * rsrc`, with the source
+row supplied by the caller instead of read out of the matrix.
+
+`rsrc` must be a freshly materialized row, as {name}`Hex.Matrix.getRow` produces.
+A borrow of `A`'s own backing buffer would leave that buffer multiply
+referenced, and the {name}`Hex.Matrix.modifyEntries` below would copy the whole
+matrix per row addition instead of writing in place. -/
+@[inline, expose]
+def rowAddFrom (q : UInt32) (A : Matrix UInt32 n m) (rsrc : Vector UInt32 m)
+    (dst : Fin n) (c : UInt32) : Matrix UInt32 n m :=
+  A.modifyEntries dst.val fun k x => addMod q x (mulMod q c rsrc[k])
+
 /-- Replace row `dst` of a packed matrix by `row dst + c * row src`. Mirrors
 {name}`Hex.Matrix.rowAdd`: the source row is read once, then the destination
 row's entries are updated in place. -/
 @[expose]
 def rowAdd (q : UInt32) (A : Matrix UInt32 n m) (src dst : Fin n) (c : UInt32) :
     Matrix UInt32 n m :=
-  let rsrc := Matrix.getRow A src
-  A.modifyEntries dst.val fun k x => addMod q x (mulMod q c rsrc[k])
+  rowAddFrom q A (Matrix.getRow A src) dst c
 
 theorem getElem_rowScale (q : UInt32) (A : Matrix UInt32 n m) (i r : Fin n)
     (c : UInt32) (k : Fin m) :
@@ -302,15 +313,21 @@ theorem getElem_rowScale (q : UInt32) (A : Matrix UInt32 n m) (i r : Fin n)
   · rw [if_pos (congrArg Fin.val h), if_pos h, h]
   · rw [if_neg (fun hv => h (Fin.ext hv)), if_neg h]
 
+theorem getElem_rowAddFrom (q : UInt32) (A : Matrix UInt32 n m)
+    (rsrc : Vector UInt32 m) (dst r : Fin n) (c : UInt32) (k : Fin m) :
+    (rowAddFrom q A rsrc dst c)[r][k] =
+      if r = dst then addMod q A[dst][k] (mulMod q c rsrc[k]) else A[r][k] := by
+  rw [rowAddFrom, Matrix.getElem_modifyEntries]
+  by_cases h : r = dst
+  · rw [if_pos (congrArg Fin.val h), if_pos h, h]
+  · rw [if_neg (fun hv => h (Fin.ext hv)), if_neg h]
+
 theorem getElem_rowAdd (q : UInt32) (A : Matrix UInt32 n m) (src dst r : Fin n)
     (c : UInt32) (k : Fin m) :
     (rowAdd q A src dst c)[r][k] =
       if r = dst then addMod q A[dst][k] (mulMod q c A[src][k]) else A[r][k] := by
-  rw [rowAdd, Matrix.getElem_modifyEntries]
-  by_cases h : r = dst
-  · rw [if_pos (congrArg Fin.val h), if_pos h, h]
-    rw [show (Matrix.getRow A src)[k] = A[src][k] from rfl]
-  · rw [if_neg (fun hv => h (Fin.ext hv)), if_neg h]
+  rw [rowAdd]
+  exact getElem_rowAddFrom q A (Matrix.getRow A src) dst r c k
 
 /-! ## The packed elementary operations interpret to the generic ones -/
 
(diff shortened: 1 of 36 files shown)


kim-em:
Codex second opinion: no soundness, semantic-equivalence, aliasing, attribute or retained-API blocker. It confirmed the new invariant `∀ k, rsrc[k] = A[pivotRow][k]` is exactly strong enough, that `elim_step_pivotRow` is correct in every branch, that both folds still read the elimination coefficient from their accumulated state (only the pivot row is snapshotted), and that the final `rep_eliminateColumn` statement is unchanged. It read the generated C for the aliasing claim: `Array.ofFn` completes and releases its temporary closure over the matrix before the fold begins, so the accumulator stays uniquely referenced.

It raised five report and script accuracy findings, all confirmed and all now fixed in 8ed3c378:

- the `a5a55cbf` run reproduces #9160's paired attribution on the heavy rows (within 1.5%) but not on the sparse ones, where two of the four intervening commits change the factorization path. The report claimed 1% on every arithmetic row.
- 22 of 24 after-run paired ranges straddle zero, not 12. This understated the null-control result.
- `fixedSpaceKernelVectors` is the median of eight one-call observations, whose raw samples are not retained, not a single observation.
- the Wilkinson rows show about 2% of cross-run drift on a 3 to 7 us no-work row; they do not bound a floor for the millisecond rows.
- `packed_ladder_diff.py` now refuses to call a difference paired when the sample arrays differ in length or are odd, instead of letting `zip` truncate.

It also corrected the `eliminateColumn` docstring's copy count: the hoist takes a dense column's `n - 1` source-row copies to one, a saving of `n - 2`.

CI's `check_factor_sweep_freshness.py` is red because touching `HexBerlekamp/PackedKernel.lean` invalidates the committed `48bdfb8e` factorization sweep. A fresh hex-factor sweep at 8ed3c378 and regenerated figures are running now.

PR #9178, opened 2026-08-05
feat: identify the multiquadratic minimal polynomial with the sign-pattern product

### LINKED ISSUE #9169: Primitive element and minimal polynomial of a multiquadratic sum

kim-em:
This PR proves parts (2) and (3) of the tower theorem in
`reports/hexbz-quadratic-norm-certificate.md`, in the new
`HexBerlekampZassenhausMathlib/Multiquadratic.lean` on top of #9174's
`SquareClass.lean`. With independent square classes `d₁, …, dₙ`, any translation
`c` and any choice of complex square roots `rᵢ`, the sum `α = c + ∑ᵢ rᵢ`
generates the tower `K = ℚ(√d₁, …, √dₙ)`, and the product of `X - β` over the
`2ⁿ` sign patterns `β = c + ∑ᵢ ± rᵢ` is `α`'s minimal polynomial, hence
irreducible over `ℚ` and, by Gauss, over `ℤ`.

`K / ℚ` is Galois by list induction: `adjoinSqrt (d :: ds) = adjoinSqrt ds ⊔ ℚ⟮r⟯`
and each `ℚ⟮r⟯` splits `X² - d`, so `normal_sup` applies; separability is
characteristic zero. `signOf_bijective` gives the concrete `Gal(K/ℚ) ≅ (ℤ/2)ⁿ`
the directive asks for.

The `2ⁿ` conjugates are never constructed. The directive routes the minimal
polynomial through surjectivity of `σ ↦ (εᵢ(σ))` and an orbit count, but an
automorphism is only ever read, never built: `σ(rᵢ)² = rᵢ²` forces
`σ(rᵢ) = ± rᵢ`, so each `σ` permutes the sign patterns by exclusive-or with its
own and fixes the product, whose coefficients are therefore rational. That
product is monic of degree `2ⁿ` and vanishes at `α`, and `[ℚ(α) : ℚ] = 2ⁿ` by
the Galois correspondence, which pins it. `signOf_bijective` is proved anyway,
since the directive asks for it, but nothing downstream uses it.

Both the trivial stabilizer and the distinctness of the `2ⁿ` sign sums are the
one arithmetic fact `eq_zero_of_sum_eq_zero`, that no nontrivial integer linear
relation holds among the `rᵢ`. It is stated with arbitrary integer coefficients
so that it covers both, and it is the only place independence is used at an
arbitrary index, through `List.eraseIdx` and `sqrt_notMem_adjoinSqrt`.

The consumable statement takes the polynomial as input:
`irreducible_of_map_eq_signPoly` says that a rational polynomial *equal to* the
sign-pattern product is irreducible, and `irreducible_int_of_map_eq_signPoly`
says the same for a monic integer one. `Irreducible (minpoly ℚ α)` would be
true and useless.

One thing this does not close. #9175 landed the other half while this was in
progress, but writes the same product as `signPatternPoly`, a `List.prod` over a
fold-built list of signed sums, where this file uses `signPoly`, a `Finset.prod`
over `Fin n → Bool`. The `Fin n → Bool` indexing is load-bearing here -- the
reindexing equivalence is the whole argument -- so composing the two halves
needs a lemma identifying the encodings. That is stated in the module docstring
and in the progress file, and belongs with #9170's integration.

Zero `sorry`/`axiom`/`native_decide`; `#print axioms` on the headline theorems
reports only `propext, Classical.choice, Quot.sound`, and
`lake exe runLinter HexBerlekampZassenhausMathlib.Multiquadratic` passes.
Mentions the progress file `progress/20260805T023000Z_issue-9169.md`.

Closes #9169.

🤖 Prepared with Claude Code

### DIFF
--- a/HexBerlekampZassenhausMathlib/All.lean
+++ b/HexBerlekampZassenhausMathlib/All.lean
@@ -51,6 +51,7 @@ public import HexBerlekampZassenhausMathlib.Classical.Factorization
 public import HexBerlekampZassenhausMathlib.FactorIrreducibility
 public import HexBerlekampZassenhausMathlib.SquareClass
 public import HexBerlekampZassenhausMathlib.QuadraticNorm
+public import HexBerlekampZassenhausMathlib.Multiquadratic
 
 public section
 
--- a/HexBerlekampZassenhausMathlib/Multiquadratic.lean
+++ b/HexBerlekampZassenhausMathlib/Multiquadratic.lean
@@ -32,9 +32,10 @@ generates `K`, and the product of `X - β` over the `2ⁿ` sign patterns
 * `Hex.SquareClass.isGalois_adjoinSqrt`: `K / ℚ` is Galois. Normality is a
   compositum of the quadratic splitting fields of the `X² - dᵢ`, and
   separability is characteristic zero.
-* `Hex.SquareClass.signOf_bijective`: `Gal(K/ℚ) ≅ (ℤ/2)ⁿ`, concretely that
-  reading `σ` off as the sign pattern by which it acts on the chosen roots is a
-  bijection onto `Fin n → Bool`.
+* `Hex.SquareClass.signOf_bijective` and `Hex.SquareClass.signOf_mul`:
+  `Gal(K/ℚ) ≅ (ℤ/2)ⁿ`, concretely that reading `σ` off as the sign pattern by
+  which it acts on the chosen roots is a bijection onto `Fin n → Bool`, and that
+  it carries composition to pointwise `==`.
 * `Hex.SquareClass.eq_one_of_gen_fixed`: the stabilizer of `α` is trivial.
 * `Hex.SquareClass.adjoin_signSum_eq`: `ℚ(α) = K`.
 * `Hex.SquareClass.minpoly_map_eq_signPoly`: the minimal polynomial of `α` is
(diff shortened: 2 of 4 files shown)


kim-em:
Second opinion (Codex) reviewed the statements for soundness: vacuous
quantifiers, missing or unsatisfiable hypotheses, accidentally weakened
conclusions, and the degenerate cases (`ds = []`, zero radicands, repeated
radicands, lemmas that do not assume `Independent`). No high- or
medium-severity defect.

It confirmed the minimal-polynomial argument has no surjectivity gap:
`ε ↦ (fun i => ε i == signOf σ i)` is an involution of the *whole* Boolean
cube, so every actual automorphism permutes every factor of `signPoly`
whether or not each pattern is realised, and the degree `2ⁿ` comes
independently from `ℚ(α) = K`.

Its one finding: `signOf_bijective`'s docstring claimed
`Gal(K/ℚ) ≅ (ℤ/2)ⁿ` while the theorem gave only a bijection of sets. Fixed
in b95f843e by adding `signOf_one` and `signOf_mul` — composition goes to
pointwise `==`, which is `ℤ/2` with `true` as identity. The law needs the
roots nonzero (a zero radicand leaves the sign ambiguous), so it carries
`Independent` via the new `getElem_ne_zero` and `root_ne_neg_self`.

PR #9188, opened 2026-08-10
refactor: separate lift support from residual images in proposal search

### LINKED ISSUE #9152: BZ proposal search: separate lift support from residual images
Tracked by #9126. Follows #9130 / PR #9149. The modulus representation prerequisite #9151 was resolved by PR #9184.

## Problem

The proposal traversal currently packages two mathematical objects with different lifetimes in one `SupportMeta basis target`:

- the prepared lift modulus, lifted-factor degrees, and trailing coefficients depend only on `LiftData` and are constant for the whole peel run;
- `TargetImage target` depends on the current exact residual and is constant only while that residual is searched.

`scanDirectSubsetLevel` constructs the combined object for each subset-cardinality level. Consequently the lift-wide data is rebuilt after every exact split and both parts are rebuilt when one residual exhausts a level and proceeds to the next.

The earlier Wilkinson justification was too strong. Recorded Wilkinson traces have `completedLevels = []`: every residual succeeds at cardinality one. Current code therefore already constructs one target image per Wilkinson residual, and that image cannot soundly be reused after exact division changes the polynomial. Splitting the lifetimes will avoid rebuilding the lift-wide data there, but it will not remove the target reductions responsible for most of the obstruction overhead. The old `wilkinson_56` sample attributed 0.62% of total time to `supportMeta` and 1.00% to `ZPoly.modP`; this issue should not promise to recover the old Wilkinson-family regression.

The refactor remains worthwhile as a small simplification, a possible modest saving on searches that enter several cardinality levels, and the clean mathematical boundary needed by #9153.

## Proposed shape

Replace the mixed object with the two objects its dependencies describe:

1. Construct a `LiftSupport basis` once from `LiftData`, containing the prepared lift modulus and the degree and trailing-coefficient arrays with their proofs.
2. Construct one `TargetImage target` for the current exact residual before walking that residual's cardinality schedule.
3. Pass both objects explicitly through the one production subset traversal and reuse them across every level tried against that residual.
4. After an exact split, retain `LiftSupport basis` and construct a new `TargetImage split.quotient` for the quotient.
5. Remove the mixed `SupportMeta` API rather than retaining a parallel compatibility path.

The names should describe the mathematical objects. Avoid cache, engine, provider, context, and other software-architecture terminology.

## Correctness obligations

- `LiftSupport basis` records exactly `basis.p ^ basis.k`, each lifted factor's degree, and each lifted factor's trailing coefficient.
- Every `TargetImage target` remains dependently indexed by, and proved equal to the reduction of, the exact residual against which it is used.
- A target image is never reused after the residual polynomial changes.
- The refactor leaves selected and complementary supports, reconstructed candidates, exact quotients, budgets, counters, completed levels, and decline reasons unchanged.
- Proposal reconstruction and proved classical replay continue to consume the single production result.

No `axiom`, `sorry`, `native_decide`, unchecked oracle, or duplicate production traversal.

## Measurements

Instrument construction counts before and after implementation. Measure Wilkinson 16--56 as the immediate-success control, plus rows whose recorded traces contain completed cardinality levels, including `sd5_x_phi11`, `sd4_x_sd4shift1`, `xpow48/105/120`, cyclotomic products, Legendre controls, and random products. Use the production `factor` entry; if a newer certificate bypasses proposal search, report that fact and use the phase profiler only to measure the traversal itself.

Record per row:

- distinct lift runs and `LiftSupport` constructions;
- distinct residuals and target-image constructions;
- cardinality levels attempted and completed per residual;
- candidates reaching the obstruction and modular rejections;
- lift-support construction, target reduction, recombination, and total time;
- small allocations.

The construction-count expectations are:

- exactly one lift-support construction per `LiftData` used by a peel run;
- exactly one target-image construction per distinct residual that enters proposal search;
- no reduction in Wilkinson target-image constructions, because those residuals attempt only cardinality one;
- a reduction from one target image per attempted level to one per residual only on searches that enter multiple levels.

Use fully interleaved, verified-idle-core before/after runs. Treat timing as validation of the structural change, not as a prerequisite that the small construction-count saving clear the noise floor.

## Acceptance criteria

- The object lifetimes follow their mathematical dependencies, with no target image shared by different residual polynomials.
- Production contains one proposal traversal and no compatibility wrapper retaining the mixed object.
- Instrumented counts satisfy the expectations above and all existing factorization results, traces, budgets, and counters agree.
- Any timing claim is outside the measured noise floor; a result indistinguishable from noise is an acceptable measured no-go.
- No material testcase regresses by more than 5%, and corpus and family medians remain within the measured noise floor or improve.
- Full build, conformance, factorization oracles, trace gates, and proof tests pass.
- The PR records a fresh full Hex sweep and regenerates every factorization cactus and runtime-by-degree figure, reusing unchanged external comparator records.

## Dependencies

#9151 is complete, so the prepared lift-modulus representation is settled. This issue blocks #9153, whose retained-prime information should extend `LiftSupport` rather than another intermediate traversal boundary.

kim-em:
This PR splits `Hex.SupportMeta basis target` into the two objects its
dependencies describe. `Hex.LiftSupport basis` carries the prepared lift
modulus and the lifted-factor degree and trailing-coefficient arrays with
their proof fields and mentions no target; the existing
`Hex.TargetImage target` carries the residual's image in `F_q[X]`.
`Hex.peelDirect` prepares the lift support once per peel run,
`Hex.peelDirectAux` reduces each residual once, and
`Hex.findDirectSubset` reuses that reduction across the residual's whole
cardinality schedule; an exact split retains the lift support and builds
a fresh image for the quotient, so no image outlives the polynomial its
type names. The mixed structure and `Hex.supportMeta` are removed rather
than kept behind a wrapper, and the head-forced traversal takes the same
two objects.

Over the 23 corpus rows that enter proposal search, 269 lift
preparations become 23 and 269 target reductions become 247; the 369
other rows never call `Hex.peelDirect` and build neither object in either
arm. Every field of every proposal trace is identical on both arms across
all 392 rows, and the regenerated `bz.jsonl` fixtures are byte-identical.

Timing is a measured no-effect, which #9152 pre-registered as an
acceptable result. In a four-block paired corpus sweep the after/before
median is 1.0060x, but so is the median over the rows this change cannot
reach (1.0058x); the gap between proposal-search rows and the rest is
+0.33%, against +0.48% and +0.28% in the two same-binary noise-floor
sweeps. Six rows exceed 1.05x, against three in the same-arm control, and
none costs more than 0.3 ms.

Two findings the issue did not anticipate are recorded in
`reports/hexbz-lift-support-lifetimes.md`. The rows it nominates for
measurement (`sd5_x_phi11`, `sd4_x_sd4shift1`, `xpow48/105/120`) are not
proposal-eligible at all, so their completed cardinality levels belong to
the head-forced classical engine rather than to the traversal this
changes; the rows that do exercise it are the Swinnerton-Dyer sixes and
sevens, the van Hoeij instances, and Wilkinson 24 and up. And the exact
construction counts are derivable from the recorded proposal trace with
no new production counters, which is what
`scripts/bench/proposal_construction_counts.py` does.

The head-forced traversal still prepares both objects per cardinality
level. Hoisting them adds parameters to `Hex.findDirectHead` and
`Hex.searchDirectAux`, which the classical completeness and correctness
proofs quantify over, and removes one preparation per level amortized
over that level's leaves; #9152 scopes the lifetime split to the proposal
traversal, so that is left for its own issue.

Includes a fresh 392-instance hex-factor sweep and every regenerated
cactus and runtime-by-degree figure, reusing the external comparator
record unchanged.

Closes #9152

🤖 Prepared with Claude Code


kim-em:
Second opinion from Codex (via `/second-opinion`): no correctness or soundness blocker. It independently checked the two invariants the issue names and confirmed both hold structurally: a mismatched `TargetImage` would need an equality cast that only typechecks when the targets are equal, and `LiftSupport`'s proof fields pin the modulus and every array entry to the exact basis, with the shrinking residual support staying a list of indices into the original basis (the smaller `residualLift` is built only after peeling finishes).

Four low-severity findings, all addressed in c07f5dfa and ec7ff226:

1. The docstrings claimed more than the code does. `liftSupport basis` is evaluated by `peelDirect` even when the input is already `1`, and a residual's image is built before `findDirectSubset` checks whether any level fits the budget. The docstrings and the report now say "once per nonunit residual reaching `findDirectSubset`" and note the eager placement; a start/worker split for the sake of a case that peels nothing did not seem worth it.
2. The `LiftSupport` docstring said "computed once per lift" while `scanDirectLevel` still computes one per head-forced level, and the report understated the remaining cost as one modulus preparation, when it is a whole lift-support preparation including both full-basis arrays. Both reworded.
3. Trace equality is weaker than output equality. `proposal_construction_counts.py` now records a SHA-256 of the returned factorization; re-running both arms over all 392 rows diffs clean with the digests included. Eight rows return a proposal factorization and their digests match coefficient for coefficient; `hexbz_emit_fixtures` reproducing `bz.jsonl` byte for byte (220 cases, full coefficients) covers the production factorizer more widely.
4. The report claimed every target-row spread straddles 1.0; `hoeij_S8`'s is 1.008x to 1.020x over nine repeats. Corrected, and "no effect" is now "no effect detectable at whole-factorization scale".

The documentation commit touches two `.lean` files, so the published sweep was re-recorded and the figures regenerated. The rebuilt service binary hashes identically (`822e8128...`) to the one every paired measurement used, so those records still price the code as merged.

Codex also declined to request a rename over `LiftSupport` being mildly overloaded with the index-support lists, which matches my read of the call sites.

PR #9196, opened 2026-08-11
interval: validate the arbitrary-function framework with Real.sin

kim-em:
## Summary

- add a Mathlib-free range lattice and separately assembled source, negation, and sine packages
- recognize `sin (-x)` structurally and instantiate `sin x`, `-(sin x)`, plus a retained oddness equality
- run a live policy session through matcher → instantiation → sine propagation → negation propagation → equality transport
- validate the complete authoritative trace and close the original caller target
- prove the ordinary theorem `0 ≤ x → x ≤ 1 → Real.sin (-x) ≤ 0` with actual `Real.sin` semantics

## Why this is stacked

This is the first meaningful arbitrary-function vertical on top of #9195. It deliberately does not extend the rational backend.

## Trust boundary

The runtime trace guard and the ordinary theorem are both real, but remain separate in this experiment. The theorem is assembled from the same package-owned transparent soundness combinators; it is not extracted from opaque runtime computation. A full transparent multi-event emitter is the next gate and is explicitly not claimed here.

## Validation

- `lake build HexInterval.Experiment.SineSign HexIntervalMathlib.Experiment.SineSign HexIntervalMathlib.SineSignConformance`
- factor measurement freshness, copyright, line-count, dependency, and diff checks
- no `native_decide`, axioms, or sorries in the new Lean files

## Design signal

The example forces the architecture to support independent arbitrary propagators and instantiation: the sine rule knows nothing about negation, the negation rule knows nothing about sine, and the matcher/equality layer reconnects their generated expressions to the original `sin (-x)` target.


kim-em:
Opus review follow-up landed in 51922ff: the hand-written oddness edge now uses equality payload 1 (matching the live trace), and the SPEC now states the delivered live replay/separate theorem boundary while leaving transparent emission to the stacked follow-up. Focused 1960-job build is green; fresh exact-head review and CI are running.

PR #9197, opened 2026-08-11
interval: emit a kernel proof from the complete sine trace

kim-em:
## Summary

- extend transparent proof emission from rule facts to expression instantiation and equality transport
- quote the exact live `Real.sin` instance, sine fact, dependent negation fact, equality edge, and transport event as plain data
- replay each event through its package-owned theorem schema with kernel-reducible checks
- compose the emitted dependencies and return the theorem to the caller original expression graph
- prove `0 ≤ x → x ≤ 1 → Real.sin (-x) ≤ 0` from the emitted trace

## Trust boundary

Opaque search is used only to plan and to test that the literals were quoted field-for-field. The theorem does not depend on the runtime `#guard` or opaque session. Every replay success proof is `rfl`; only package soundness lemmas and generic extension/meet/transport/closure combinators enter the proof. There is no `native_decide`, axiom, or sorry.

## Why this is stacked

This is the proof-emission bridge on top of #9196, which supplies the actual `Real.sin` semantics, arbitrary-function packages, and live trace.

## Validation

- `lake build HexInterval.Experiment.ProofEmitter HexInterval.ProofEmitterConformance HexIntervalMathlib.SineProofConformance`
- factor measurement freshness, copyright, line-count, dependency, and diff checks
- banned-proof audit on changed Lean sources

## Still open

The quoted definitions are the output shape of the future tactic frontend; this PR does not yet implement the elaborator that automatically converts an opaque successful search trace into those literal applications. A generic transparent fold is also still available as an alternative, but is no longer required for the direct-emission architecture.


kim-em:
Opus review follow-up landed in 2c652f0: quotation fidelity now compares all 14 instance-event fields, malformed payloads are rejected by both new replay transitions, the SPEC attributes checks to the correct transition, and the lower equality-payload fix is merged. Focused proof-emitter build is green; exact-head verification and CI are running.

PR #9245, opened 2026-08-14
refactor: move the integer row-lattice vocabulary to hex-matrix

kim-em:
This PR moves `Hex.Matrix.memLattice`, its transport lemmas, and the packed certificate checkers `mulEqCert` and `sameLatticeCert` from `hex-lll` into `hex-matrix`, and makes `mulEqCert` public in `Hex.Matrix` rather than `Hex.Internal`.

None of them mentions lattice reduction. `memLattice b v` says that `v` is an integer combination of the rows of `b`; `mulEqCert M A C` decides `M * A = C` by packing rows into single integers in balanced base `2 ^ K` so the product is never formed. They lived in `hex-lll`, which sits at the top of the matrix graph, so anything below it that wanted to say "is an integer combination of the rows" could not.

The evidence that this was already a problem is in the tree: `HexGramSchmidt/Int/Core.lean` defined its own `memLattice`, character for character, with a docstring explaining that it "mirrors the LLL predicate without making `hex-gram-schmidt` depend on the downstream LLL library". That duplicate is deleted here and its two uses now refer to the shared predicate.

What stays in `hex-lll` is what genuinely belongs there: `independent`, which needs the Gram determinants, and the lemmas that an adjacent row swap and a size-reduction row combination each preserve the lattice, which need the row-operation multiplicative laws from `hex-row-reduce` and so cannot sit in `hex-matrix` at all.

`HexLLL/` and `HexMatrix/` are on the factorization sweep's tracked source list, so the move marks the recorded measurements stale. The four affected paths are recorded in `scripts/bench/proof_only_runtime_exemptions.json` with their exact blob transitions: the moved bodies are unchanged, `memLattice` and its transport lemmas never reach code generation, and the certificate checkers serve the external LLL provider dispatch rather than the `ZPoly.factorize` path.

Specified in `SPEC/Libraries/hex-hermite.md`, whose planned certificate checker is the third consumer.

🤖 Prepared with Claude Code


kim-em:
Marking this draft rather than merging. An adversarial review caught a factual error in my own justification, and the resolution is a call I should not make alone.

**The exemption reason I wrote was false.** I claimed nothing on the `ZPoly.factorize` path imports the certificate checkers. It does: `lll` dispatches through `ExternalReducer.certifiedReduction` to `certCheck`, which calls `sameLatticeCert` and `mulEqCert`, whenever an external LLL provider is installed, and Berlekamp-Zassenhaus reduces its proposal lattice through exactly that selector. The reason is now corrected to state what is actually true: every declaration body is byte-identical and the call graph is unchanged; what moves is the module home and two generated symbol names.

**The open question is whether that still warrants an exemption.** The mechanism's stated criterion is "no runtime effect", and a relocation with unchanged bodies arguably meets it. But the existing entries cover proof-only edits, unreachable modules, test modules, and renames of erased instances, not relocated executable code that is conditionally on the benchmarked path. Re-running the factorization sweep is the unimpeachable resolution and needs dedicated hardware. Held pending that decision.

Everything else from the review is addressed: the corrected reason, the Hermite and Smith SPEC snippets retargeted at `Hex.Matrix.mulEqCert`, the stale `memLattice` reference in the hex-gram-schmidt README, and the redundant imports.

One recommendation I did not take: adding a deprecated alias for `Hex.GramSchmidt.Int.memLattice`. There is no `@[deprecated]` anywhere in the Hex libraries, and the project's stated model is that a released repo may advance with breaking API changes its consumers have not adopted. An alias would also keep alive the duplicate name this PR exists to remove. Happy to add one if that reading is wrong.

kim-em:
Closing as a stale draft. `main` has independently moved the same code — `HexLLL/Certificate.lean` and `HexMatrix/Certificate.lean` now collide as a rename/rename, and the branch is roughly 680 commits behind. Re-cutting the hex-matrix vocabulary move against current `main` will be cleaner than reconciling the two moves.

PR #9254, opened 2026-08-14
doc: specify the modular techniques as three libraries

kim-em:
This PR expands the "Modular techniques" entry in `SPEC/future-work.md` into three library SPECs, split along the dependency seams: rational reconstruction has consumers that want no matrices, and a matrix library and a polynomial library should not depend on each other.

`SPEC/Libraries/hex-modular.md` holds symmetric representatives, incremental integer Chinese remaindering for a scalar and for a residue vector sharing a modulus, rational reconstruction (checked signature, uniqueness, completeness, the common-denominator vector form, and the maximal-quotient heuristic), the modulus supply, and the loop combinator that adds moduli until a caller's own check accepts.

`SPEC/Libraries/hex-modular-matrix.md` holds the multi-modular determinant with a Hadamard bound discharged in its companion, the Abbott-Bronstein-Mulders determinant divisor, a two-sided rank certificate, the Dixon p-adic solve, and the rational kernel basis. `reports/hex-bareiss-performance.md` records an adjusted ratio of 0.049x to 0.057x against FLINT's multi-modular determinant at n = 320 to 512, and hex-bareiss classifies that comparator informational because the algorithms differ; implementing the same algorithm makes the comparison like-for-like, so this SPEC classifies it gating with thresholds written down in advance.

`SPEC/Libraries/hex-poly-z-gcd.md` holds Brown's modular gcd for `Z[x]` with cofactors, a modular coprimality witness, exact division, and the route stack from structural reductions through the coprime fast path, the heuristic gcd, and Brown to a subresultant fallback. The tree has no integer polynomial gcd today: `primitiveSquareFreeDecomposition` converts to `DensePoly Rat` and runs the field Euclidean algorithm, which is the baseline the benchmark family measures against.

Four corrections to what the entry said. Primality is not among the checker's obligations, because reduction modulo any `m` is a ring homomorphism and what elimination needs is invertible pivots, which the extended gcd discovers rather than assumes; pairwise coprimality is the property the reconstruction uses, and distinct moduli need not be coprime. A checker that does name a prime pays for it in the kernel, priced by trial division at 0.04 s for 16 bits against 6.2 s for 31 bits, measured on this tree with `decide +kernel` on `Hex.Nat.isPrimeTrial`. The modular rank is a lower bound because a nonvanishing minor modulo `p` is a nonvanishing integer minor, and that submatrix is what the certificate carries. And the Bezout witness for coprime cofactors over `Z[x]` does not exist, since `Z[x]` is not a Bezout domain; the replacement is a modular witness, unconditional in one variable because the content recursion bottoms out in `Int.dvd_gcd`.

The three operations differ in whether they have a checker, which drives the matrix SPEC: the linear solve is checked by one matrix-vector product, the rank has a complete two-sided certificate, and the determinant has neither, so it is a proved algorithm and certified dispatch to an untrusted external implementation is unavailable for it alone.

Each SPEC names its prerequisite relocations. `zmod64FieldOfPrime` is in hex-poly-fp though it is about a `ZMod64` type; Hadamard's inequality is in hex-poly-z-mathlib though it is a determinant inequality; `floorSqrt` and `ceilSqrt` are under the `Hex.ZPoly` namespace; hex-matrix has no entrywise map; and `Hex.Int.extGcd` returns only the final Bezout pair, so the truncated remainder sequence rational reconstruction needs is new.

`hex-mv-gcd` gains a pointer to `hex-poly-z-gcd` for its arity-one case, and `SPEC/Libraries/README.md` gains the six new entries.

https://claude.ai/code/session_0145q6AMz5KFcfS9VUeNCULv

### DIFF
--- a/SPEC/Libraries/README.md
+++ b/SPEC/Libraries/README.md
@@ -83,7 +83,7 @@ Each library with its immediate dependencies:
 - **hex-hermite**: hex-row-reduce, hex-arith, hex-bareiss
 - **hex-smith**: hex-hermite
 - **hex-mod-arith**: hex-arith
-- **hex-modular**: hex-arith, hex-mod-arith
+- **hex-modular**: hex-arith
 - **hex-modular-matrix**: hex-modular, hex-matrix, hex-row-reduce, hex-determinant, hex-mod-arith, hex-arith, hex-basic
 - **hex-gram-schmidt**: hex-row-reduce, hex-determinant, hex-bareiss
 - **hex-lll**: hex-gram-schmidt, hex-matrix, hex-basic
@@ -116,7 +116,7 @@ Mathlib companion libraries (each also depends on Mathlib):
 - **hex-mv-poly-mathlib**: hex-mv-poly, hex-poly-mathlib
 - **hex-mv-gcd-mathlib**: hex-mv-gcd, hex-mv-poly-mathlib, hex-resultant-mathlib, hex-poly-mathlib
 - **hex-poly-z-mathlib**: hex-poly-z, hex-poly-mathlib
-- **hex-poly-z-gcd-mathlib**: hex-poly-z-gcd, hex-poly-z-mathlib, hex-poly-mathlib, hex-mod-arith-mathlib
+- **hex-poly-z-gcd-mathlib**: hex-poly-z-gcd, hex-poly-z-mathlib, hex-poly-mathlib
 - **hex-roots-mathlib**: hex-roots, hex-poly-z-mathlib
 - **hex-real-roots-mathlib**: hex-real-roots, hex-poly-z-mathlib
 - **hex-interval-mathlib**: hex-interval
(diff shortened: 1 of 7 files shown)


kim-em:
Second opinion: Codex reviewed the three drafts adversarially with read access to the tree and returned 22 findings. The second commit fixes all of them; `progress/20260814T004834Z_modular-techniques-specs.md` records what changed and why.

The one worth reading if you read nothing else: the `2^31` cap on a `ZMod64` modulus is not only a performance limit. Writing `L` for the lcm of everything below `2^31`, every allowed modulus divides `L`, and so does every product of pairwise coprime allowed moduli. On the `1x1` matrix `[L]` the determinant loop never reaches its bound, no rank certificate exists because every minor residue is zero, and for `f = x`, `h = x + L` no modular coprimality witness exists. Three separate totality claims, one cause. The determinant now falls back to `Hex.Matrix.bareiss`, the rank certificate carries an arbitrary `Nat` modulus checked in `Int` arithmetic, and the gcd certificate carries a second constructor that names no prime.

Four other design errors: a dependency cycle (milestone 6 proposed rewriting a hex-poly-z function to call a library above it), a `Crt` state with modulus zero under which three stated theorems are jointly unsatisfiable, a common-denominator reconstruction that multiplied denominators where `lcm` is required, and a Dixon numerator bound that named one replaced column where Cramer's rule uses all of them.

Codex tried and failed to break the rational-reconstruction uniqueness proof, the derivation of `gcd(q, m) = 1`, the necessity of Garner's outer reduction, determinant reduction over arbitrary moduli, the two-sided rank argument, Dixon's exact division, the univariate coprimality checker's unconditional soundness, and the Mathlib-free route to maximality. The last got stronger: the Gauss descent step already exists as `ZPoly.dvd_of_toRatPoly_dvd_of_primitive` (`HexPolyZ/Decomposition.lean:797`).

PR #9330, opened 2026-08-21
refactor: move shared exact division into HexBasic

### LINKED ISSUE #9315: refactor: move shared exact division into HexBasic
## Objective

Move the coefficient-independent exact-division contract below both
hex-resultant and hex-bareiss, without changing its semantics.

## Deliverables

- Inspect `HexResultant/ExactDiv.lean` and every user of `ExactDivLaws` and
  `exactDiv` on current `main`.
- Add `HexBasic/ExactDiv.lean` containing the total operation, law package,
  and coefficient-independent cancellation lemmas.
- Leave `DensePoly` operations and polynomial-specific instances with their
  polynomial consumer.
- Preserve public imports and existing behavior, including deterministic
  division by zero.
- Update affected umbrella imports, SPECs, and dependency declarations.

Do not redesign exact division or generalize unrelated polynomial code. Check
the proposed move against the actual import graph before editing; if a named
declaration cannot move without a new dependency cycle, document the concrete
boundary and move the maximal coherent lower layer.

## Verification

- `lake build`
- Relevant resultant and Bareiss conformance targets
- `git diff --check`

## Dependencies

None. This is a root issue in the SPEC-work DAG.

kim-em:
This PR moves the coefficient-independent exact-division contract below both
hex-resultant and hex-bareiss, without changing its semantics. `HexBasic/ExactDiv.lean`
is new and holds the `ExactDivLaws` class with `mul_right_cancel` and `mul_ne_zero`,
the total `exactDiv` wrapper with `exactDiv_zero_right`, `exactDiv_eq_div_of_ne` and
`exactDiv_mul_right`, the nontriviality lemmas `one_ne_zero_of_nonzero` and
`negOne_ne_zero_of_one`, and the `Int` and `Lean.Grind.Field` instances. Every
declaration moves verbatim; division by zero is still the deterministic zero branch.
`HexResultant/ExactDiv.lean` keeps the `DensePoly` scalar-division operations, the
lightweight dense-polynomial ring tower, and the recursive `ExactDivLaws (DensePoly R)`
instance, and re-exports the shared contract through a public import, so no import
outside those two files changes and downstream consumers see an unchanged API.

The maximal coherent lower layer is smaller than the directive's wording suggests, and
the reason is a name-collision boundary rather than a dependency cycle. `HexBasic` sits
below the whole graph, so a declaration `Hex.foo` there is in scope for every
`Hex*Mathlib` file that says `open Hex`; if Mathlib also defines `_root_.foo`, each bare
`foo` in those files becomes an ambiguous-term error. `Hex.mul_pow`, `Hex.pow_mul` and
`Hex.pow_ne_zero` all collide that way, and moving them breaks
`HexGFqMathlib/Primitivity.lean` (`HexHenselMathlib/HenselLemmas.lean` is the other live
site). They stay with `powNat` and `divExp` in `HexResultant/ExactDiv.lean`, which is
the only consumer that needs them: `divExp` is Brown's scalar update, and generic
Bareiss needs only `exactDiv`, `ExactDivLaws` and cancellation. The boundary is recorded
in both module docstrings and in `SPEC/future-work.md`, since it constrains anything
else placed in `HexBasic`.

`libraries.yml` gains `HexBasic` in `HexResultant`'s `deps`. `lake build` is green at
9804 jobs; `check_dag.py`, `check_file_line_counts.py`, `check_copyright_headers.py` and
`git diff --check` pass; the resultant and Bareiss conformance and emit-fixture targets
build clean.

Closes #9315

🤖 Prepared with Claude Code

### DIFF
--- a/HexBasic.lean
+++ b/HexBasic.lean
@@ -8,6 +8,7 @@ module
 
 public import HexBasic.ArrayDecEq
 public import HexBasic.Conditional
+public import HexBasic.ExactDiv
 public import HexBasic.ExtTreeMap
 public import HexBasic.Fold
 public import HexBasic.List
@@ -28,4 +29,9 @@ the `Batteries` list lemmas reproduced in `HexBasic.ListShim`, the
 kernel-reducible `Array`/`Vector` equality (`HexBasic.ArrayDecEq`) and
 `ofFn` (`HexBasic.OfFn`), plus conditional reduction lemmas with names stable
 across supported Lean versions (`HexBasic.Conditional`).
+
+It also holds the shared exact-division contract (`HexBasic.ExactDiv`): the
+total `exactDiv` wrapper with deterministic division by zero, the
+`ExactDivLaws` package, and the coefficient-independent cancellation lemmas
+that fraction-free algorithms in `hex-resultant` and `hex-bareiss` both need.
 -/
--- a/HexBasic/ExactDiv.lean
+++ b/HexBasic/ExactDiv.lean
@@ -1,0 +1,218 @@
+/-
+Copyright (c) 2026 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
+-/
+
+module
+
+public import Std
+public import Init.Grind.Ring.Field
+
+public section
+
+/-!
+Shared exact scalar division.
+
+The executable operations reuse the coefficient type's existing `/` and make
+division by zero deterministic. Algebraic correctness is isolated in
+`ExactDivLaws`; computational consumers need only the operations, while proof
+consumers opt into the law package.
+
+This module is coefficient-independent: it fixes the contract and the
+cancellation lemmas that hold for every carrier, leaving carrier-specific
+operations and instances (dense polynomials, matrices) to their consumers.
+-/
+namespace Hex
+
+universe u
+
+/-- A quotient operation is exact when multiplication by every nonzero right
+factor can be undone by division by that factor. -/
+class ExactDivLaws (R : Type u) [Zero R] [Mul R] [Div R] : Prop where
+  /-- Right multiplication followed by division by a nonzero factor cancels. -/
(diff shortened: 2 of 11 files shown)


kim-em:
Reviewed with Codex while CI ran. It agreed with the boundary call (keep the seven power declarations in hex-resultant rather than moving them and qualifying call sites with `_root_`, on the grounds that the qualification is an ongoing namespace-pollution tax at the lowest library level and Bareiss gains no reuse from the move), and confirmed the move is semantics-preserving: `exactDiv` is verbatim, `@[expose]`/`@[simp, grind =]`/`@[simp]`/`@[csimp]` all survive, every old declaration appears exactly once across the two files, and no declaration docstring changed.

It caught one factual error, since fixed. I had cited `HexHenselMathlib/HenselLemmas.lean` as a second collision site based on a file-level grep, but the hazard is lexical: that file's bare `pow_ne_zero` uses are at lines 391 and 418 while its first `open Hex` is at line 526, so it is unaffected. `HexGFqMathlib/Primitivity.lean:93` is the one live site. The docstrings, the SPEC, and the progress entry now say "wherever `open Hex` is in effect" rather than treating `open` as file-wide.

Also applied: dropped the unused `public import Std` from `HexBasic/ExactDiv.lean` and the now-redundant `public import Init.Grind.Ring.Field` from `HexResultant/ExactDiv.lean` (the new module supplies it publicly); recorded `HexBasic` in the `HexResultant` umbrella docstring and the hex-resultant SPEC title, both of which still said the dependency surface was `HexPoly` alone; and rewrote the `SPEC/future-work.md` entry as current design rather than a record of the refactor.

The first CI run failed on `check_factor_sweep_freshness.py`: `HexBasic/` is one of the check's Hex source prefixes, so adding a module there invalidates the recorded factorization sweep. Added a proof-only exemption, which is what that list is for. The module is a verbatim relocation; the factorization path does reach it through the `HexBasic` umbrella, but no executable factorization definition calls `exactDiv` or mentions `ExactDivLaws`, and both instances were already global via hex-resultant.

PR #9407, opened 2026-08-22
feat(primality): add the primality term elaborator and tactic

kim-em:
This PR adds `HexPrimality/Elab.lean`, completing milestone 3 of HexPrimality/SPEC/hex-primality.md. `primality n` elaborates to a proof of `Hex.Nat.Prime n` following the `irreducibility` architecture: the compiled certificate search runs at elaboration time from the reproducible seed `Rand.ofSeed n` with `defaultPrimeFuel n`, behind a defeq transparency check on the literal (with the "missing `@[expose]`?" diagnostic) and a bit budget; compositeness names the Miller-Rabin witness base and exhaustion reports seed, attempts, and fuel; the search result is self-checked with the same compiled `checkPrime` the kernel will replay before anything is emitted; and the emitted term is `prime_of_checkPrimeAt nE (reified certificate) (Eq.refl true)`, so the kernel replays only the checker, never the search. The reifier is pure constructor applications over `Nat` literals, far simpler than the Rabin-certificate reifier because the certificate carries no proof slots by the subject-readoff design. Tactic forms are `primality`, `primality n`, and `primality h : n`; the companion library will register an additional `@[tactic primalityTac]` handler for `Nat.Prime` goal shapes, as the module docstring records. In-module tests exercise every syntax form across the table, trial, and certificate tiers — `2^31 - 1` kernel-replays a two-level certificate — plus `#guard_msgs` on the failure messages.

Stacked on #9402.

🤖 Prepared with Claude Code


kim-em:
Auto-closed by GitHub when its base branch merged and was deleted; superseded by https://github.com/kim-em/hex-dev/pull/9560 targeting main. Same head branch, no content change.

🤖 Prepared with Claude Code

PR #9470, opened 2026-08-23
ci: check Lean snippets in released READMEs

kim-em:
## Summary

- pin a tested `lean-readme` revision and check every released README source in the existing CI job
- check displayed theorem quotations definitionally against explicit, fully qualified imported declarations with `lean recall Fully.Qualified.name`
- collect imports from checked command blocks before validation, so later examples can stay self-contained while recall fences reuse earlier imports
- keep the recall implementation independent of Mathlib and avoid ambiguous suffix matching
- preserve concrete factorization, irreducibility, and evaluation examples; hidden native-runtime prefixes execute examples backed by precompiled implementations
- correct stale quoted signatures and document the convention

The general recall checker and import handling are proposed independently as https://github.com/david-christiansen/lean-readme/pull/1 and https://github.com/david-christiansen/lean-readme/pull/2. This PR pins a tested private-fork commit that combines them with the explicit-target extension, so it does not wait for either upstream PR.

## Validation

- all 115 README Lean blocks pass `lean-readme`, including 44 explicitly targeted theorem-signature blocks and the concrete Berlekamp–Zassenhaus examples
- all recall blocks omit repeated `import` and `open` setup
- zero `lean nocheck` blocks
- the pinned `lean-readme` revision passes `lake build`, `lake test`, `lake shake`, and its self-check
- `lake build`
- `lake build HexAggregateCheck lean-readme`
- `python3 scripts/release/check_released_manifest.py`
- `git diff --check`


kim-em:
Closing for rework during repository cleanup. This branch now conflicts with `main`, and its latest CI is not merely stale: the README check aborts on the native implementation of `Hex.ZMod64.inv`. The useful `lean-readme` idea should return as a fresh, current, narrowly scoped change after the native-execution boundary is designed and tested; the branch and all 25 commits remain recoverable.

PR #9491, opened 2026-08-23
ci: share only validated Hex caches

kim-em:
This PR makes GitHub's Hex build cache shared and bounded, with R2 as the fallback rather than an unconditional second download path.

- scope keys by runner platform, Lean toolchain, and Lake manifest
- restore on PRs and Pages, but save only after a trusted `main` build passes every gate
- disable `lean-action`'s duplicate whole-`.lake` cache
- call anonymous Lake/R2 restore only when GitHub has no compatible cache
- move both shared-cache publications after the fail-closed verification gate
- document `r2.dev` as an interim, non-production endpoint pending a custom domain
- repair the current `main` freshness failure with an exact blob-pinned exemption for unrelated sparse-poly registrations that landed during the v4.34 measurement run

Local checks:

- workflow YAML parsing
- `uv run --with matplotlib==3.11.1 python scripts/plots/hexbz-cactus.py --check` (25 figures current)
- `python3 -m json.tool scripts/bench/proof_only_runtime_exemptions.json`
- `git diff --check`

The first PR run is expected to miss the intentionally new GitHub cache namespace and exercise the R2 fallback. A successful post-merge `main` run will create the first dependency-scoped GitHub snapshot and publish R2 only after all verification passes.


kim-em:
End-to-end PR CI passed: https://github.com/kim-em/hex-dev/actions/runs/32673684513

Observed cache behavior:

- GitHub correctly missed the new dependency-scoped key (`lake-build-Linux-X64-39faf4...`); no old v4.33 snapshot was downloaded.
- The anonymous R2 fallback ran. It was reachable and non-fatal, but correctly found no v4.34 outputs in 100 ancestors because the first v4.34 `main` run had failed before its upload step.
- The resulting cold build passed all library/bench targets (38m), conformance/fixture targets (6m50s), HexManual (4m23s), benchmark smoke gate (4m32s), and conformance/oracle gate (14m22s).
- The fail-closed gate passed.
- GitHub cache save and R2 publish were both skipped on the PR ref, as intended.

The next required acceptance test is post-merge: a successful `main` run should save the first dependency-scoped GitHub snapshot and publish the v4.34 Lake mapping after all gates pass. A later PR should then restore GitHub and skip R2.

PR #9498, opened 2026-08-24
fix(primality): cap the rho inner iteration budget

kim-em:
This PR caps `rhoInnerFuel` at `2^22` iterations per Brent restart. Uncapped, the budget scaled as `16 · (n^(1/4) + 2)`, so at 256+ bits a single restart was effectively unbounded and a search whose smallest factor was out of rho's reach hung inside its first restart instead of returning `RhoStop.exhausted`. The cap binds only past `n ≈ 2^72` and still covers factors to about `2^44`, beyond rho's documented `~10^12` remit. Measured: a semiprime with two `~2^35` factors still splits in 182 ms; one with two `~2^60` factors now returns a clean `exhausted` after 4 restarts at ~2.3 s per restart, where it previously never returned. `rhoFactor?_spec` and every downstream proof are fuel-agnostic, so no proof changes; the `#guard` regressions use inputs far below the cap threshold and the emitted conformance fixtures are byte-identical to the committed snapshot.

🤖 Prepared with Claude Code


kim-em:
Superseded: the inner-budget cap was cherry-picked into the search PR after the chunk-2 Codex review flagged the same issue, and landed on main via https://github.com/kim-em/hex-dev/pull/9401. Branch kept until the remaining stack merges.

🤖 Prepared with Claude Code

PR #9532, opened 2026-08-24
feat(interval): generalize frontend initial facts

kim-em:
Generalize frontend version-zero facts from source-only arrays to a structurally checked, node-indexed InitialContext with separate positional containment authority.

This adds bounded inputInitialWithin construction, computed-node closure theorems, and preserves the existing source-only wrapper. InitialContext.setFact is transactional and refuses out-of-range or nominally mistargeted writes. Conformance exercises a real zero-event replay from a computed-node fact, structural and semantic anti-permutation, every resource refusal, and guarded trust reports.

Validation: lake build HexIntervalMathlib.FrontendConformance completed across 8,792 targets on exact head; DAG and published trust-surface checks pass. Independent review approved the soundness boundary and its requested replay-coverage hardening is included in the final two-commit head.

### DIFF
--- a/HexIntervalMathlib/Frontend.lean
+++ b/HexIntervalMathlib/Frontend.lean
@@ -295,14 +295,19 @@ domain top. -/
 def Result.initialContext (result : Result) : InitialContext :=
   { rows := result.entries.map fun entry => { node := entry.node } }
 
-/-- Replace one node-indexed selection. Out-of-range writes leave the plain
-context unchanged; `inputInitialWithin` remains the authoritative structural
-validator. -/
+/-- Replace one node-indexed selection only when the requested node is present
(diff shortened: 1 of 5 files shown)


kim-em:
Final-review fixes are pushed at 014968221. The Mathlib SPEC now describes the generalized InitialContext authority and source-wrapper contract; all primary frontend closure axiom reports are enforcing guard_msgs checks. The shared bounded input preflight is factored into inputCheckWithin, so inputWithin preserves its source-error ordering without running the full structural/depth/result scan twice. The SPEC also states the current computed-target version-zero coverage boundary. Exact 8,792-target FrontendConformance build plus trust, DAG, headers, forbidden-term, and diff checks are green. Fresh exact-head CI and re-review are required before merge.

PR #9580, opened 2026-08-25
docs: specify fast polynomial arithmetic

kim-em:
## Summary

- specify a proof-carrying `hex-poly-fast` layer covering multiplication plans, clipped products, Newton division, half-gcd, product/remainder trees, multipoint evaluation/interpolation, and Padé approximation
- incorporate the three high-priority Harvey directions: multipoint Kronecker substitution, redundant-representation NTTs, and cache-friendly truncated transforms
- assign coefficient-specific NTT/CRT machinery and consumer contracts to the existing owning libraries, while keeping dependency directions explicit
- replace the future-work sketch with the detailed SPEC and update the library index and prior-art map

## Validation

- relative-link and Markdown-fence validation across all changed Markdown files
- `git diff --check`

### DIFF
--- a/HexBerlekamp/SPEC/hex-berlekamp.md
+++ b/HexBerlekamp/SPEC/hex-berlekamp.md
@@ -138,6 +138,21 @@ The emitted term contains:
 `hex-berlekamp-mathlib` adds the same syntax for
 `Polynomial (ZMod p)`.
 
+## Fast-arithmetic adoption
+
+After [hex-poly-fast](../../SPEC/Libraries/hex-poly-fast.md), the Fp production
+plan is audited in the repeated Frobenius reductions, distinct-degree gcd
+chain, Rabin powers, and Berlekamp splits. Fast division and half-gcd must
+agree exactly with the existing `DensePoly` operations, so certificate shapes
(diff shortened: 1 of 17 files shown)


kim-em:
Implementation sanity-check found a false required theorem in **Reversal and truncated series**. The displayed signature

```lean
theorem coeff_reverseSeries (f : DensePoly R) (n i : Nat) (hi : i < n) :
    (reverseSeries f n).coeff i = f.coeff (f.size - 1 - i)
```

cannot coexist with the preceding requirement that reversal zero-extend to exactly `n` entries. Take `f = DensePoly.C 1`, `n = 2`, and `i = 1`. Zero-extension requires the left side to be `0`, while Nat subtraction makes `f.size - 1 - i = 0`, so the right side is `f.coeff 0 = 1`.

The theorem needs the advertised range hypothesis `i < f.size`, or its RHS needs an explicit guard:

```lean
if i < f.size then f.coeff (f.size - 1 - i) else 0
```

I have started the lawful multiplication-plan foundation, but I am stopping before the reversal/division layer until this contract is corrected, as required by the repo directive rules.

PR #9602, opened 2026-08-25
Add two-point Kronecker multiplication

kim-em:
Implements the first coefficient-owner kernel from the multipoint Kronecker milestone in `SPEC/Libraries/hex-poly-fast.md`.

- adds forced `Hex.ZPoly.mulKronecker2`, evaluating at `B` and `-B` with exactly two packed `Int` multiplications
- derives the slot bound from `min(size) * maxAbs(p) * maxAbs(q)`
- proves signed digit recovery, even/odd stream separation and interleaving, and `mulKronecker2_eq`
- extends ZPoly conformance with signed, reversed-operand, and digit-boundary KS2 cases
- extends the existing crossover grid with KS2 timing and checksum agreement

The full local crossover sweep found KS2 faster than KS1 in only 2/276 cells on this machine. This PR therefore intentionally exposes KS2 as a forced kernel without putting it in production dispatch; the measured dispatcher remains work for the completed KS2/KS3/KS4 comparison.

Validation:
- `lake build HexPolyZ HexPolyZ.Conformance HexPolyZ.KroneckerCrossover`
- full 276-cell `hexpolyz_kronecker_crossover` sweep with matching checksums
- `python3 scripts/check_phase4.py --base origin/main`
- `git diff --check`


kim-em:
Closing: this kernel is on `main` as `Hex.ZPoly.mulKS2`, landed in 3e41d8fe0 `feat(poly-z): add two-point Kronecker kernel`. That is a separate implementation of the same kernel rather than this branch's `mulKronecker2`, so the two collide in `HexPolyZ/KroneckerMulti.lean` rather than merging. `bench/HexPolyZ/KroneckerCrossover.lean` on `main` sweeps all four of KS1/KS2/KS3/KS4, so the crossover coverage this branch adds is there too.

PR #9605, opened 2026-08-25
Add reciprocal Kronecker multiplication

kim-em:
Implements the SPEC KS3 reciprocal-evaluation kernel in `HexPolyZ`:

- packs the forward and fixed-length reciprocal evaluations
- reconstructs overlapping coefficient slots with a linear carry-state candidate
- certifies the candidate by bounded bidirectional repacking and falls back to a proved recurrence
- proves `mulKronecker3_eq` without axioms or sorries
- extends conformance and the crossover driver

The restored 276-cell sweep found KS3 slower than KS1 in every stable cell (2.20x median, 1.58x best), so this deliberately remains a forced kernel and does not alter production dispatch. The exact runtime-freshness exemption is valid because that dispatcher and the factorization call graph are unchanged.

Depends on #9602. This targets upstream `main` per repository convention; until #9602 merges, the PR includes its KS2 base commit.


kim-em:
Closing: this kernel is on `main` as `Hex.ZPoly.mulKS3`, landed in 23a0957d9 `feat(poly-z): add reciprocal Kronecker kernel`. That is a separate implementation of the same kernel rather than this branch's `mulKronecker3`, so the two collide in `HexPolyZ/KroneckerMulti.lean` rather than merging. `bench/HexPolyZ/KroneckerCrossover.lean` on `main` sweeps all four of KS1/KS2/KS3/KS4, so the crossover coverage this branch adds is there too.

PR #9606, opened 2026-08-25
Add four-point Kronecker multiplication

kim-em:
Implements the SPEC KS4 kernel in `HexPolyZ`:

- performs forward, negated, reciprocal, and negated-reciprocal packed products
- separates even/odd streams and pairs reciprocal streams correctly for both output parities
- reconstructs each stream with the checked reciprocal recovery
- proves `kronecker4_identity` and `mulKronecker4_eq` without axioms or sorries
- extends conformance and the crossover driver with KS4

The restored 276-cell sweep found KS4 slower than KS1 in every stable cell (3.09x median, 1.95x best). KS4 therefore remains a forced kernel and production dispatch is unchanged. The factorization runtime-freshness exemption records that exact fact.

Depends on #9605 (and transitively #9602). This targets upstream `main` per repository convention and includes the stacked base until those PRs merge.


kim-em:
Closing: this kernel is on `main` as `Hex.ZPoly.mulKS4`, landed in ce601dd10 `feat(poly-z): add four-point Kronecker kernel`. That is a separate implementation of the same kernel rather than this branch's `mulKronecker4`, so the two collide in `HexPolyZ/KroneckerMulti.lean` rather than merging. `bench/HexPolyZ/KroneckerCrossover.lean` on `main` sweeps all four of KS1/KS2/KS3/KS4, so the crossover coverage this branch adds is there too.

Issue #9588, opened 2026-08-25
bench-found: hex-hermite HNF coefficient growth blocks Phase 4


kim-em:
## Finding

The Phase-4 audit for `HexHermite` found that the executable HNF sweep does not scale with its declared `O(r · n · m)` matrix-update model on genuine full-rank bounded dense inputs. The corrected deterministic `random-dense-hermite` generator has entries in `[-10, 10]` and FLINT confirms full rank at every scheduled rung.

On `chungus2` (AMD EPYC 9455, Lean 4.34.0-rc2), one-trial exploratory timings were:

| family | n | time/call | time / n^3 |
|---|---:|---:|---:|
| random-dense-hermite | 20 | 1.378 ms | 172.3 |
| random-dense-hermite | 32 | 148.402 ms | 4528.9 |
| random-dense-hermite | 48 | >10 s cap | — |
| rank-deficient-hermite | 20 | 474.230 µs | 59.3 |
| rank-deficient-hermite | 32 | 3.740 ms | 114.2 |
| rank-deficient-hermite | 48 | 36.120 ms | 326.6 |
| rank-deficient-hermite | 64 | 192.294 ms | 733.5 |
| rank-deficient-hermite | 80 | 482.932 ms | 943.2 |

The dense verdict is below resolution after the n=48 cap, and the deficient verdict is inconclusive with a steadily increasing normalized cost. Peak RSS also rises to 75 MB / 105 MB respectively. In contrast, the controlled redundant tall family and the known-form unimodular-conjugate family remain consistent with cubic matrix-update counts, so this is input-dependent integer coefficient growth rather than a harness-wide calibration problem.

The initially drafted dense generator accidentally reduced a polynomial hash directly modulo 21, making rows periodic and capping rank at 21. That fixture produced a false cubic verdict. Adding quotient mixing removes the periodicity; independent python-flint rank checks give full rank for n = 16, 20, 24, 32, 40, 48, 64, and 80.

Per `SPEC/benchmarking.md`, the Phase-4 audit must not fit a new complexity declaration to the implementation or select a benign seed. `HexHermite` must remain at `done_through: 3` until this is resolved.

## Required resolution

- Identify the coefficient-growth mechanism in the eager-reduction/extGcd row sweep.
- Implement a mathematically justified mitigation (for example a bounded-growth or modular/LLL-backed route) without weakening the HNF/certificate API.
- Keep the bounded full-rank dense fixture and the rank-deficient fixture honest.
[...]

PR #9615, opened 2026-08-25
fix(hermite): control HNF coefficient growth

kim-em:
## Summary

- add a rectangular fraction-free rank profile and principal-block HNF schedule
- validate the candidate against the full HNF shape and retain the proved column sweep as a total fallback
- prove transform, inverse, rank, and accumulator agreement for the guarded public path
- extend conformance and coefficient-growth instrumentation
- document Kannan–Bachem, Havas–Majewski–Matthews, and Storjohann–Labahn as future reference points

## Experimental result

On the fixed bounded family, peak/output coefficient sizes remained 7/5, 9/6, 10/6, 11/7, and 11/7 bits for dimensions 20, 32, 48, 64, and 80.

## Validation

- lake build
- lake build HexHermite HexConformance hexhermite_bench after rebasing onto current main (10,177 jobs)
- fixture regeneration diff
- git diff --check

Follow-up to #9588. This PR implements and verifies the growth-control mechanism; it does not by itself claim the dedicated three-trial Phase-4 publication ladder.

### DIFF
--- a/HexHermite/Hermite.lean
+++ b/HexHermite/Hermite.lean
@@ -24,6 +24,17 @@ structure Result (α : Type) (n m : Nat) where
   pivots : List (Fin m)
   accumulator : α
 
+/-- Convert the stored pivot list to the dependent vector expected by
+`RowEchelonData`. -/
+@[expose]
+def Result.pivotVector (s : Result α n m) : Vector (Fin m) s.pivots.length :=
+  ⟨s.pivots.toArray, by simp⟩
+
+@[simp] theorem Result.pivotVector_get (s : Result α n m)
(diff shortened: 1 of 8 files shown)


kim-em:
CI note: the remaining failure is inherited from current main, not introduced by this PR. Both the PR run (32838019312) and main push run (32821635006) report the same stale hex-factor source stamp because lakefile.lean differs from recorded revision 5d48ebfd697a. This PR changes only HexHermite source/SPEC/conformance/bench files; its Phase-4 registration check passes, and the rebased local aggregate build completed all 10,177 jobs.

Issue #9648, opened 2026-08-25
feat: prove multiplicativity of coprime-residue counts


kim-em:
Parent directive: #9619
Prerequisite for the totient correction #9641.

## Current state

The Mathlib-free layer has no finite CRT counting theorem showing that coprime-residue counts multiply across coprime moduli. The SPEC explicitly names this infrastructure as required for totient_eq_count.

## Deliverables

- Construct a Mathlib-free bijection/counting proof for residues modulo m*n versus pairs modulo m and n when Nat.Coprime m n.
- State a short reusable theorem showing the coprime-residue count is multiplicative.
- Do not import Mathlib or weaken totient_eq_count.
- Add focused proof checks for unit modulus and nontrivial coprime composites.

## Context

- HexIntFactor/SPEC/hex-int-factor.md §The divisor-function API and milestone 2
- HexIntFactor/Divisors.lean
- HexArith Mathlib-free CRT/coprimality infrastructure

## Verification

- lake build HexIntFactor
- banned-token and linter checks

depends-on: none

Issue #9641, opened 2026-08-25
fix: compute totient from certified prime powers


kim-em:
Parent directive: #9619
Found by independent review #9622.

## Current state

totient ignores CheckedFactorization and counts coprime residues in List.range n. The SPEC advertises the prime-factor product formula and O(k) arithmetic operations; the current body is O(n) and totient_eq_count is only a definitional restatement. #9647 and #9648 provide the two missing Mathlib-free counting lemmas.

## Deliverables

- Implement totient from checked prime powers using the canonical product p^(e-1) * (p-1), including n = 1.
- Compose #9647 and #9648 to prove totient_eq_count for the new body without weakening the public statement.
- Repair HexIntFactorMathlib.totient_eq so it no longer depends on definitional equality with Mathlibs implementation.
- Add focused conformance for prime powers, mixed composites, and n = 1.

## Context

- HexIntFactor/Divisors.lean:47-50, 97-99
- HexIntFactorMathlib/Factorization.lean:125-134
- HexIntFactor/SPEC/hex-int-factor.md §The divisor-function API and §Complexity
- PLAN/Phase2.md canonical-complexity checks

## Verification

- lake build HexIntFactor HexIntFactorMathlib HexConformance
- benchmark/conformance evidence that totient no longer scans n residues

depends-on: #9647
depends-on: #9648

PR #9685, opened 2026-08-26
feat: count coprime residues modulo a prime power

### LINKED ISSUE #9647: feat: count coprime residues modulo a prime power
Parent directive: #9619
Prerequisite for the totient correction #9641.

## Current state

The Mathlib-free layer has no theorem counting residues coprime to a certified prime power. The SPEC names this as distinct proof work needed before totient can use its O(k) product formula.

## Deliverables

- Prove the count of a in List.range (p^e) with Nat.Coprime a (p^e) equals p^(e-1) * (p-1), under Hex.Nat.Prime p and 0 < e.
- Keep the proof Mathlib-free and expose a short, downstream-usable theorem with a documented statement.
- Add focused proof checks for e = 1 and higher powers.

## Context

- HexIntFactor/SPEC/hex-int-factor.md §The divisor-function API and milestone 2
- HexIntFactor/Divisors.lean
- HexArith Mathlib-free Nat divisibility/coprimality infrastructure

## Verification

- lake build HexIntFactor
- banned-token and linter checks

depends-on: none

kim-em:
Closes #9647

## Summary
- prove that coprimality modulo a positive prime power reduces to the base prime
- count residues in periodic blocks of length `p` to obtain `p^(e-1) * (p-1)`
- expose the theorem in the SPEC and add exponent-one and higher-power proof checks

## Verification
- `lake build HexIntFactor`
- `lake build HexIntFactor HexIntFactorMathlib HexConformance`
- `python3 scripts/release/check_trust_surface.py`

### DIFF
--- a/HexIntFactor/Divisors.lean
+++ b/HexIntFactor/Divisors.lean
@@ -224,6 +224,72 @@ theorem totient_eq_count {n : Nat} (F : CheckedFactorization n) :
     totient F = ((List.range n).filter (fun a => Nat.Coprime a n)).length := by
   simp [totient]
 
+private theorem coprimeCount_prime {p : Nat} (hp : Prime p) :
+    ((List.range p).filter fun a => Nat.Coprime a p).length = p - 1 := by
+  have hp2 := hp.two_le
+  obtain ⟨q, rfl⟩ : ∃ q, p = q + 1 := ⟨p - 1, by omega⟩
+  rw [List.range_succ_eq_map]
+  rw [List.filter_cons_of_neg (by simp; omega), List.filter_map,
+    List.length_map]
+  calc
+    (List.filter ((fun a => decide (Nat.Coprime a (q + 1))) ∘ Nat.succ)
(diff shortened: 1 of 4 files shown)


kim-em:
Fresh Claude Opus second opinion: no mathematical, soundness, edge-case, or Mathlib-free blocker. It independently checked `p = 2`, `e = 1`, truncated subtraction guards, the periodic gcd step, and downstream compatibility with #9641.

Addressed in `109d4d6a9`:
- added compiled `#guard` enumeration checks, so conformance independently evaluates both counts;
- made `coprime_primePow_iff` public and documented it in the SPEC for #9648/#9641 reuse;
- corrected the SPEC API prose and outstanding-CRT explanation;
- made the range split, periodic gcd lemma, and filter congruence explicit.

I retained the compact `progress/20260826T...` filename: current repository progress entries consistently use that ordering format, including all immediately preceding issue work, so the suggested dashed rename would diverge from the active convention. The pre-existing totient complexity mismatch remains owned by #9641.

Issue #9725, opened 2026-08-27
audit-found: runExactLadder's BHKS envelope names a phase that is not the cost


kim-em:
library: HexNumberField

## Symptom

**Declared expectation.** `runExactLadder` declares the classical BHKS
worst-case factorization envelope:

```lean
setup_benchmark runExactLadder n => n ^ 9 + n ^ 7 * (Nat.log2 (n + 2)) ^ 2
```

`PLAN/Phase4.md` §Discipline requires the declared model to be the textbook
complexity rather than the observed one, and its §Exit criteria requires
`run NAME` to return *consistent with declared complexity* for every
parametric registration. `SPEC/benchmarking.md` §"Cover downstream call
patterns" adds that the parameter schedule must exercise the range the model
describes.

**Observed result.** On `chungus2`, Lean 4.34.0-rc2, LeanBench 0.1.0, five
outer trials per rung on an idle host, source commit `066f6fc29`, it returns
`inconclusive` in the faster-than-declared direction. Recomputing `C` against the same
declared function evaluated with a real logarithm rather than the integer
`Nat.log2` step (analysis only; the registrations are unchanged):

`runExactLadder`, input family `(X^n - 2)(X + 3)`:

| n | median per call | C |
|---:|---:|---:|
| 3 | 4.969 ms | 157.9 |
| 4 | 6.507 ms | 17.51 |
| 6 | 89.701 ms | 7.121 |
| 8 | 309.332 ms | 1.966 |

log-log slope **-4.143**; the declared envelope over-predicts by about `n^4`
across four rungs, with three-trial spreads at or below 1.3%.

**Root cause (profiled).** samply at 999 Hz, `n = 8`, 5,334 retained samples,
calibration residual 0.574 ms, confidence and sensitivity passed. Inclusive
cost inside `Hex.*`: `AlgebraicRoot.exact?` 95.63%, `exactFactor?` 95.58%,
`isolate` 95.28%, `Component.refineAll` 86.65%,
`AlgebraicNumber.canonicalRep?` 47.39%.

The Berlekamp-Zassenhaus factorization the declared envelope bounds does not
appear in the ranking at all. The cost is the certification work `exact?` does
*after* factoring: re-isolating candidate factors at separation depth and
canonicalising the selected one.

So the mismatch is not simply "this family makes recombination cheap". The
profile does not measure recombination at any rung, and two rational
irreducible factors would not by themselves establish that modular
recombination is free. The mismatch is that the declared model names a phase
[...]

Issue #9724, opened 2026-08-27
bench-found: QAdjoin.roots? spends 83% of a call re-exactifying roots during duplicate removal


kim-em:
library: HexNumberField

## Symptom

**Declared expectation.** `bench/HexNumberField/Bench.lean` registers

```lean
setup_benchmark runQAdjoinRootsLadder n => n ^ 5 * (Nat.log2 (n + 2)) ^ 2
```

with the adjacent derivation: Yun decomposition is lower order, one shared
double-resultant evaluation eliminant is built and "is not root-isolated", and
"the degree-`d = 2n` norm eliminant's separation-depth isolation remains the
declared ceiling ... (`O(d^3 * B^2)` with `tau, B = O(d log d)`), giving the
same `n^5 log^2 n` wall shape".

**Observed result.** On `chungus2` (Linux x86-64, AMD EPYC 9455, 96 logical
CPUs), Lean 4.34.0-rc2, LeanBench 0.1.0, three outer trials per rung, source
commit `066f6fc29`, input family `g^2 * (X - 1)` over `ℚ(√2)` with `g` dense of
degree `n` and every coefficient `√2`-dependent:

```
Hex.NumberFieldBench.runQAdjoinRootsLadder    expected complexity: n ^ 5 * (Nat.log2 (n + 2)) ^ 2
  param  median per call  spread
      2   126.180 ms       0.91%   (dropped as the leading rung)
      3     1.143 s        1.91%
      4     7.426 s        1.17%
      6    97.730 s        1.57%
      8   593.803 s        0.54%
  verdict: inconclusive (cMin=1175861.253, cMax=2013492.926, beta=—)
```

`fitSlope` rejects the log-log fit because the ladder spans under two octaves,
so the harness falls back to its multiplicative range check: `cMax / cMin` is
1.712 against a bound of `exp(0.35 * log(8/3))` = 1.410.

Recomputing `C` against the same declared function evaluated with a real
logarithm rather than the integer `Nat.log2` step (analysis only; the
registration is unchanged) makes the deviation unambiguous and monotone:

| n | C = t / (n^5 log2(n+2)^2) |
|---:|---:|
| 3 | 8.724e5 |
| 4 | 1.085e6 |
| 6 | 1.396e6 |
| 8 | 1.642e6 |

That is a log-log slope of **+0.641**: the operation is slower than its
declared model by about `n^0.64` over the measured range, in the
slower-than-declared direction, with three-trial spreads at or below 2% at
[...]

Issue #9722, opened 2026-08-26
audit-found: HexNumberField's advertised API surface is not fully registered for Phase 4


kim-em:
library: HexNumberField

## Current problem

Current main has scientific registrations for the principal QAdjoin arithmetic ladders, eliminant construction, lazy addition, exactification and canonical representation, fixed-field roots, algebraic-polynomial roots, duplicate merging, and common-field presentation. It still does not assign every advertised compiled operation to a benchmark registration, as required by PLAN/Phase4.md and the attribution rule.

## Remaining registration matrix

Audit these current public surfaces and add an adjacent justified fixed or parametric registration for each nontrivial operation:

- QAdjoin: approx; subtraction and negation; Div; rational scalar actions; toAlgebraicNumber? and its total wrapper.
- AlgebraicRoot: sub?, mul?, div?, inv?, neg, and their total wrappers.
- AlgebraicNumber: zeroRep; arithmetic; ofRat; Nat/Int powers; scalar actions and numeric casts; isZero.
- Zero/containment decisions: AlgebraicRoot.isZero and RefinedIsolation.containsZero.
- AlgebraicPoly: ofArray, coeff, size, degree?, isZero, and beq.
- Disambiguation.evalMajorant.
- The public Hex.AlgebraicPoly.Common primitives below presentation?: signedShift, rational?, add?, mul?, scale?, shift?, degree, extendShiftStep, extendShift?, extend?, primitive?, powers?, trace?, and coordinates?.
- Separable roots phases Roots.normEliminant and Roots.evalEliminant, so the end-to-end root evidence can attribute its dominant phase rather than infer it.

Before adding a target, confirm the symbol is still public and nontrivial on current main. Group constant-time wrappers into representative fixed registrations when one measurement honestly covers the same implementation route; do not create ceremonial targets or duplicate an existing registration.

## Deliverables

- Add the missing registrations and adjacent cost-model or fixed-budget derivations.
- Update the NumberField SPEC with one current compiled-track assignment covering the advertised API.
- Run any new scientific ladders and refresh the report's target inventory, profiles, and conclusions.
- Leave the report with no unresolved coverage concern.

## Verification

- lake exe hexnumberfield_bench list accounts for every advertised compiled operation.
- lake exe hexnumberfield_bench verify
- python3 scripts/check_phase4.py --all
- NumberField conformance/oracle path
- full CI

## Out of scope

- QAdjoin.inv's remaining model decision, tracked by #9743.
[...]

Issue #9721, opened 2026-08-26
bench-found: QAdjoin.inv's unnormalised rational extended gcd swells its coefficients


kim-em:
library: HexNumberField

## Symptom

**Declared expectation.** `bench/HexNumberField/Bench.lean` registers

```lean
setup_benchmark runQAdjoinInvLadder n => n * n * (Nat.log2 (n + 2) + 1)
```

with the adjacent derivation: "the Euclidean remainder sequence performs a
quadratic number of rational coefficient operations (SPEC §Complexity)
... Intermediate numerator/denominator growth across the chain is modelled with
the same logarithmic limb-growth proxy the HexResultant Brown-chain
registrations use".

**Observed result.** On `chungus2` (Linux x86-64, AMD EPYC 9455, 96 logical
CPUs), Lean 4.34.0-rc2, LeanBench 0.1.0, three outer trials per rung, at source
commit `066f6fc29`:

```
Hex.NumberFieldBench.runQAdjoinInvLadder    expected complexity: n * n * (Nat.log2 (n + 2) + 1)
  param  median per call  spread     C
      4    36.442 us       75.13%   759.215  (dropped as the leading rung)
      6   116.005 us       15.47%   805.591
      8   260.326 us       14.68%  1016.899
     12   817.263 us        6.26%  1418.860
     16     2.381 ms        8.72%  1860.257
     20     4.744 ms        2.60%  2372.143
  verdict: inconclusive (cMin=805.591, cMax=2372.143, beta=+0.887, looks slower than declared by ~n^0.887)
```

`C` rises monotonically by a factor of 2.9 across a factor-3.33 range of `n`,
and the three-trial spread at the rungs carrying that signal is 2.6% to 8.7%,
so this is not host noise. The measured growth is `n^3.07` against a declared
model whose effective exponent over the same range is `2.185`. The row is in
the committed export
`reports/bench-results/hex-number-field-phase4-scientific.json`.

Note that the fixture is height-controlled: numerators cycle modulo 11 and
denominators modulo 6, so the reduced common denominator of the degree-`n`
input stays under six bits at every rung. The growth below is produced by the
chain, not by the inputs.

**Root-cause hypothesis.** `QAdjoin.inv` calls `Hex.DensePoly.xgcdLeft`
(`HexPoly/Euclid/DivGcd.lean`). Its `xgcdLeftAux` recursion never rescales a
remainder:

```lean
let qr := divMod r₀ r₁
[...]

PR #9726, opened 2026-08-27
perf(number-field): measure the Phase-4 ladders and repair the ones that were mis-sized

kim-em:
This PR runs the `HexNumberField` Phase-4 scientific ladders for the first
time, repairs the registrations those measurements showed were mis-sized, and
records the evidence in `reports/hex-number-field-performance.md` with four raw
exports under `reports/bench-results/`. The parametric arithmetic fixture built
coefficient `i` as `±(i + salt + 1) / (i + 2)`, whose reduced common
denominator is `lcm (2, …, n + 1)` and therefore `Θ(n)` bits, so every ladder
over modulus degree silently varied coefficient height as well and no
fixed-height cost model could fit it; numerators now cycle modulo 11 and
denominators modulo 6, holding the common denominator under six bits at every
degree measured. Several ladders returned `inconclusive` only because their
schedules sat below the regime their declared models describe, or because a
wallclock cap truncated the top rung and left the verdict resting on the
`C`-spread fallback rather than a fitted slope; retuned, they fit their
unchanged declared models. The PARI comparator schedules are densified from
three rungs to six per family, because both families cross the ratio 1 inside
the measured range and three points cannot establish that.

Performance rationale per `PLAN/Phase4.md`: no `setup_benchmark` complexity
expression changes in this PR. Every repair is a fixture, schedule or
wallclock change, so each declared model is still the textbook complexity
PR #9424 derived, now measured over a range that can test it.

Five of the nine ladders now fit their declared models —
`runQAdjoinAddLadder` (β = -0.054), `runQAdjoinMulLadder` (-0.009),
`runAddEliminantLadder` (+0.015), `runCommonPresentationLadder` (+0.027) and
`runLazyAddLadder` (-0.123). All twelve Lean/PARI comparator pairs join on
identical result hashes, all 34 fixed registrations agree across repeats, and
the SPEC's 100 ms degree-10 field-operation budget is met with three orders to
spare. All five declared input families are profiled and every profile passes
calibration, confidence and the sensitivity check.

`libraries.yml` is unchanged at `done_through: 3`, because three Phase-4 exit
criteria do not pass. `QAdjoin.inv` returns the slower-than-declared direction
at β = +1.038 on an idle host with 1.8% spread; replaying the shipped
`DensePoly.xgcdLeft` recursion shows its rational entries growing to `Θ(n²)`
bits against the `Θ(n log n)` Hadamard bound a normalised chain attains, and
since `QAdjoin.inv` divides the returned cofactor by the chain's constant gcd,
rescaling each remainder monic is semantics-preserving for this caller. The
PARI comparator corroborates that independently on identical inputs: Hex's
inversion grows as `n^3.02` against PARI's `n^2.14`. Profiling then found two
more mismatches between what the models describe and where the time goes —
`QAdjoin.roots?` spends 83.04% of a call in duplicate removal via `sameValue?`
→ `AlgebraicRoot.exact?` while the phase its model describes is 7.61%, and
`AlgebraicRoot.exact?` spends 95.58% in candidate re-isolation rather than in
the Berlekamp-Zassenhaus factorization its BHKS envelope bounds. Separately,
roughly thirty advertised compiled operations have no registration at all.
Each is filed with its evidence and linked from the report's §Concerns:
#9721, #9722, #9724, #9725.

🤖 Prepared with Claude Code


kim-em:
Second opinion from Codex, and the changes it prompted (commit `19758e0fa`).

It called two blocking problems, both correct:

**1. `prepAlgPolyInput` was still uncontrolled.** I fixed `denseRatCoeff` and missed the one other fixture with the same defect — `prepAlgPolyInput` still built coefficient `i` as `(i+1)/(i+2)`, so `runCommonPresentationLadder` and `runAlgebraicRootsLadder` were still varying height with degree. Fixed; both now share the bounded-height helper. This changed a result: `runAlgebraicRootsLadder`'s continuous-proxy slope goes from `-0.490` to `-1.406`, because the old fixture's growing coefficients were inflating the large-`n` end. Filed as #9728 rather than reported as a pass.

**2. Raising ladder floors removed the downstream-realistic range.** `runAlgebraicRootsLadder` calls `presentation?` at `n = 3..8`, below the floor of 8 that made `runCommonPresentationLadder` fit cleanly. Both floors restored. This costs margin and buys coverage: `runAddEliminantLadder` `+0.015` → `+0.115` (0.15 tolerance), `runCommonPresentationLadder` `+0.027` → `-0.245` (0.35 tolerance). Both still fit their unchanged declared models.

It also judged the arithmetic ceiling drop from 32 to 24 to be the pattern `SPEC/benchmarking.md` §Anti-patterns prohibits. I measured before deciding: the fixture costs 61 s at `n = 28` and over eleven minutes at `n = 32` on an *idle* box, so restoring 32 is not viable and the anti-pattern's remedy — name the problem — applies. Filed as #9727, and both the registration comment and the report now label 24 provisional rather than scientific.

Smaller corrections applied: the `lcm (2,…,n+1)` claim was imprecise (the reduced denominator is `(i+2)/gcd(i+2, salt-1)`, whose lcm is still `Θ(n)` bits); #9724's `O(n²)` exactification count is wrong for this family and is now `O(n)` cross-component comparisons with no caching between them; #9721's `Θ` claims are now stated as observed growth over four rungs; the smooth-log recomputation is relabelled a sensitivity check against a continuous proxy, explicitly *not* the declared function; the PARI corroboration is softened because its mid-rungs contain a discontinuity; and a stale `+0.887` and a stale Concern for #9724 are fixed.

On the central question Codex's verdict was that `denseRatCoeffs`, the addEliminant extension and the cap increases are honest repairs, that period 66 is not degeneracy at the measured maximum of 24, and that not bumping `done_through` "is not conservative — it is mandatory".

Issue #9727, opened 2026-08-27
bench-found: the QAdjoin arithmetic fixture, not the timed operation, caps its ladder at n = 24


kim-em:
library: HexNumberField

## Symptom

**Declared expectation.** `bench/HexNumberField/Bench.lean` registers
`runQAdjoinAddLadder` and `runQAdjoinMulLadder` over `ℚ(2^(1/n))`. The
operations they time are microseconds — 2.6 us and 393 us respectively at
`n = 24`. Nothing about the measured code prevents the ladder from reaching
`n = 32` or beyond.

**Observed result.** The fixture does. `prepFieldInput` and `prepInvInput`
build a certified root of `X^n - 2` through

```lean
private def refinedOf? (p : ZPoly) (h : HasOnlySimpleRoots p) :
    Option (RefinedIsolation p) := do
  let isolations ← isolate p h (separationDepth p : Int)
  let iso ← isolations[0]?
  iso.toRefined?
```

which isolates **all `n` complex roots** at separation depth in order to take
the first one. Measured on an idle `chungus2` (load average 1.4), timing a
whole `_child` invocation at `--target-nanos 1000000` so the figure is
essentially the fixture alone:

| n | fixture cost |
|---:|---:|
| 12 | 1.3 s |
| 16 | 4.9 s |
| 20 | 14.2 s |
| 24 | 30.8 s |
| 28 | 61 s |
| 32 | over 11 minutes (still running when abandoned) |

The rungs up to 28 roughly double per `+4`. The jump at 32 breaks that trend
by more than an order of magnitude, and it is not host contention: the box was
idle and the same shape was observed earlier under load.

**Root-cause hypothesis.** `isolate` floors its target at
`separationDepth p = mahlerPrec p + ceilLog2 (max 2 (deg p)) + sepSlack`, so
the working precision is set by the Mahler bound of `X^n - 2`, and every one
of the `n` roots is refined to it. The bench needs one certified root, and
`QAdjoin` arithmetic never reads the isolation at all — it is a type index.
The cost is entirely in machinery the benchmark does not use.

## Current state

The ladders are registered at `paramCeiling := 24` (arithmetic) and `20`
(inversion). Those are the largest rungs reachable in a sane wallclock, not
the largest rungs the operations support, so **the current range is provisional
rather than a scientific range chosen on its merits**.
`reports/hex-number-field-performance.md` §Bench targets says so and links
here. This does not affect the verdicts already recorded — both arithmetic
[...]

PR #9735, opened 2026-08-27
fix(number-field): normalize inverse polynomial gcd

### LINKED ISSUE #9721: bench-found: QAdjoin.inv's unnormalised rational extended gcd swells its coefficients

kim-em:
## Summary

- add a monic-normalized one-sided polynomial extended gcd while preserving the existing exact-cofactor API
- use it in `QAdjoin.inv` and carry the divisibility/Bezout obligations through the Mathlib field-law proof
- preserve the conformance corpus and add direct normal-xgcd edge/legacy-equivalence regressions
- correct the scientific inversion model to the `n³ log n` worst-case bit-cost bound and record a five-trial degree-96 rerun

At `n = 96`, the normalized chain measures 96.194 ms versus 2.986 s for the shipped chain's idle-host single-root median, a 31.04× speedup. Result hashes match at every rung. The corrected worst-case verdict is inconclusive in the faster direction (`beta = -0.556`), rather than slower than declared; #9743 tracks a tighter aggregate-work model without widening tolerance.

## Verification

- `lake build HexPoly HexNumberField HexNumberFieldMathlib HexPoly.Conformance`
- `lake build HexNumberField.Conformance`
- fresh `hexnumberfield_emit_fixtures` output matches the committed corpus byte-for-byte
- `lake exe hexnumberfield_bench run Hex.NumberFieldBench.runQAdjoinInvLadder --outer-trials 5`
- all 31 Lean-native `hexnumberfield_bench verify` targets pass; the 12 PARI process-call targets require `cypari2`, which is not installed locally and is provisioned by CI

Closes #9721


kim-em:
Independent second opinion (Claude Opus) completed after the initial push.

I incorporated the actionable findings:

- added direct `HexPoly` conformance guards for the normalized representative, one-sided Bezout congruence, zero-left/divisible, and zero-right cases;
- documented the field-only normalization contract and separated it from the future exact-cofactor `xgcdLeftWith` contract;
- corrected the benchmark fixture comment: #9727 currently prevents reaching the asymptotic coefficient-size regime;
- kept the faster-but-inconclusive result as a provisional §Concerns item tracked by #9727;
- recorded the n=4 overhead/crossover, identical result hashes, and qualified the historical PARI comparison.

I did not splice the new Hex medians into the old PARI ratio curve because that would mix unpaired runs; the report now says a paired comparator rerun is required. I also left the tower library's exact chain unchanged because it is a separate caller outside #9721 and may depend on exact cofactors.

The prior CI failure was an upstream-main release-manifest failure: four newly Phase-7 libraries remained in `prepublished_floor`. The same check failed on `origin/main`; this branch now removes precisely those four graduated entries, and both `check_released_manifest.py` and its 36 sync tests pass locally.

kim-em:
The factor-sweep freshness failure is also resolved without changing benchmark provenance: exact blob-to-blob exemptions document that `DivGcd.lean` only adds a separately named operation with no factorization caller, while `Field.lean` adds proofs only. The freshness checker and its 15 tests pass locally. The cactus plot check could not run locally because matplotlib is absent; CI provisions it.

kim-em:
Rebased onto #9734 and reran the normalized inversion ladder with its new single-root degree schedule. The five-trial degree-96 result supersedes the earlier preliminary six-rung measurement: beta = -0.556 against the conservative n³ log n bound, spreads ≤ 2.55%, and 96.194 ms versus the shipped chain’s 2.986 s median at n = 96 (31.04× faster). Result hashes agree at all seven rungs. The remaining faster-than-declared model gap is now tracked in #9743, as requested by the Opus review, without widening tolerance. Post-rebase Lean builds, NumberField conformance, fixture identity, Phase-4 checks, factor freshness, and all 31 local Lean-native benchmark verifiers pass.

Issue #9733, opened 2026-08-27
spec: allow both tight and upper-bound complexity models, with an ordered rule for choosing


kim-em:
Cross-cutting: `SPEC/benchmarking.md` and `PLAN/Phase4.md`. Affects
`HexBerlekampZassenhaus`, `HexPolySmith`, `HexGFqField`, `HexHensel`,
`HexRoots` and `HexNumberField`.

## Current state

lean-bench decides a parametric verdict with a two-sided test —
`if β.abs ≤ slopeTolerance then consistent else inconclusive` — where `β` is the
log-log slope of `C = t / declaredModel`. `PLAN/Phase4.md` §Discipline says to
declare the *textbook* complexity, which is normally an upper bound. So an
implementation that is genuinely faster than its worst-case bound on the
measured family reads `inconclusive` forever, and no amount of correct
engineering changes that. §Exit criteria meanwhile requires every parametric
verdict to be consistent and §Concerns to be empty.

Several libraries have hit this, and — correcting an error in the first version
of this issue — they are **not** all the same case. Counting the word
"inconclusive" in a report does not tell you which direction the residual goes:

| library | `done_through` | what is actually going on |
|---|---:|---|
| `HexBerlekampZassenhaus` | 7 | the genuine faster-than-envelope case: eight ladders `looks faster than declared`, §Concerns "None." |
| `HexPolySmith` | 7 | **slower** than declared (`β = +2.342`, `+1.768`, `+2.331`), and its SPEC says the declarations are algebraic-operation proxies not claiming coefficient-bit bounds. A different problem. |
| `HexGFqField` | 7 | tight mode-1 models; the six inconclusive results are diagnosed as narrow-ladder miscalibration (#2784). Also a different problem. |
| `HexHensel` | 7 | honestly derived mode-1 models with three failing verdicts, carried as Concerns. Mode applies and fails. |
| `HexRoots` | 5 | demoted `runIsolate` to fixed with absolute ceilings (#8750) |
| `HexNumberField` | 3 | reports inconclusive and does not advance |

Only `HexBerlekampZassenhaus` is the case this policy is about. The others are
listed so a future reader does not repeat the mistake of assuming they are.

Scale: 325 parametric registrations repo-wide; `inconclusive` appears in 9 of
~58 reports. This is a tail of hard algorithms, not a pervasive problem.

## The decision

**Both parametric modes are allowed, and the SPEC states how to choose.** The
[...]

PR #9736, opened 2026-08-27
bench(number-field): separate exactification cost evidence

### LINKED ISSUE #9725: audit-found: runExactLadder's BHKS envelope names a phase that is not the cost

kim-em:
Closes #9725

## Summary

- preserve (X^n - 2)(X + 3) as a fixed candidate-selection and canonicalisation benchmark
- add separate exactFactor? and public canonicalRep? parametric phase registrations
- replace the end-to-end exactification ladder with a growing product from the BZ adversarial family, using actual degree and coefficient bit height in the BHKS envelope
- record a confirming profile where factorization is 18.64% inclusive and direct recombination is present, plus refreshed scientific artifacts
- leave the faster-than-upper-bound verdict explicit and linked to policy issue #9733

## Verification

- lake build (10,401 jobs)
- lake build from a cold Mathlib cache (10,407 jobs)
- affected four-target hexnumberfield_bench verify
- python3 scripts/check_phase4.py
- three-trial exactification audit and fixed selection evidence

The full 42-target verifier also passed every Lean target. Its 10 PARI fixed comparisons could not run locally because cypari2 is not installed.


kim-em:
Independent Claude Opus review completed and addressed.

Changes made from the review:
- pin every certification fixture to a root of X^n - 2 by disc intersection, removing isolator-emission-order dependence
- panic on impossible exactFactor? failure instead of timing sentinel values; refreshed evidence now has successful non-sentinel hashes at every rung
- include the full BZ irreducibility factorization in the exactFactor? textbook model
- specify and cross-reference the three exactification input families
- correct profile classification to 98.51% by treating top-level compiled Dyadic helpers consistently with other Lean standard-library code
- distinguish aggregate BZ time from the enclosing-factorization-specific Hensel/recombination frames
- report the actual cMax/cMin mismatches explicitly and keep them open under #9733
- refresh the fixed selection run on the idle host (308.418 ms, five agreeing repeats)
- measure the four affected verify targets together at 1.052 s

One spawn-floor concern did not reproduce as stated: lean-bench compares total_nanos for the amplified child batch, not per_call_nanos. The committed end-to-end rows are 53.36–95.37 ms against a 41.85 ms spawn floor, so they do not time parent-side startup as algorithm work.

PR #9739, opened 2026-08-27
spec: order tight, upper-bound, and fixed benchmark modes

### LINKED ISSUE #9733: spec: allow both tight and upper-bound complexity models, with an ordered rule for choosing

kim-em:
Closes #9733.

## Summary

- add the ordered two-sided, upper-bound, fixed-budget, blocked selection rule to `SPEC/benchmarking.md`, including the no-fitting and worst-case-contract guards
- align Phase 4 discipline, exit criteria, report requirements, release gating, fixed-budget meaning, and anti-patterns with per-mode passing verdicts
- reconcile all six affected reports: HexBerlekampZassenhaus uses mode 2; HexRoots uses modes 1 and 3; HexNumberField and HexPolySmith identify where no mode applies; HexGFqField and HexHensel retain failed mode-1 results rather than reclassifying them
- roll HexPolySmith, HexGFqField, and HexHensel back to `done_through: 3` so the manifest agrees with their unresolved reports; follow-up findings are #9742, #9740, and #9741
- link the upstream harness request kim-em/lean-bench#70 for one-sided verdicts and signed verdict vocabulary

## Performance rationale

The benchmark claim now differs from the per-library worst-case contract only when the ordered rule justifies it: mode 1 declares independently derived expected scaling on the registered family; mode 2 retains a cited published worst-case upper bound but accepts a distinctly rendered faster observation only when profiling confirms that the family exercises the bounded phase; mode 3 retains the SPEC contract while replacing an unreachable scalar wall model with a canonical fixed input and an operation-specific absolute regression budget.

## Verification

- `python3 scripts/status.py HexPolySmith`
- `python3 scripts/status.py HexGFqField`
- `python3 scripts/status.py HexHensel`
- `git diff --check`

### DIFF
--- a/PLAN/Phase4.md
+++ b/PLAN/Phase4.md
@@ -6,11 +6,12 @@
 
 Phase 4 makes algorithmic complexity a first-class deliverable. By
 the end of Phase 4 every advertised compiled operation in the library's API
-has a textbook complexity model declared at its `setup_benchmark`
-registration and a benchmark family whose verdict is *consistent
-with declared complexity*; every advertised proof/tactic operation has the
-fresh-module evidence defined below. An *inconclusive* compiled verdict is not
-a Phase 4 exit; it is a finding that triggers a rollback per
+has the strongest applicable benchmark mode from
+[`SPEC/benchmarking.md` §Choosing the complexity claim](../SPEC/benchmarking.md#choosing-the-complexity-claim)
+and a passing result in that mode; every advertised proof/tactic operation has
+the fresh-module evidence defined below. An *inconclusive* compiled verdict is not
+a Phase 4 exit unless it is the current harness wording for a documented,
+passing one-sided upper-bound result. A failing result triggers a rollback per
 [Conventions.md §Rollback is a normal action](Conventions.md#rollback-is-a-normal-action)
 and a fix at the rolled-back phase.
 
@@ -27,7 +28,7 @@ advertised operation to exactly one row.
 
 | Surface | Required evidence | Generic requirements replaced |
 | --- | --- | --- |
-| Mathlib-free compiled computation | An ordinary LeanBench executable, registrations with controlled one-parameter ladders and adjacent textbook cost derivations, `list`/`verify`, scientific verdicts, comparator coverage, and timed-region sampling profiles. | None. |
+| Mathlib-free compiled computation | An ordinary LeanBench executable, registrations with controlled one-parameter ladders and adjacent independent cost derivations, `list`/`verify`, scientific verdicts, comparator coverage, and timed-region sampling profiles. | None. |
 | Elaboration, proof-search tactics, emitted proof terms, or kernel checking | Externally timed fresh-module builds below an explicit `libraries.yml` `proof_probes` root, with matched import baselines, rotated raw samples, compiler/proof artefacts, and the trust/provenance record in `SPEC/benchmarking.md`. | No LeanBench registration or executable, no `list`/`verify` entry for that surface, no complexity verdict, and no timed-region sampling profile. |
 
 A `mathlib: true` library with a separable compiled core is a **mixed**
@@ -46,8 +47,10 @@ For each library `HexFoo` advancing through Phase 4:
    It registers every compiled operation in the library's SPEC API surface
    with `setup_benchmark` (parametric) or
    `setup_fixed_benchmark` (canonical input). The complexity
-   expression in each `setup_benchmark` is the *textbook*
-   complexity, not the observed one. Proof-track operations instead have
+   expression in each `setup_benchmark` is the independently derived expected
+   family scaling for a two-sided registration or the cited published bound
+   for a one-sided registration, never a model read from observed timings.
+   Proof-track operations instead have
    named fresh-module probes and matched baselines under an explicit manifest
    `proof_probes` directory; those probes are not registrations.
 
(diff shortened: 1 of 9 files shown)


kim-em:
Closing without merge after the release guard confirmed the directive premise is inconsistent. See the evidence comment on #9733. The branch contains a policy implementation and review work, but no honest reconciliation can make all six reports, Phase-4 state, and the released-component invariant agree without source work that #9733 declares out of scope.

PR #9768, opened 2026-08-27
fix(number-field): use fixed isolation benchmarks

### LINKED ISSUE #9728: bench-found: AlgebraicPoly.roots? carries a composed, unciteable complexity model; it may not admit one
library: HexNumberField
depends-on: #9724

## Symptom

**Declared expectation.** `bench/HexNumberField/Bench.lean` registers

```lean
setup_benchmark runAlgebraicRootsLadder n => n ^ 5 * (Nat.log2 (n + 2)) ^ 2
```

deriving `n^5 log^2 n` from the HexRoots heuristic `O(d^3 * B^2)` at norm-eliminant
degree `d = 2n`, with the working bit length taken as `B = O(d log d)`.

**Observed result.** Idle `chungus2`, Lean 4.34.0-rc2, LeanBench 0.1.0, three outer
trials per rung, height-controlled fixture:

| n | median per call | 3-trial spread |
|---:|---:|---:|
| 3 | 192.997 ms | 0.38% |
| 4 | 961.672 ms | 0.12% |
| 5 | 2.568 s | 3.11% |
| 6 | 5.993 s | 0.81% |
| 8 | 18.966 s | 0.67% |

`verdict: inconclusive (cMin=64309.017, cMax=234783.128)`. The harness declines to
fit a slope at all — the ladder spans under two octaves — and falls back to its
range check. Against a continuous proxy the residual slope is about `-1.4`: the
call grows markedly more slowly than the declared model.

**What the profile settles, and what it does not.** samply at 999 Hz, `n = 6`,
5,965 retained samples, calibration residual 0.633 ms, confidence and sensitivity
passed: `QAdjoin.roots?` 91.45%, `componentRoots?` 91.45%, `isolate` / `isolateLoop`
91.87%, `taylor` 82.90%.

So the *phase* is not in doubt — this family's cost really is norm-eliminant
isolation, exactly where the derivation says. Its single squarefree component means
the duplicate-removal path that dominates `runQAdjoinRootsLadder` (#9724) is never
reached. What is in doubt is whether `n^5 log^2 n` is the right model for isolating
*this* family's eliminants.

## What this issue is asking for

**Probably: stop declaring a parametric model for this operation at all.**

The declared `n^5 log^2 n` is not a textbook complexity. It was composed like
this: `HexRoots/SPEC/hex-roots.md` §Complexity contract says `isolate` at degree
`n` and precision `prec` is "**heuristically** `O(n^3 * B^2)` bit operations" —
the SPEC's own word — and this registration then assumed `B = O(d log d)`,
assumed `d = 2n`, and multiplied out. Neither assumption was measured, and the
result is a guess built on a self-declared heuristic. There is no result in the
literature that says certified root isolation of a norm eliminant over an
algebraic coefficient field costs `n^5 log^2 n`. The underlying algorithm
(Becker–Sagraloff–Sharma–Yap) does have proven bounds, but what is declared here
is not one of them.

HexRoots — which owns the algorithm — does not model it parametrically either.
Its Phase-4 validation for isolation is a table of **absolute wallclock
ceilings** at degrees 10, 20, 50 and 100, described as "ceilings for regression
detection, not aspirations", and it demoted its own `runIsolate` from a
parametric registration to a fixed one for exactly this reason (#8750, quoted
below).

`setup_fixed_benchmark` is a first-class choice, not a fallback:
`PLAN/Phase4.md` §Deliverables says each compiled operation is registered
"with `setup_benchmark` (parametric) **or** `setup_fixed_benchmark` (canonical
input)", and `SPEC/benchmarking.md` §Fixed-problem benchmarks is written for
precisely this case — "absolute-wall-clock measurements on canonical hard
inputs, not parameter sweeps".

So the question this issue should settle is not "what is the right exponent for
`AlgebraicPoly.roots?`". It is: **is there an honestly derivable model for this
operation at all, and if not, should it be a fixed registration with a wallclock
budget, mirroring how HexRoots handles the same work?**

## Prior art — read this first

HexRoots hit the same wall on its own `runIsolate` registration and resolved it.
From `reports/hex-roots-performance.md`:

> `runIsolate` was parametric at `n⁵` in the previous revision of this report and
> its `4..10` range check passed, but post-merge review showed the adjacent
> derivation could not support `n⁵` on this family: `separatedPoly n` has
> `log ‖p‖∞ = Θ(n·log n)`, so the `separationDepth` floor makes the emission-level
> bit-length `B = Θ(n²·log n)` and the honestly derived wall from `O(n³·B²)` is
> `~n⁷`, unreachable in the 30 s/call band. Per the no-fitting rule the
> registration was demoted to a fixed case rather than kept on a fitted model.

(#8750, closed.) Demotion to a fixed registration is therefore an accepted outcome
here too, and may well be the right one: if the honest model for this family is
out of reach inside the wallclock band, a parametric ladder carrying a model
nobody can test is worse than a fixed case.

## Deliverables

- Work the ordered rule in `SPEC/benchmarking.md` (#9733) for this operation,
  strongest mode first, and record the outcome in the report:
  - *Mode 1, two-sided.* Is there a derivable expected scaling for isolating
    this family's norm eliminants? The declared `n^5 log^2 n` is not one — it
    was composed by taking HexRoots' self-declared *heuristic* `O(n^3 * B^2)`,
    assuming `B = O(d log d)` and `d = 2n`, and multiplying out.
  - *Mode 2, upper bound.* Is there a published bound covering isolation of
    polynomials of this shape, which the profile confirms is the dominant
    phase? Becker-Sagraloff-Sharma-Yap has proven bounds; establish whether one
    applies here, with all parameters mapped.
  - *Mode 3, fixed.* Otherwise demote to `setup_fixed_benchmark` on a canonical
    input with an absolute budget in the SPEC's §Complexity and Phase 4 budgets,
    following HexRoots' own handling of isolation, and say in the report that
    asymptotic detection is given up for this operation.
  - *Mode 4.* If none of those is honest, the operation stays blocked.
- Record the eliminant's degree after `squareFreeCore`, its `coeffAbsMax` and
  coefficient bit height, and the realised precision, at each rung. A mode-3
  budget still needs its canonical input characterised, and modes 1 and 2 need
  these quantities to derive or check a bound.
- Apply the same question to `runLazyAddLadder`, which carries the same composed
  `n^5 log^2 n`. It currently reports *consistent*, but with `tau = Theta(d)` on
  its eliminant the honest substitution gives `B = Theta(d^2)` and a wall nearer
  `d^7`, so its pass is more likely a coincidence over a two-octave range than a
  validated model. A model that passes for the wrong reason is worth no more
  than one that fails.

## Library placement

`bench/HexNumberField/Bench.lean` for the registration, instrumentation and
schedule; `reports/` for the refreshed evidence. If the outcome is a corrected
model, the derivation belongs in the adjacent comment there.

`HexNumberField/SPEC/hex-number-field.md` §Polynomial roots:
"`AlgebraicPoly.roots?` first embeds all nonzero coefficients into one computed
primitive `QAdjoin`, then invokes the fixed-field algorithm."

1. **Which SPEC § governs this placement?** `PLAN/Phase4.md` §Discipline and
   §Exit criteria; `HexRoots/SPEC/hex-roots.md` for the isolation cost contract.
4. **Does the deliverable presuppose missing infrastructure?** No, though the
   instrumentation may want a HexRoots-side counter surface.

## Context

- `reports/hex-roots-performance.md` and #8750 — the same problem, already solved
  once, with the demote-to-fixed resolution
- `HexRoots/MahlerPrec.lean` (`mahlerPrec`, `separationDepth`),
  `HexRoots/IsolateAll.lean` (`isolate`)
- #9724 — the sibling fixed-field ladder, whose residual has a different cause
  (duplicate removal, not isolation)
- `reports/bench-results/hex-number-field-phase4-scientific-fixture-corrected.json`

## Verification

- the recorded per-rung degree/height/precision table
- `lake exe hexnumberfield_bench run Hex.NumberFieldBench.runAlgebraicRootsLadder`
  returns *consistent with declared complexity*, or the registration is demoted
  with the reason recorded

Sequencing note: #9724 changes `AlgebraicRoot` equality and the exactification
path that `QAdjoin.roots?` calls. This family's profile does not currently reach
that path, so the effect should be small — but "should be small" is a prediction,
and it is cheaper to land #9724 first and re-measure than to argue about it.

## Out of scope

- `runExactLadder` (#9725) and `runLazyAddLadder`. Their residuals may share a
  cause with this one, but that is a hypothesis to test after this ladder's inputs
  are measured, not an assumption to build a shared issue on.
- #9721, #9722, #9724, #9727.

kim-em:
Closes #9728

## Summary
- demote the unsupported lazy-add and algebraic-roots scaling models to fixed canonical regressions
- add compiled eliminant degree, coefficient maximum, bit-height, and isolation-precision instrumentation
- record the ordered mode analysis, canonical budgets, and fresh benchmark evidence

## Verification
- lake build
- python3 scripts/check_phase4.py
- lake exe hexnumberfield_bench verify (both changed targets pass; 12 unrelated optional PARI comparators require cypari2, which is unavailable locally)
- .lake/build/bin/hexnumberfield_bench run Hex.NumberFieldBench.runLazyAddLadder Hex.NumberFieldBench.runAlgebraicRootsLadder
- .lake/build/bin/hexnumberfield_bench isolation-stats


kim-em:
Independent Claude Opus review completed while CI ran. Addressed its substantive findings by making both expensive fixtures lazy, measuring and documenting the 11.84 s smoke cost, correcting libraries.yml coverage, replacing unlanded mode terminology with current SPEC references, renaming realised precision to isolationTarget, adding fixture guards, clarifying ceiling enforcement and project-internal provenance, pinning BSSY v4 and its tau mapping, fixing stale report prose, and regenerating both artifacts from clean commits after rebasing. I did not broaden this issue into demoting the three other registrations that contain an isolation proxy: the report now names them explicitly and their distinct status under #9724/#9733; #9728 authorizes this operation plus runLazyAddLadder, while the directive marks the sibling work out of scope.

PR #9826, opened 2026-08-28
bench(hensel): isolate degree and precision evidence

### LINKED ISSUE #9741: bench-found: HexHensel mixed schedules fail their mode-1 verdicts
library: HexHensel

## Symptom

**Declared expectation.** The nine registrations declare independently derived mode-1 scaling for bridge, linear, quadratic, product, and multifactor lifting over their degree or encoded degree/precision families.

**Observed result.** The current headline export has six passing mode-1 verdicts and three inconclusive results: iterated linear lift, linear multifactor lift, and quadratic multifactor lift. The ordered rule in #9733 does not make these upper-bound passes because mode 1 is honestly derivable; they remain failed mode-1 results.

**Root-cause hypothesis.** The mixed degree/precision schedule and the optimized high-precision rung change the ladder shape. The family encoding may not isolate the intended scaling, or the implementation may be off-model.

## Deliverables

1. Roll `HexHensel.done_through` back to 3 while the three verdicts fail.
2. Diagnose fixture/schedule versus implementation for each failed registration without fitting the model.
3. Re-run scientific evidence and re-promote only after every declared mode passes.

## Context

- Policy issue: #9733
- Earlier investigation: #1335
- Report: `reports/hex-hensel-performance.md`
- Bench: `bench/HexHensel/Bench.lean`

## Verification

- `lake exe hexhensel_bench verify`
- scientific runs for the three affected registrations
- `git diff --check`

kim-em:
Closes #9741.

## Summary

- replace the mixed encoded degree/precision schedules with separate fixed-precision degree ladders and fixed-degree precision ladders
- correct the iterated fixture to use dense monic factors with the explicit Bezout relation `h - q*g = 1`, while keeping the error below the product's leading degree
- retain tight mode-1 `n²` claims on the isolated degree families and add source-derived mode-2 coefficient-bit bounds for the precision families
- correct the product dispatcher/fold/tree wall-clock bounds to include coefficient-width growth found by the all-target rerun
- calibrate the persistent python-flint protocol in-harness and publish current ratios across all 30 paired rungs
- replace the stale headline report with one clean-revision, fourteen-registration campaign and one confidence-passing profile per declared input family
- promote `HexHensel.done_through` from 3 to 4 only after every declared mode passes

## Scientific result

Eight mode-1 registrations are consistent with their independently derived tight models. Six mode-2 registrations are reported as `within declared upper bound (observed faster)`; their adjacent derivations use coefficient bit width and published schoolbook integer-operation bounds, not fitted slopes. Both multifactor compare groups agree on every common parameter.

The committed clean artifacts are:

- `reports/bench-results/hex-hensel-c22f45d2-headline-chungus2.json`
- `reports/bench-results/hex-hensel-c22f45d2-flint-chungus2.json`
- `reports/bench-results/hex-hensel-c22f45d2-modp-confirm-chungus2.json`

## Verification

- `lake build` (10,500 jobs)
- `uv run --with python-flint .lake/build/bin/hexhensel_bench verify` (75 registrations)
- degree and precision `hexhensel_bench compare` groups (all common hashes agree)
- clean scientific runs for all 14 registrations, three outer trials per rung
- clean five-trial stability confirmation for `runModPChecksum`
- four samply/lean-bench-samply timed-region profiles, all confidence-passing
- `git diff --check`


kim-em:
Closing after the corrected monic, co-varying degree/precision experiment showed that all three declared mode-1 models still fail. The fixed-precision pass in this PR bypassed the slope test and did not validate precision scaling. Detailed counterevidence is recorded on #9741; HexHensel must remain at done_through 3 pending a revised directive.

PR #9831, opened 2026-08-28
fix(hex-primality): bound duplicate-subject preflight

### LINKED ISSUE #9775: fix(hex-primality): bound duplicate-subject preflight asymptotics
Parent directive: #9752
Found by the independent certificate/order review #9753.

## Finding

`subjectsOk` in `HexPrimality/Cert.lean` calls `rest.any` for every factor entry. It therefore performs `k(k-1)/2` subject comparisons on a `k`-entry list before certificate recursion. The list is untrusted and can be arbitrarily long on a rejected certificate.

The SPEC describes cheap structural rejection before arithmetic and gives the one-level/full-tree checker costs only as linear-in-entry modular-multiplication counts. Those counts are true as counts of modular multiplications, but they omit a quadratic attacker-controlled preflight term. This is a wrong asymptotic shape for the proof-bearing checker boundary unless explicitly accepted and budgeted.

## Deliverables

- Give the factor-subject invariant a kernel-replayable check whose accepted contract and worst-case rejected-input cost are honest; a canonical sorted/distinct representation is one possible route.
- Preserve the positive-exponent/child-subject invariant and Pocklington soundness proof without introducing trusted preprocessing.
- Update the SPEC complexity statement to include all asymptotically material checker work.
- Add long-list and late-duplicate rejection coverage.

## Verification

- `lake build HexPrimality HexPrimalityKernelProbe`
- focused checker cost/route regression

depends-on: none

kim-em:
Closes #9775

## Summary
- require canonical strictly ascending factor subjects with a linear kernel-replayable check
- sort untrusted search output before the public checker validates it
- account for structural, bounded-product, division, and gcd work in the checker complexity contract
- add long-list and final-slot duplicate native and kernel replay regressions

## Verification
- `lake build HexPrimality HexPrimalityKernelProbe`
- `lake build HexPrimalityKernelProbe`
- `lake build hexprimality_bench HexManual`


kim-em:
Independent Opus review follow-up:

**Agreed and addressed in `fac1684b3`:** aligned the Python oracle with strict ascending subjects; added distinct-but-unsorted Lean/conformance/fixture rejection; documented the producer-facing canonical format; corrected the subject-comparison count and tree-bound rationale; updated the kernel-probe module contract.

**Not taken:** the suggested `Nat.blt` wrapper micro-optimization was not needed for the asymptotic fix; `certProduct` recursion is pre-existing and outside this issue; an extra search-order-only guard would duplicate the existing certificate-search round-trip that already fails if canonicalization is removed.

Post-review verification is green for `HexPrimality`, `HexPrimalityKernelProbe`, fixture emission/regeneration, and the full `HexConformance` aggregate. The focused Python oracle regression passes; the full local oracle self-check skips because `cypari2` is not installed.

PR #9845, opened 2026-08-30
fix(hex-primality): integrate counted p-minus-one search

### LINKED ISSUE #9783: fix(hex-primality): integrate shared p-minus-one into certificate factoring
Parent directive: #9752
Found by skeptical runtime-route review: #9754

## Finding

`HexPrimality/PMinusOne.lean` implements genuine capped Pollard p−1 stage 1 and HexIntFactor consumes it, but `HexPrimality/Search.lean::partialFactor` still imports no p−1 module and runs only stored-table trial division followed by the rho worklist. This contradicts the SPEC section “Taking up downstream factoring advances”, which says the shared stage-1 primitive joins rho and widens `partialFactor` cheaply once HexIntFactor lands.

The shared primitive also does not yet have the SPEC-promised resumable/counted failure shape: `pMinusOneStage1` returns only `PMinusOneResult`, with no attempt total or state, so HexIntFactor reconstructs route charges from `smooth.events.length`. The deterministic primitive need not invent Rand use, but its public/counting boundary and both consumers must expose one consistent semantic attempt contract.

Focused evaluation confirmed the primitive distinguishes all three terminal gcd outcomes and finds factor `13` for `(n,base,bound) = (299,2,5)`; the missing behavior is integration and shared accounting, not a fake p−1 algorithm. The SPEC’s private `partialFactor` signature is also stale: it still shows `PartialFactors × Rand` plus a positivity hypothesis, while the implementation returns counted `PartialSearch` and proves its product invariant for every `Nat`.

## Deliverables

- Add a bounded p−1 stage-1 phase to certificate partial factorization at the specified point relative to table trial division and rho.
- Specify and implement its base/bound/fuel policy without importing HexIntFactor or duplicating that dispatcher.
- Provide the shared counted/resumable attempt boundary promised by the SPEC, keeping deterministic p−1 state behavior explicit and adapting HexIntFactor to consume it directly.
- Preserve the partial-product theorem, exact advanced Rand state, and attempt accounting; count each p−1 call exactly once.
- Distinguish `noFactor`, proper factor, and whole-modulus outcomes and retain an unsplit residual on exhaustion.
- Align the SPEC’s private partial-search shape and theorem with the counted implementation.
- Add p−1-friendly route evidence showing certificate reach improves and checker replay remains the only acceptance boundary.

## Verification

- `lake build HexPrimality HexConformance HexIntFactor`
- focused p−1 factor/no-factor/whole, counted failure, and certificate-route evaluations

depends-on: none

kim-em:
## Summary

- add a shared counted Pollard p−1 stage-1 boundary with explicit attempt and unchanged Rand state
- run one bounded base-2/bound-64 p−1 phase between certificate trial division and rho while preserving the partial-product theorem
- make HexIntFactor consume the shared count directly instead of reconstructing charges from route events
- align both SPECs and add factor/no-factor/whole, exhaustion, exact-state, and checker-replay route coverage

## Verification

- `lake build HexPrimality HexConformance HexIntFactor`
- focused `HexPrimality.PMinusOne`, `HexPrimality.Search`, `HexPrimality.Conformance`, `HexIntFactor.Factor`, and `HexIntFactor.Conformance` builds

Fixes #9783


kim-em:
Claude Opus second-opinion review found no soundness or attempt/Rand accounting defect. I addressed its actionable merge concerns in 4a1e09127:

- eliminated a duplicate Miller–Rabin screen by moving probable-prime cofactors directly into the factor list before rho
- strengthened the p−1 phase product invariant over both accumulated factors and the worklist
- removed the unnecessary public `@[expose]` wrapper body and its redundant projection lemmas
- added direct owner-library factor/no-factor/whole, exact-attempt, Rand-preservation, cap, and factor-contract coverage

The fixed base/bound policy remains intentionally independent of the rho-only `PrimeCertBudget` and is governed by positive partial-search fuel as documented. The other low findings concerned pre-existing internals or optional follow-up optimization. Reverification: `lake build HexPrimality HexConformance HexIntFactor`; targeted `runCertSearch` sweep completed with the 512-bit rung at ~11 ms (5 s per-call cap).

PR #9870, opened 2026-08-31
perf(hex-poly-fast): remove Karatsuba traversal allocation

### LINKED ISSUE #9744: HexPolyFast: eliminate Karatsuba inner-loop and block-tail allocation
Phase 2 follow-up from the skeptical scaffolding review.

Karatsuba.Raw.schoolbookCoeff constructs List.range a.size for every output coefficient. This adds substantial allocation in every base-case convolution even though the shorter operand now correctly drives the loop.

Karatsuba.Raw.blocks also recurses on Raw.high blockSize long. That materializes the entire remaining suffix at each block, giving quadratic suffix-copy traffic in the number of blocks even though coefficient multiplication itself is correctly blocked.

Replace the diagonal fold with an allocation-free index loop, and carry an offset into the original long coefficient array during block traversal. Preserve the existing ofCoeffs agreement theorems, ratio-under-2 and 64:1 conformance, and benchmark both changes against the current implementations.

kim-em:
Closes #9744

## Summary
- replace per-diagonal `List.range` folds with allocation-free index loops while preserving the raw schoolbook specification
- carry an offset through unbalanced Karatsuba block traversal so each block is copied once instead of copying every remaining suffix
- retain proofs through `ofCoeffs`, benchmark both kernels against their former implementations, and synchronize the performance-report registry

## Verification
- `lake build HexPolyFast.Karatsuba HexPolyFast.Conformance hexpolyfast_bench hexpolyfast_emit_fixtures`
- fresh conformance fixture output matches `conformance-fixtures/HexPolyFast/polyfast.jsonl`
- oracle fallback checked all 120 emitted cases (local `python-flint` unavailable)
- `lake exe hexpolyfast_bench verify` passes all 57 local targets; the seven FLINT-backed fixed targets fail only because `python-flint` is unavailable locally
- Phase-4 headline/registration guard passes for all four new targets
- focused warm-cache comparisons at n=1024: diagonal list 24.213 ms vs loop 7.457 ms; at 64:1 n=256: suffix-copy 114.626 ms vs offset 28.791 ms; paired hashes agree

### DIFF
--- a/HexPolyFast/Karatsuba.lean
+++ b/HexPolyFast/Karatsuba.lean
@@ -174,6 +174,16 @@ def combine (k : Nat) (z₀ z₁ z₂ : Array R) : Array R :=
 
 /-- One schoolbook convolution diagonal over raw arrays. -/
 def schoolbookCoeff (a b : Array R) (d : Nat) : R :=
+  Id.run do
+    let mut acc := 0
+    for i in [0:a.size] do
+      acc := if d < i then acc
+        else if d - i < b.size then acc + a.getD i 0 * b.getD (d - i) 0
+        else acc
+    return acc
+
+/-- List-based specification of one raw schoolbook diagonal. -/
+private def schoolbookCoeffList (a b : Array R) (d : Nat) : R :=
   (List.range a.size).foldl
     (fun acc i =>
       if d < i then acc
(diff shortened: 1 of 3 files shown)


kim-em:
Fresh Claude Opus second opinion completed while CI was running.

Accepted and addressed:
- made the retained old raw comparators generic over the same `CommRing` interface as production
- normalized all four raw benchmark outputs through `ofCoeffs` before hashing
- synchronized the performance-report registry and added the required explicit cost-model derivations

Independently verified: the source/proof chain is sound, `Raw.ofCoeffs_schoolbook` and `Raw.ofCoeffs_blocks` remain intact, and the named list/suffix allocations are removed.

Not expanded in this PR: a single mutable output buffer for `addShift`, a block-count scaling campaign, and narrowing the diagonal loop bounds. Those are broader/pre-existing optimization questions; #9744 specifically requests an allocation-free diagonal loop and offset-based block traversal, and no theorem or conformance obligation was weakened.

PR #9872, opened 2026-08-31
perf(poly-fast): stop retaining product-tree levels

### LINKED ISSUE #9745: HexPolyFast: align ProductTree storage with its node-product API
Phase 2 follow-up from the skeptical scaffolding review.

ProductTree stores every balanced level in levelData, but nodeProduct? ignores those cached products and recomputes the requested node by folding its leaf block. The only levelData consumers are the level observation accessors and conformance guards; real downstream users currently read only root.

Choose one coherent representation: either prove the stored-level indexing law and return cached nodes from nodeProduct?, or remove eager level storage and retain the documented semantic fold. Measure ProductTree construction after the change. Also assess whether the executable PointNode and general RemainderNode trees can share one representation now that both use sibling-derived internal capacities.

kim-em:
Closes #9745

## Summary

- retain only the original leaves and balanced root product in `ProductTree`
- reconstruct the observational level API on demand while preserving semantic node products
- document why proof-indexed point and general remainder trees remain separate
- extend conformance coverage for reconstructed, empty, singleton, and root levels

## Premise validation

Only `root` has production downstream consumers; `level?` is observational/conformance API and `nodeProduct?` is already specified as a semantic fold over its leaf block. The point and general remainder nodes share the count-balanced split and equivalent sibling-capacity arithmetic for linear point leaves, but their dependent proof payloads and traversals differ enough that a common representation would weaken invariants or add abstraction without code sharing.

## Benchmark

`Hex.PolyFastBench.runProductTree`, warm cache on `chungus2`, Lean 4.34.0-rc2, compared with the stored clean baseline at `53ef234e` (the ProductTree source and benchmark body are unchanged between that baseline and this branch base):

| leaves | baseline | candidate | delta |
|---:|---:|---:|---:|
| 256 | 513.263 us | 475.209 us | -7.4% |
| 1,024 | 4.874 ms | 4.899 ms | +0.5% |
| 4,096 | 48.132 ms | 48.515 ms | +0.8% |
| 16,384 | 473.481 ms | 468.598 ms | -1.0% |

Peak RSS across the ladder fell from 82,784 KiB to 66,188 KiB (-20.0%). The candidate remains consistent with the declared Karatsuba complexity. A separate cold three-trial ladder over 256--4,096 leaves was also consistent.

## Validation

- `lake build HexPolyFast.Tree`
- `lake build HexPolyFast.Conformance`
- `lake build HexPolyFast.Conformance HexManual.Chapters.HexPolyFast` after rebasing onto current `origin/main`
- `.lake/build/bin/hexpolyfast_bench verify Hex.PolyFastBench.runProductTree`
- full `lake build` before and after rebasing onto current `origin/main`
- no added `axiom`, `sorry`, or `native_decide`

### DIFF
--- a/HexManual/Chapters/HexPolyFast.lean
+++ b/HexManual/Chapters/HexPolyFast.lean
@@ -221,6 +221,8 @@ from their represented leaf blocks.
 
 {docstring Hex.DensePoly.ProductTree.build}
 
+{docstring Hex.DensePoly.ProductTree.level?}
+
 {docstring Hex.DensePoly.ProductTree.root}
 
 {docstring Hex.DensePoly.ProductTree.nodeProduct?}
--- a/HexManual/Chapters/HexPolyFast.lean
+++ b/HexManual/Chapters/HexPolyFast.lean
@@ -214,9 +214,10 @@ sequence; a checked fallback preserves exact executable behavior.
 tag := "hex-poly-fast-multipoint"
 %%%
 
-{name}`Hex.DensePoly.ProductTree` stores a balanced tree of nonempty levels.
-Its observations expose the original leaves, the root product, and individual
-node products while hiding the internal level representation.
+{name}`Hex.DensePoly.ProductTree` caches the root of a balanced tree without
+retaining its intermediate levels. Its observations expose the original
+leaves, reconstruct levels on demand, and compute individual node products
+from their represented leaf blocks.
 
 {docstring Hex.DensePoly.ProductTree.build}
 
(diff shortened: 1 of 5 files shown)


kim-em:
Fresh Claude Opus second opinion completed while CI was running. It found no blocking correctness defect and independently checked `levelCount`, `level?`, `buildRoot`, `nodeProduct?`, and the shared tree shape across empty, singleton, power-of-two, and non-power-of-two sizes.

Addressed in `3f4282b4b`:
- removed stale descriptions of an eager level representation
- documented the empty-tree `#[1]` level convention and rendered `level?` in the manual
- clarified that point and general remainder trees compute the same linear-leaf capacity through different formulas

The review also suggested formal observation-indexing theorems and shared take/drop proof helpers. I verified these are optional extensions, not defects: issue #9745 explicitly permits removing eager storage while retaining the documented semantic fold, and no production consumer uses the observational level API. I kept them out of this focused change.

The requested construction measurement is in the PR body; the matched large-size timings are neutral and peak RSS falls about 20%.

PR #9874, opened 2026-08-31
chore: centrally manage released-mirror CI caching

### LINKED ISSUE #9373: chore: centrally manage released-mirror CI caching
## Current problem

Released mirrors use leanprover/lean-action@v1 with build: false, then build the root, conformance, and bench projects in later steps. The composite action saves its cache before those builds, so the outputs actually produced by CI are not persisted and sidecar .lake trees are not covered.

The CI skeleton is currently outside the sync-managed paths. Direct coordinated edits to released mirrors would contradict hex-dev's source-of-truth model.

## Deliverables

- Establish a single CI workflow template or generated artifact owned in hex-dev.
- Extend the release-sync manifest/driver so the mirror CI skeleton is previewed and published from that source of truth, with the same divergence guard as other managed content.
- In the managed workflow:
  - disable lean-action's GitHub cache;
  - explicitly cache the root, bench, and conformance build outputs with a run-unique save key and stable restore prefix;
  - disable lean-action's automatic Mathlib cache when the workflow already runs the explicit Mathlib cache fetch.
- Apply the generated workflow uniformly to every released mirror to which the configuration applies.
- Preserve the existing one-job CI shape and all existing build/conformance/bench commands.

## Verification

- Release-sync dry-run shows only the intended managed workflow changes.
- A representative Mathlib-free mirror and Mathlib-backed mirror restore caches produced after their real builds.
- Root and sidecar builds remain green.
- No released repository is edited outside the guarded sync path.

kim-em:
Closes #9373.

## Summary

- add a complete centrally owned CI workflow artifact for every released mirror
- publish `.github/workflows/ci.yml` through the guarded release sync, including the aggregate mirror
- disable lean-action caching and explicitly restore/save root, sidecar, and published-Hex dependency build outputs only after verification completes
- keep main runs alive through their terminal cache save while retaining cancellation for superseded PR runs
- validate workflow coverage, triggers, cache paths/keys/order, helper availability, and no-matrix policy
- document the required Contents and Workflows fine-grained PAT grants and the managed boundary
- print the managed workflow diff during release dry runs

## Verification

- `python3 -m unittest discover -s scripts/release -p "test_*.py"` (54 tests)
- `python3 scripts/release/check_released_manifest.py`
- `python3 scripts/release/sync_released.py --dry-run --only hex-arith`
- `python3 scripts/release/sync_released.py --dry-run --only hex-poly-mathlib`
- full release-sync dry run (all 45 targets; divergence guard preserved)
- verified every referenced `scripts/ci` helper in live divergent mirrors
- compared all 45 managed workflows against live mirror `main`; before the intended cache/concurrency changes, commands and job shapes are byte-for-byte identical
- `lake build`
- `lake build HexConformance`
- `lake build hexarith_bench hexpoly_bench`


kim-em:
Fresh Claude Opus second-opinion review completed while CI was running.

Validated and addressed:
- publishing PATs need separate Workflows read/write permission for managed `.github/workflows/ci.yml`; docs and CLI diagnostics now say so
- caches must include released Hex dependency builds under each `.lake/packages` tree while still excluding Mathlib
- main runs must not be cancelled before the terminal cache save
- checker now verifies exact restore/save paths, exact restore-prefix/key relationship, triggers, concurrency, and job shape
- sync now validates referenced unmanaged `scripts/ci` helpers and dry runs print the workflow diff
- mutation coverage now includes pins-only CI overwrite and cache-policy failures

Not adopted: sweeping unrelated extra workflow files and broader parser/performance refactors, which are outside this issue and do not affect the managed `ci.yml` contract. The suggested command/job-fidelity concern was independently checked against all 45 live mirror workflows.

PR #9905, opened 2026-09-02
perf: byte-chunked bit primitives for hex-graph-iso

kim-em:
This PR replace the per-bit recursions behind popCount, set-bit listing, and lowest-set-bit with byte-chunked implementations attached by @[csimp] equalities: per-byte tables defined as the specification over range 256, one shift per byte, early exit on empty remainders. Kernel- and proof-facing definitions are untouched, so the kdecide ladder and all proofs are unaffected; the fixture stream is byte-identical. The run stage improves 20-41% across the profile instances, with the largest wins on the families past the 63-bit small-scalar boundary (circulant64 from 10.1x to 6.9x nauty, gnp96 from 19.3x to 12.8x). The recorded sweep, figures, and before/after comparison ride along per the SPEC's benchmarks requirements.

🤖 Prepared with Claude Code


kim-em:
Before/after cactus comparison per SPEC § Benchmarks (trace-driven producer → byte-chunked bit primitives):

before/after cactus

|---|---|---|---|
| canonical hex fast tier | 0.88x | 20.2ms -> 15.8ms | 3.0ms -> 2.0ms |
| canonical hex checked tier | 0.93x | 206.2ms -> 183.8ms | 23.3ms -> 21.8ms |
| graph_iso tactic | 1.02x | 253.32s -> 256.11s | 103.03s -> 103.03s |
| pairs hex fast tier | 0.93x | 4.0ms -> 3.6ms | 674us -> 600us |
| pairs hex checked tier | 0.96x | 29.5ms -> 28.2ms | 6.7ms -> 6.3ms |

Curve shift is the rank-wise geometric-mean ratio of the sorted curves. The tactic row is flat by design (these primitives never enter the kernel path; 1.02x is repeat noise). Animation now three frames: reports/figures/hexgraphiso-cactus-animation.gif.

PR #9906, opened 2026-09-02
perf: limb-kernel image and permset transport for hex-graph-iso

kim-em:
This PR attach limb-kernel runtime implementations to the image and permset bitset transport by @[csimp] equalities: the set is peeled once into 63-bit small-Nat limbs, per-bit work runs entirely on small scalars against an in-place limb accumulator, and single-limb calls (every n ≤ 63 instance) never touch GMP. The kernel- and proof-facing definitions are untouched; the ~230-line equality proof transfers by testBit extensionality through the limb representation. The certificate replay and translation both ride this transport, so the checked tier improves 13-31% on the profile instances with the fast tier untouched. Recorded sweep, figures, and before/after comparison ride along per the SPEC's benchmarks requirements.

🤖 Prepared with Claude Code


kim-em:
Before/after cactus comparison per SPEC § Benchmarks (byte-chunked primitives → limb-kernel image transport):

before/after cactus

|---|---|---|---|
| canonical hex fast tier | 1.00x | 15.8ms -> 15.9ms | 2.0ms -> 2.0ms |
| canonical hex checked tier | 0.94x | 183.8ms -> 159.2ms | 21.8ms -> 20.0ms |
| graph_iso tactic | 1.01x | 256.11s -> 266.65s | 103.03s -> 110.03s |
| pairs hex fast tier | 0.99x | 3.6ms -> 3.6ms | 600us -> 600us |
| pairs hex checked tier | 1.00x | 28.2ms -> 27.2ms | 6.3ms -> 5.7ms |

Fast tier flat by design (this is checked-path transport); the tactic rows are kernel-time repeat noise (the kernel path is definitionally unchanged by csimp). Animation now four frames.

PR #9915, opened 2026-09-02
perf: emission checks reduced to the closure-unimplied conjuncts

### LINKED ISSUE #9895: graph-iso wave 2: record refine outputs in the trace for the translator
On `graphiso-opt`, translation (trace → certificate, `certifyNodeAutom` + `achieverCodes`) costs ~9.4ms of paley61's 19.6ms `canonicalizeChecked`, structurally mirroring the trusted replay's per-node `refine` recomputation — but the translator is UNTRUSTED, so unlike the replay it does not have to recompute anything: the traced search can record per-node refine outputs (or just the fields `childAutomOk`/target-cell selection need) and the translator can consume them directly. SPEC § Trace-driven production already frames the trace as arbitrary untrusted data; extending `TraceRun` carries no proof obligations.

Requirements: recording must not perturb the untraced fast path (separate entry point already exists) or the traced walk's observables (fixture stream byte-pinned, traced/untraced RunResult identical); certificate record counts must not change (92/42/67 on the profile instances); the replay is untouched. Measure which recorded subset pays: full refine outputs may cost more to allocate than they save — start by recording only what the per-child emission decisions consume, and measure the allocation delta in the traced run (currently 499µs vs 495µs untraced; keep the overhead proportionate).

Estimated win from profiling: roughly half of translation, ~4-5ms off paley61 canonicalizeChecked. Verification: profile produce/canonchecked before/after at two rep counts on a quiet machine; fixture byte-identity; stats records unchanged; full build; campaign.

🤖 Prepared with Claude Code

kim-em:
This PR reduce the certificate translator's per-record emission check to the conjuncts admission closure does not already imply: witness compositions of individually-verified generators are automorphisms by closure, so emission now evaluates only the earlier-offset bound and the cell-transport predicate (childCellsOk), which is the conjunct a witness-search bug could actually violate. Profiling attributed ~80% of translation to the skipped re-validation, refuting the recorded-refine-outputs hypothesis of https://github.com/kim-em/hex-dev/issues/9895 (refine is 2-3% of the translator; no trace extension is warranted). Nothing trusted moves: the replay's full acceptance conjunction is unchanged and proof-carrying, admission still filters every raw trace entry through isautom, and a corrupt trace still costs only wall clock — now as a replay rejection into the exhaustive fallback rather than graceful descent, a trade the 32,788-case campaign backstops. Translation drops 9.8 to 1.7ms on paley61 and canonicalizeChecked from 20.1 to 11.4ms in isolation; the recorded checked curve shifts 0.77x. Certificate records are unchanged (92/42/67), fixtures byte-identical, and the SPEC's layer-two sharing bullet now states the admission-closure split. Recorded sweep, figures, and comparison ride along.

Closes https://github.com/kim-em/hex-dev/issues/9895.

🤖 Prepared with Claude Code


kim-em:
Before/after cactus comparison per SPEC § Benchmarks (allocation-free rows → closure-implied emission checks):

before/after cactus

|---|---|---|---|
| canonical hex fast tier | 0.99x | 13.5ms -> 13.6ms | 2.0ms -> 2.1ms |
| canonical hex checked tier | **0.77x** | 152.0ms -> 104.8ms | 19.7ms -> 11.2ms |
| graph_iso tactic | 1.02x | 26.59s -> 26.53s | 12.13s -> 12.12s |
| pairs hex fast tier | 1.02x | 3.4ms -> 3.5ms | 561us -> 574us |
| pairs hex checked tier | 0.83x | 25.8ms -> 20.2ms | 5.4ms -> 3.9ms |

Certificate records exactly 92/42/67 (the number that must not move). Animation now nine frames; the γ-dedup replay landing compounds this next.

PR #9916, opened 2026-09-02
perf: validate each distinct automorphism once in the certificate replay

### LINKED ISSUE #9900: graph-iso wave 2: validate each distinct automorphism once in the replay
Measured on graphiso-opt (see https://github.com/kim-em/hex-dev/issues/9898#issuecomment-5505765770, which refuted the byte-chunking premise and found this instead): `imageGo` reached through `checkAutom` is 17.34% of the checked path — the largest single symbol — and mostly redundant, because certificates reuse a few distinct generators across many `.autom` records: paley61 has 89 autom records over 11 distinct γs, circulant64 has 64 over 2. The replay validates the same γ from scratch at every record.

Two candidate designs, measure before choosing:

1. **Validated-γ accumulator in the checker**: thread a list of already-validated generators through `checkNode`/`checkChildren`; at each `.autom` record, an O(n) array-equality lookup against the accumulator replaces the O(n · rows) `checkAutom` re-validation on hits. No certificate-format change; the soundness proofs gain an invariant (every accumulator member passed `checkAutom`) threaded through the existing induction.
2. **Generator side-table in the certificate format**: `.autom` records reference an index into a per-certificate γ table validated once up front. Bigger win on lookup cost and certificate size, but changes `CertNode`, the producer/translator, the reifiers in Tactic.lean, and the whole soundness induction — and the kernel replay obligations for the tactic shrink correspondingly (the kernel currently re-validates γs too, so this compounds with the literal-tie work).

Constraints: the checker is trusted — full soundness proofs updated, no weakening; kdecide ladder must keep kernel-reducing; fixture stream byte-identical; certificate-size conformance guards respected (design 2 should shrink certificates, never grow them); no axioms, no native_decide, Mathlib-free. Coordinate with https://github.com/kim-em/hex-dev/issues/9894 (row plumbing) and note the tactic-side reifier impact for design 2.

Verification: perf showing the checkAutom share collapsed; ckey/canonchecked before/after at two rep counts on a quiet machine; stats records; full battery.

🤖 Prepared with Claude Code

kim-em:
This PR make the trusted replay validate each distinct generator once: checkKey collects the certificate's generators (fuel-bounded, deduplicated), filters them through checkAutom up front, and threads the validated list through checkNode/checkChildren as a read-only parameter, with .autom records doing a membership lookup — certificates reuse a handful of distinct generators across many records (paley61: 89 records over 11), and per-record revalidation was ~70% of the replay. Soundness gains one hypothesis threaded through the existing induction with no weakening; equality routes through List BEq on toList because the Array BEq instance does not kernel-reduce; the kdecide ladder covers the collector, filter, and lookup; checkKeyFlat receives the same validated list over its literal-rebuilt rows (an arity-forced integration) so the negative literal route stays proven and in sync. The replay drops 74% on paley61 (8.8 to 2.3ms) and the recorded checked curve shifts 0.75x with the slowest instance at 5.9ms; the tactic rows are flat because the kernel already deduplicates syntactically identical subterm evaluations, as the earlier sharing probe measured. Records unchanged (92/42/67), fixtures byte-identical, campaign green. Recorded sweep, figures, and comparison ride along.

Closes https://github.com/kim-em/hex-dev/issues/9900.

🤖 Prepared with Claude Code


kim-em:
Before/after cactus comparison per SPEC § Benchmarks (closure-implied emission → deduplicated generator validation):

before/after cactus

|---|---|---|---|
| canonical hex fast tier | 0.98x | 13.6ms -> 13.3ms | 2.1ms -> 2.0ms |
| canonical hex checked tier | **0.75x** | 104.8ms -> 67.1ms | 11.2ms -> 5.9ms |
| graph_iso tactic | 0.99x | 26.53s -> 26.94s | 12.12s -> 12.21s |
| pairs hex fast tier | 0.98x | 3.5ms -> 3.4ms | 574us -> 563us |
| pairs hex checked tier | 0.86x | 20.2ms -> 16.1ms | 3.9ms -> 2.7ms |

Cumulative checked-path story across frames 2-10: paley61 slowest instance 26.9ms -> 5.9ms. Animation now ten frames.

PR #9919, opened 2026-09-02
bench: add irregular negative pairs to the graph-iso cactus corpus

### LINKED ISSUE #9918: graph-iso: add irregular negative pairs to the cactus corpus
Every negative pair in the cactus corpus is regular, which is exactly the class where the invariant separators (https://github.com/kim-em/hex-dev/issues/9903) cannot fire at the root and certificates must run — the separator landing was corpus-flat for that reason alone. Add a small set of irregular negative pairs (different degree sequences at matched sizes, plus at least one pair distinguished only at the individualized tier) so the pairs plot covers the easy class and the tiered routing's effect is visible. This is a benchmark-set change: document the series break in the animation metadata (a labelled frame marking the corpus change) and in the SPEC's benchmarks section per the existing series-break convention; all four tiers re-time on the extended set in the same PR.

🤖 Prepared with Claude Code

kim-em:
This PR extend the pairs corpus with four irregular negative pairs built from the deterministic families (grid-vs-circulant, grid-vs-hypercube, grid-vs-complete-bipartite, grid-vs-two-grids at n = 12/16/20/24), covering the class where the invariant separators fire at the root — every previous corpus negative was regular and routed through certificates, so the separator landing was corpus-flat by construction. All four new pairs verify as root-separable, their polarity is certain by distinct degree multisets and revalidated at runtime, and the bipartite pair doubles as a generator-heavy checked-path outlier the corpus lacked. Two constructions were sought and honestly not found from the current generators: an equal-degree-sequence irregular root-inseparable pair, and a pair exercising the individualized rescue tier; both are recorded for future corpus work. This is a benchmark-set change: the SPEC's benchmarks section documents the series break, the recorded sweep lands as its own labelled animation frame, and curve comparisons are meaningful only within a corpus. The sweep wrapper also gains a hardening fix found during this work: it now stages the relevant paths before fingerprinting, so an unstaged edit can no longer record data under a stale key.

Closes https://github.com/kim-em/hex-dev/issues/9918.

🤖 Prepared with Claude Code


kim-em:
Series break: corpus extension, so no curve-shift table (comparisons are only meaningful within a corpus). The new pairs and their measured tier routing:

| pair | n | tactic tier | nauty | fast | checked | tactic |
|---|---|---|---|---|---|---|
| neg-grid3x4-vs-circulant12 | 12 | root | 9us | 53us | 0.24ms | 0.33s |
| neg-grid4x4-vs-q4 | 16 | root | 14us | 99us | 0.52ms | 0.62s |
| neg-grid4x5-vs-k8-12 | 20 | root | 30us | 419us | 19.91ms | 1.31s |
| neg-grid4x6-vs-2grid3x4 | 24 | root | 27us | 174us | 0.69ms | 3.99s |

The root-separator tier shows its measured value against the regular negatives' certificate route (~12s worst on the previous corpus). Animation gains its eleventh, series-break-labelled frame.

PR #9921, opened 2026-09-02
perf: fused cell walks in the refine passes

### LINKED ISSUE #9917: graph-iso: splitter allocation in the refine machinery
After the wave-2 landings, the fast path's remaining allocator traffic concentrates in Refine.lean's splitter machinery — cells_go, countsOf, refineTrivial_go — per the allocation fork's closing profile on https://github.com/kim-em/hex-dev/issues/9896 (which also established, with IR evidence, that record updates and rowsOf were already clean). List conses in the pending-splitter walk and count buckets are the suspects. Compiled-only restructuring: index loops or reused arrays in place of List construction, verified at the IR level, with the fixture stream byte-checked after every step (refine's observables — codes, cells, positions — are conformance-pinned). Proof contact possible in CellPerm's statements about these functions: prefer definitional-equality-preserving shapes or csimp splits per house pattern.

Verification: perf allocator share before/after, run-stage timings at two rep counts on a quiet machine, fixtures, full build, sweep+figures per the freshness check.

🤖 Prepared with Claude Code

kim-em:
This PR fuse the refine passes' cell walks: both passes built a boxed pair list of every cell per pass and then walked it once; the replacement recurses on boundary indices against a partition snapshot bound before the pass writes, trading a per-cell pair-and-cons allocation for one array copy at the pass's first write. Attached by @[csimp] equalities with short structural inductions; proof-facing definitions untouched, so the CellPerm surface is unaffected. Run stage improves 5-10% across the profile instances (paley61 at 295us). The commit also records two measured refutations: the labelling-handoff hypothesis (the compiler's borrow inference already passes the field uniquely — the suspect lean_incs are all in the shared-fallback path) and confirmation that the pipeline's textual duplication is fully CSE'd. The closing profile shows no single fast-path lever at 5% or more remains; the residual counting-sort List churn and distributed big-Nat set tax are one composite design problem for the whole-search-state migration campaign. Fixtures byte-identical; recorded sweep, figures, and comparison ride along.

Closes https://github.com/kim-em/hex-dev/issues/9917.

🤖 Prepared with Claude Code


kim-em:
Before/after cactus comparison per SPEC § Benchmarks (corpus series break → fused cell walks; first comparison within the extended-corpus era):

before/after cactus

|---|---|---|---|
| canonical hex fast tier | 0.94x | 13.4ms -> 12.5ms | 2.0ms -> 2.0ms |
| canonical hex checked tier | 0.95x | 67.5ms -> 63.9ms | 5.9ms -> 5.9ms |
| graph_iso tactic | 0.98x | 33.72s -> 33.69s | 12.11s -> 12.21s |
| pairs hex fast tier | 0.99x | 4.2ms -> 4.2ms | 562us -> 537us |
| pairs hex checked tier | 1.00x | 37.5ms -> 37.2ms | 19.9ms -> 19.8ms |

Animation now twelve frames.

PR #9933, opened 2026-09-02
perf: certificate replay over kernel-cheap list state

### LINKED ISSUE #9920: graph-iso: kernel-cheap row and partition access in the certificate replay
The kernel-negative cost is now precisely attributed (SWAR execution on https://github.com/kim-em/hex-dev/issues/9902): the replay's refine spends its kernel time walking Array/List spines per element — splitCellLoop, windowScan, writeSegment, and lab[...]! probes at ~14-35us per step — not on arithmetic. The literal-tie work (https://github.com/kim-em/hex-dev/issues/9897) already proved the pattern that fixes this class: rebuild the data once from a tied literal into a shape the kernel accesses cheaply (bare-match chunked lists, atD accessors), proven equal to the trusted form once.

Extend that pattern through the checker proper: a checkNodeLit variant whose Ctx carries pre-chunked row lists and whose partition state uses kernel-cheap accessors, proven equal to checkNode, emitted by the tactic on the negative route (and inherited by the positive lit route's general case). The γ-dedup validated-generator list and the held SWAR machinery both compose with it — SWAR lands afterward when popCount resurfaces as the bottleneck.

Constraints: full proofs, no axioms/native_decide/sorry, Mathlib-free; kdecide rungs; both test suites unchanged; fixtures byte-identical; the compiled paths keep their csimp implementations untouched. Verification: the negative cactus ladder before/after; the ~12s regular-negative worst case is the target.

🤖 Prepared with Claude Code

kim-em:
This PR give the tactic's certificate obligations a replay over kernel-cheap list state. Microbenchmarks pinned the kernel's array cost precisely — 422us per Array getElem! access (the bounds machinery re-walks the spine after every mutation) against 81us for a bare-match list accessor — so the replay's call tree is mirrored over List Nat state with bodies identical to the originals, making every equality a structural induction through two unconditional bridges: ~30 proven clones from splitCellLoop through checkNodeL, the checkKeyLit surface over the tied flat literal, the tactic emitting it, and a kdecide rung proving the whole list-state closure kernel-reduces. The recorded same-machine re-time shows the negative ladder at 1.6-1.8x (worst 12.6 to 6.9s; total tactic curve 33.7 to 22.2s); compiled paths untouched and all other layers flat. Sorry- and axiom-free, fixtures byte-identical, both suites unchanged.

Closes https://github.com/kim-em/hex-dev/issues/9920.

🤖 Prepared with Claude Code


kim-em:
Before/after cactus comparison per SPEC § Benchmarks (refinement foundation → list-state kernel replay):

before/after cactus

|---|---|---|---|
| canonical hex fast tier | 0.99x | 12.5ms -> 12.4ms | 2.0ms -> 2.0ms |
| canonical hex checked tier | 0.99x | 64.0ms -> 63.4ms | 5.9ms -> 5.8ms |
| graph_iso tactic | **0.86x** | 33.69s -> 22.15s | 12.61s -> **6.88s** |
| pairs hex fast tier | 0.99x | 3.9ms -> 3.9ms | 529us -> 526us |
| pairs hex checked tier | 0.99x | 36.8ms -> 36.3ms | 20.0ms -> 19.7ms |

Negative-tail arc across the campaign: 25.6s (positive-route frame) -> 12.2s (literal ties) -> 6.9s (list state); the held SWAR machinery re-evaluates next now that arithmetic's share has doubled. Animation now fourteen frames.

PR #9939, opened 2026-09-02
feat: one code coordinate system for the graph-iso spec and transcription

### LINKED ISSUE #9927: graph-iso refinement, phase 0: one code coordinate system
Gating step of layer 3 per the scoping on https://github.com/kim-em/hex-dev/issues/9926: change the child-call convention in specNode, checkNode/checkChildren, certifyNode, certifyNodeAutom, and achieverCodes from numcells+1 to the recomputed true count (rs.numcells+1), so the spec's code coordinates coincide definitionally with the transcription's. Prove refine_seed en route (across numcells seeds, refine's lab/ptn evolution agrees and only longcode/numcells differ; refineLoop induction with a refineStep-is-identity-on-discrete sublemma). Then delete achieverCodes: the traced run's own code chain is checker-valid, so the trace carries codes directly — record best-path codes in TraceRun at best-update points and drop the post-run walk.

Non-negotiables: the transcription (Search.lean's run) is untouched; the fixture stream is byte-identical; certificate SIZES and record counts unchanged (key code values change — they are serialized nowhere); every soundness/completeness induction re-elaborates without weakening; the kdecide ladder recomputes its probe constants; both TacticTests suites unchanged. SPEC touch-ups: the nauty-compatible section's code-comparison wording if it fixes a convention, and a sentence at the requirement-1/requirement-2 boundary noting the coordinate unification (the declarative max-leaf-key characterization is unchanged).

Sequencing: do not start while the layer-2-bridges and kernel-access forks hold Translator/CertAutom/Cert additions; coordinate through the session owner.

🤖 Prepared with Claude Code

kim-em:
This PR execute phase 0 of SPEC § Verified search refinement: prove refine_seed (the code seed is a pure accumulator — a SeedShift congruence through the splitting machinery, carrying the guard-exit hypothesis that sound completion requires, since hypothesis-free seed-independence is false), flip the child-call convention across specNode, checkNode, searchNode, certifyNode, certifyNodeAutom, nodeSep, and the checkNodeL mirror to the recomputed true cell count so the spec's refinement codes coincide definitionally with the transcription's, and re-elaborate every downstream induction without weakening. With one coordinate system the transcription's own code chain is checker-valid, so achieverCodes is deleted: TraceRun carries the best-path codes and produceCand keys certificates on them directly — the checked tier gains ~5% from retiring that walk. Certificate key values change (they are serialized nowhere); certificate sizes, the fixture stream, and all observables are unchanged, verified twice (fixtures byte-identical, records 92/42/67, campaign 32,788 green). Probe literals and the manual chapter's worked key recomputed; the SPEC records the unification at the requirement-one/requirement-two boundary. Layer three of the refinement programme now lives in a single coordinate system.

Closes https://github.com/kim-em/hex-dev/issues/9927.

🤖 Prepared with Claude Code


kim-em:
Before/after cactus comparison per SPEC § Benchmarks (counting-sort activation → phase-0 coordinate unification):

before/after cactus

|---|---|---|---|
| canonical hex fast tier | 1.00x | 12.0ms -> 11.9ms | 2.0ms -> 2.0ms |
| canonical hex checked tier | 0.95x | 62.1ms -> 59.5ms | 5.7ms -> 5.6ms |
| graph_iso tactic | 1.02x | 22.70s -> 23.91s | 7.06s -> 7.36s |
| pairs hex fast tier | 0.99x | 3.8ms -> 3.7ms | 495us -> 490us |
| pairs hex checked tier | 0.94x | 36.0ms -> 34.7ms | 19.8ms -> 19.5ms |

Flat where a semantics-internal change must be, with the checked tier's ~5% dividend from the achieverCodes deletion. Animation now sixteen frames.

PR #9955, opened 2026-09-03
perf(number-field-tower): benchmark dense toPrimitive coverage

### LINKED ISSUE #9945: perf(number-field-tower): replace the sparse toPrimitive diagnostic with dense coverage
library: HexNumberFieldTower

## Context

PR #9940 optimized toPrimitiveWith to skip zero coordinates. The existing “full-basis” ladder applies the map to D unit vectors, so after that repair every individual call is sparse. Its near-cubic timing result therefore does not exercise the advertised dense public-operation bound and is inadmissible Phase-4 evidence.

The implementation of one genuinely dense call folds over D nonzero coefficients; each scalar/add operation acts on a degree-D primitive representation. The intended family-specific model is therefore quadratic in total tower dimension.

## Deliverables

- Replace the performance diagnostic with a prepared, bounded-height element whose D coordinates are all nonzero, and time one public toPrimitive call.
- Keep flattening/image construction and fixture preparation outside the timed body. Include structural result hashing and account for its lower-order linear traversal.
- Retain any useful sparse/unit-vector case only as a correctness or fixed anchor; do not cite it as dense complexity evidence.
- Commit the quadratic registration, parameter map, dimension schedule, and source-derived rationale before collecting the official measurements.
- Collect the required clean repeated export and inclusive profile. Update the NumberFieldTower SPEC/report/registration and remove the concern only if the pre-registered quadratic model passes the two-sided criterion.
- Run the relevant bench verification and full project checks required by the touched files.

If a genuinely dense family contradicts the quadratic derivation, stop and report the implementation-level cause. Do not choose a replacement exponent from the timings.

Parent: #9665
Evidence baseline: #9940

kim-em:
## Summary

- replace the sparse full-basis `toPrimitive` ladder with one prepared bounded-height, all-nonzero element and one timed public call plus structural hash
- preregister and preserve the quadratic rational-operation model before measurement
- record the clean repeated export and inclusive profile
- retain the concern after the official run contradicts the quadratic wall-time model, with the exact-rational primitive-image cost documented as the implementation-level cause

The official five-trial run was collected at clean commit `85f9c303f` on CPU 16 after `idle_core.py` certified it and SMT sibling 64 idle. It is inconclusive against `n²` (`β = +0.996107`, normalized interval `1067.20–3785.42`), independently reproducing the earlier CPU-1 residual of `+1.022529`. Per the directive, this PR does not fit or adopt a replacement exponent. The dimension-18 profile attributes 95.73% inclusively to `Flatten.toPrimitiveWith`, while the result checksum is only 1.89%.

## Verification

- `lake build HexNumberFieldTower HexNumberFieldTower.Conformance hexnumberfieldtower_emit_fixtures`
- `nix-shell -p 'python3.withPackages (ps: [ps.cypari2 ps.cysignals])' --run 'lake exe hexnumberfieldtower_bench verify'` (all 49 pass)
- `lake build` (10,568 jobs after rebase)
- `lake build hexnumberfieldtower_bench` (after review fixes)

Closes #9945


kim-em:
Independent Opus review completed while CI ran. It found the dense fixture, one-call timed boundary, structural hashing, preregistered quadratic model, schedule, result interpretation, and retained concern technically sound. I addressed every actionable note:

- restored the exact preregistered `n * n` spelling and clarified the preregistration commit message;
- corrected the fixed-anchor comment so it does not imply dense mode-1 coverage;
- removed the dense target from the obsolete sparse CPU-19 reproduction command and marked that artifact superseded;
- aligned the profile calibration prose with the 0.870 ms table value and explained that the rejected samples are untimed `flatten?` setup;
- replaced the initial fixed-CPU evidence with a clean five-trial export on CPU 16 after `idle_core.py` certified it and sibling 64 idle.

The new official residual is β = +0.996107, independently corroborating the earlier +1.022529 failure. No replacement exponent was fitted.

PR #9957, opened 2026-09-03
docs: clarify the splitter selection rule in the nauty chapter

kim-em:
This PR reword the splitter selection rule in the nauty-algorithm manual chapter. The old sentence stacked two "otherwise"s and left the wrap-around trigger implicit; the new wording states the rule as the forward scan it is (matching `pickSplit`): the first active cell starting at or after the preferred position, wrapping around to the first active cell in the partition when no active cell starts there or later.

🤖 Prepared with Claude Code


kim-em:
Closing to collect all manual review feedback into a single PR; the branch (graphiso-manual-splitter) keeps the wording fix and it will be folded into the combined manual PR.

PR #9979, opened 2026-09-03
refactor(graph-iso): prefer negative certificates

### LINKED ISSUE #9978: refactor(graph-iso): decide whether the negative tactic needs two kernel routes
The `graph_iso` tactic closes a negative goal by selecting between two kernel routes by measured cost: certificate replay (two Boolean `checkKey` checks plus `checkDiff`, cost proportional to the pruned certificate records the compiled search produces) and the fully verified pairwise decision `Pairwise.decideIso?` (cost proportional to the nodes its search visits). The pairwise decision is offered a node budget equivalent to the certificate replay's cost, and `HexGraphIso/Tactic.lean` claims it "wins exactly when refinement refutes the pair almost immediately". Independently of that race, the full-budget pairwise replay is the fallback when certificate production is unavailable, since exceeding the `maxCertNodes` per-side cap abandons the certificate route rather than the tactic, and it is described as the exhaustion-semantics anchor.

The suspicion to test is that the cost race is dead weight: that the certificate route wins whenever it is available, so the selection logic exists only to choose a fallback that a simple "certificate, else pairwise" rule would choose anyway. If so, deleting the race removes a moving part from the tactic, removes a per-call cost measurement, and makes the tactic's behaviour describable in one sentence, which matters because this tactic is about to be documented in a release announcement.

Investigate empirically before changing anything. Instrument route selection to record, per negative case, which route was chosen, the measured cost of each, whether the certificate route was available at all, and the realized kernel time. Run it over the tactic corpus in `bench/HexGraphIso/`, the negative pairs in the cactus corpus including the hard same-parameter families, the conformance fixture pairs, and a spread of random non-isomorphic pairs at n from 6 to 64 across densities, including pairs that differ only deep in refinement. Report how often the pairwise route wins when both are available, by how much, and on which shapes.

Then decide and implement one of:

1. The race never or almost never pays. Delete it: take the certificate route when it is available and fall back to pairwise otherwise. Keep the fallback, since it carries the exhaustion semantics and covers the `maxCertNodes` cap.
2. The race pays on an identifiable class. Keep it, but document that class concretely in the SPEC and the tactic docstring, replacing the current unquantified claim about refutation "almost immediately" with the measured characterization.
3. The race pays unpredictably. Keep it and say so honestly, with the data recorded as evidence.

Whichever branch is taken, update `HexGraphIso/SPEC/hex-graph-iso.md` and the `Tactic.lean` module docstring to match what the code does, and record the measurements as an evidence report under `reports/` following the convention of `reports/hex-graph-iso-campaign.md`. If the outcome changes tactic timings materially, the manual's Performance section and its quoted tactic costs need refreshing too, which interacts with https://github.com/kim-em/hex-dev/issues/9975.

Note for whoever picks this up: read the executable before trusting the prose. The SPEC and the tactic docstring currently agree with each other and with the code on the two-route design, but the claim about when the pairwise route wins is exactly the kind of unquantified assertion this issue exists to check.

🤖 Prepared with Claude Code

kim-em:
Closes #9978.

This PR removes the certificate/pairwise cost race from the `graph_iso` negative path. Over 66 non-isomorphic pairs spanning the tactic tests, the cactus families, conformance fixtures, a deep CFI pair and deterministic random pairs from n=6 to n=64, 14 reached the race with both routes available and the pairwise proof fit its allowance in none of them; the closest margin was 15% over, and forced-route kernel profiles favoured certificates by 1.47x, 3.36x and 6.55x. Negative goals now take the root separator, then certificates whenever available, then the two-code separator and the full-budget pairwise fallback, which keeps the exhaustion semantics unchanged.

The measurements are recorded in `reports/hex-graph-iso-negative-routes.md`, and the SPEC, both tactic docstrings and the manual chapter now describe the route the code takes.

`HexGraphIso/TacticTests.lean` gains a case that reaches the pairwise fallback. The two `maxCertNodes := 0` regressions never did: `sepRootG p3 k3` holds, so the coloured one closes at the root separator, and `sepDiffG` holds for Petersen/prism, so the uncoloured one closes at the two-code separator. Both separator legs are charged against `maxNodes`, so a budget in `[4, 2 * (n + 1))` withdraws the two-code separator while still funding `Pairwise.decideIso?`; the cubic Petersen/prism pair closes through it in 12 nodes at `(maxCertNodes := 0) (maxNodes := 21)`, and reports exhaustion rather than closing at `(maxNodes := 11)`.

The compiled canonicalization leg of the cactus sweep is re-measured here and reproduces, 98 shared rows moving by a median of -0.3%. The tactic leg is carried across byte-identical from the figure `main` publishes: it is one elaboration per case on a shared host and does not reproduce, with the same pair measuring 0.263 s and 0.543 s in sessions forty minutes apart on an unchanged executable. This PR therefore makes no end-to-end timing claim; the report records the spread, and notes that `main`'s own tactic figure and the manual's ten-vertex range have drifted apart independently of this change.

🤖 Prepared with Claude Code


kim-em:
Handoff status:

- CI passed completely for head `acdb3651f` (run 33766507476).
- The branch/worktree is clean and all work is pushed.
- A requested Claude Opus second opinion was attempted four times. The service returned two HTTP 529 Overloaded errors; two other clients stalled silently for 10–15 minutes and were terminated. No review findings were produced.
- `main` has since advanced six commits, so GitHub now reports conflicts. In particular #9987 (`refactor(graph-iso): collapse the checked tier onto the fast surface`) touches `HexGraphIso/Tactic.lean`, both graph-iso SPECs, the manual chapter, and benchmark artifacts; the next agent should rebase and revalidate the route evidence/wording against that refactor.
- Before the upstream divergence, the empirical result was 0 pairwise wins among 14 cases where both routes were available, across 66 negative pairs; forced-route kernel checks favored certificates by 1.47x–6.55x. The full report is `reports/hex-graph-iso-negative-routes.md`.
- After rebasing, regenerate graph-iso cactus data/figures if the relevant-source fingerprint changes, rerun CI, and retry Opus if the service is available.

Issue #9926, opened 2026-09-02
graph-iso refinement, layer 3: the traced key is the spec key


kim-em:
The mountain, now precisely stated by https://github.com/kim-em/hex-dev/issues/9924's execution: for every coloured graph, the achieverCodes-derived key of the traced run equals canonSpecKey — equivalently, the transcribed pruned search's winning leaf is the maximum of the unpruned individualization-refinement tree, including exact tie-breaking. Proving it simultaneously discharges the replay spine's domination hypothesis and, together with the layer-2 bridges, the colorSortedCheck transcription property, and the definitional rows equality, yields certifyCanon?_isSome — the SPEC's total surface, dead fallback arms, and the transport of the Checked theorem surface to the short names.

This directive is scoping-first: decompose before proving. The known ingredients: the pruning-soundness arguments (first-path/best-path code comparison, automorphism pruning, orbit pruning, short-prune) each need a preservation lemma against the imperative transcription's state (Search.lean's mutual quartet), or the argument routes through an equivalence between the transcription and a functional model that Complete.lean's machinery can reason about. Deliverable of the scoping pass: the lemma dependency graph with per-lemma effort estimates and the choice of proof route (direct imperative reasoning vs functional-model equivalence), as an issue comment — then the implementation directives follow from it. Treat all prior effort estimates as refutable; the layer-2 experience says the architecture may have made this smaller than it looks.

🤖 Prepared with Claude Code

kim-em:
Scoping complete; the headline reshapes the programme.

**Layer 3 as stated is not provable**: specNode/checkNode/achieverCodes pass children numcells+1 (input count) while the transcription passes rs.numcells+1 (true post-refine count), refine seeds its longcode with that argument, and mash is a hash — so sibling code order does not transfer between the two coordinate systems, and cross-system argmax agreement is empirically pinned (the SPEC's requirement-2) but not structurally true, with certifyCanon? = none into the exhaustive fallback as the designed safety valve. Verified mitigating structure: the seed affects only codes, never the explored tree (refine_seed, provable by refineLoop induction).

[...]

PR #9981, opened 2026-09-04
feat(graph-iso): the quartet statement layers and orbit completeness

kim-em:
This PR land the accumulated domination layers of https://github.com/kim-em/hex-dev/issues/9926. `QuartetStmt.lean` states the induction's conclusions over an optional incumbent, which is the shape the root instance needs since a bare key reading is degenerate there, and proves the root assembly: the target follows from the quartet's root instance alone. `QuartetLoop.lean` closes the mathematics of skipped-child domination, proving the completeness direction of the orbit closure that the earlier soundness work deliberately left out, together with the key equalities for a skipped child against an explored one along an orbit pointer or a chase, with no ordering assumption. `QuartetNode.lean` provides the reusable frame layer for the node steps, stated over arbitrary states so it composes in any order. Supporting layers: the first-path code machinery, the stabilization and ledger interfaces, and the descent lemmas the small-cell work consumes. Proof-only; fixtures byte-identical.

The induction itself is not built, and the statement it would use is known to need repair before it can be: the node conclusion's split on the returned level does not match the search, which returns the same level both on normal completion and when a child's early return abandons the sweep. That, and the rest of the remaining work, is written up for a fresh session in PLAN/VerifiedSearchRefinement.md.

🤖 Prepared with Claude Code


kim-em:
Superseded by #9985 (feat: prove certified canonicalization total), which landed the same commits as part of the full `issue-9982` history: every declaration in this PR's five Lean files is in `main` (the one private helper, `eq_zero_of_popCount_zero`, lives in `HexGraphIso/Nauty/PopCount.lean` as a public theorem), and `main` carries later revisions of each file. Closing without merging.

PR #10006, opened 2026-09-05
perf(graph-iso): make the certificate checker kernel-friendly

### LINKED ISSUE #9992: perf(graph-iso): make the certificate checker kernel-friendly
## Current state

`graph_iso` negatives are kernel-bound. Lean profiler on four pairs from the cactus corpus (chungus2, 2026-09-05): `neg-c6-vs-2c3` 0.41 s type checking, everything else under 20 ms; `neg-circulant10-2-5-vs-1-5` 1.05 s type checking; `neg-kneser72-vs-johnson72` (21 vertices) 8.56 s type checking plus 0.41 s interpretation. Positives are 1 ms of kernel work. Three corpus negatives (48- and 96-vertex circulant pairs, Paley 61 against a circulant) exceed the 120 s tactic timeout.

The graph definitions are not the cost: the 6-vertex negative with literal edge lists in place of `Families.cycle` and `Families.copies` takes 0.39 s, the same. The cost is certificate replay. Per certificate record (both sides' `certifyKey?` sizes summed) the kernel spends 11 ms at n = 6, 18 ms at n = 8, 26 ms at n = 10, 37 ms at n = 12, 47 ms at n = 14, 63 ms at n = 16 (cycle pairs, ~n^1.8), 40 ms at n = 10 (circulant pair), 103 ms at n = 21 (Kneser/Johnson, 83 records). Compiled `checkKey` on Kneser 7-2 is 256 µs for 42 records, 6 µs per record: the kernel is 10^4 times slower per record than compiled code, and the gap grows with n.

What the kernel evaluates per record (`Nauty.checkNode`, `HexGraphIso/Nauty/Cert.lean`): a full `refine` of the node's partition on `lab ptn : Array Nat` with `Nat` bitset rows, then `specMaketargetcell`, `breakout`, `discreteAt`, `leafRows` on leaves and `checkCellsPerm` on automorphism records. `Nat` arithmetic and bit operations are GMP-accelerated kernel extensions, but `Array` has no kernel support: `getElem!` and `set!` reduce through the underlying `List`, O(index) each, so a refinement that is O(n) accesses compiled becomes O(n^2) list walks in the kernel, matching the measured ~n^1.8 per record. On top of that the search side runs in the interpreter: `HexGraphIso` is not `precompileModules`, so `evalExpr` interprets `certifyKeyBounded?` (0.4 s at n = 21, roughly 25x compiled).

## Starting point

#9987 (refactor(graph-iso): collapse the checked tier onto the fast surface) has merged to `main`; the measurements above were taken on it. Start from `main`. #9988 (branch `issue-9975`) is manual-only and does not overlap this issue.

## Deliverables

Investigate first, then implement what the measurements justify; each of the following is a separable step with its own before/after on the four profiled pairs plus the three timeouts.

1. A kernel-cost harness: a script that builds the tactic files as `scripts/plots/hexgraphiso-cactus.py` does, runs `lake env lean -Dprofiler=true`, and reports type-checking time per certificate record against n, so every change below is measured the same way. Store its output under `reports/bench-results/`.
2. Representation of the replayed state (`lab`, `ptn`, the certificate's `γ` arrays, the adjacency rows), chosen per site by measurement rather than up front. Candidates, not mutually exclusive:
   - `Lean.RArray` (core, `Init/Data/RArray.lean`): a balanced decision tree "optimized for a fast kernel-reducible `get`", built for proof by reflection and used by `bv_decide` and Mathlib's `order` tactic. O(log n) `get` with almost no abstraction overhead; no `set`, so it suits read-mostly data (rows, `γ`, the per-node `lab` if each refinement produces a fresh one via `RArray.ofFn`/`ofArray`).
   - A `Nat` packed with fixed-width fields (`get i = (a >>> (w*i)) &&& mask`, `set` by mask-and-or): every access is one or two GMP-accelerated kernel operations on an O(n)-byte number. Suits `ptn`, which is written as well as read.
   - Index-free spellings: most of `refine` is a sequential sweep over cells, so `List` folds carrying the cell being built avoid random access altogether; keep random access only where the algorithm needs it (`lab[tcr.1 + o]`, `γ[w]`, the row of a vertex).
   - Sharing: bind each node's `refine` result once (`let`, already done) and check that the kernel's whnf cache actually hits across the child sweep; `Nat.ble`/`Nat.beq` on literals in place of `compare` on `Ordering` where it changes the trace.
   Whatever is chosen stays a proven-equal fast spelling of the current `Array` definitions, the way `Refine.lean` pairs `nontrivialCellFast` with `nontrivialCell`, so the soundness proofs keep mentioning the array spelling. Measure each candidate on the harness before committing to it; the answer may differ for rows, `lab`, `ptn` and `γ`.
3. Less recomputation per record: examine whether the certificate can carry what the kernel now recomputes, with the kernel verifying instead of computing: the refinement's cell-split sequence (verified by `Nat.land` of rows against cell masks and a count check), the target cell, and the leaf's rows. The certificate grows, but verification of a split is linear in the words touched while recomputation is quadratic in the kernel. Report the record-count and kernel-time trade-off on the corpus; keep only what wins.
4. `precompileModules := true` for `HexGraphIso` (or the tactic's `evalExpr` calls routed through a precompiled module), so the compiled search runs compiled inside the tactic. Measure the interpretation category before and after.
5. Structural kernel costs to check and remove where they matter: `Id.run do` loops that reduce to bind chains, `match` on `Ordering` results, `decide` instances synthesised per obligation, the fuel `100` and `initPtn n (n + 2)` setup, and the `Bool` matrix literal round trip (`matrixListSide`) that the SPEC records as the shared floor.

## Context

The tactic must stay certificate-sized and kernel-checked; nothing here may trust compiled code (`native_decide` is banned, see SPEC). The `graph_iso` section of `HexGraphIso/SPEC/hex-graph-iso.md` describes the certificate and pairwise routes and their budgets; `HexGraphIso/ModuleBoundaryTests.lean` pins the reduction closure the kernel needs (it must stay green as definitions move). The uncoloured chapter examples (Petersen versus the prism) are the user-facing cases to keep fast.

## Verification

`lake build` green including `HexReleaseTests`, `HexGraphIsoProofProbe` and `HexGraphIso/ModuleBoundaryTests.lean`; conformance fixtures byte-identical; the kernel-cost harness shows the per-record cost and its exponent before and after each step; the cactus tactic curve regenerated with the sweep (`scripts/bench/graphiso_cactus_sweep.sh`) and the three timeouts re-examined; no `sorry`, `axiom`, `native_decide`.

🤖 Prepared with Claude Code

kim-em:
This PR make the `graph_iso` certificate checker kernel-friendly. Add `scripts/bench/graphiso_kernel_cost.py`, which profiles each negative corpus pair and reports the kernel's type-checking time per certificate record with its exponent in the vertex count (records under `reports/bench-results/hexgraphiso-kernel-*.json`, a `graph_iso` trace class supplies route and record counts). Replay certificates over packed kernel state: the labelling, partition and adjacency rows are fixed-width fields packed into one `Nat` each (`HexGraphIso/Nauty/Packed.lean`, with the `Rep` correspondence to lists), every step is spelled with the `Nat` functions the kernel accelerates, `popCount` and `lowBit` read byte tables, and counted loops run through one `Nat.rec` step per iteration. `HexGraphIso/NodePacked.lean` clones the replay over that state and proves each clone equal to the list clone, so `checkKeyP` closes the negative route through the existing `checkKey` soundness with the rows tied as one packed literal (`packRowsK`); `HexGraphIso/SeparatorPacked.lean` gives the root and two-code separators the same replay. Build `HexGraphIso` with `precompileModules` and time with `lake lean`, which loads the shared objects, so the compiled search runs compiled inside the tactic.

Measured on chungus2 against the list-state replay (`reports/hex-graph-iso-performance.md` has the tables): kernel time on the cycle ladder and the Kneser pair down 2.5 to 5x depending on host load (Kneser 9.0 s to 1.6 s in the best run and 3.5 s as recorded, of which 0.4 s per side is the family's own adjacency evaluation), per-record cost from `0.50 * n^1.80` ms to `0.42 * n^1.54` ms at n <= 21 as recorded (best of five runs `0.23 * n^1.53`), interpretation from 0.41 s to 0.01 s, and the 48- and 61-vertex timeouts now close in 60 to 76 s. The 96-vertex pair remains out of budget (377 s of kernel time with `maxRecDepth` raised). Carrying the refinement's split sequence in the certificate was examined and not adopted: the neighbour counts a verifier must compute dominate a packed pass. The SPEC's tactic section records the packed obligations. The released mirror's `lakefile.toml` does not yet set `precompileModules = true`, so the manifest flag is left off for now.

Closes #9992.

🤖 Prepared with Claude Code


kim-em:
Rebased onto main three times as main's graph-iso files moved; the sweep and kernel record are regenerated at each fingerprint. Note: main's own CI is currently red on the factor-sweep freshness step (https://github.com/kim-em/hex-dev/actions, commits 9a462147b and efa5e0c49 fail with `hex-factor: source differs from hexbz-factor-sweep-1cbe1674-chungus2.json` in HexBareiss/Bareiss.lean, HexBerlekampZassenhaus/PrimeSelection.lean, HexMatrix/Basic.lean, HexPolyFp/NttMul.lean); this PR's merge run inherits that failure until main is repaired.

PR #10014, opened 2026-09-05
perf(graph-iso): replace the Nat bitset with packed words above 63 vertices

### LINKED ISSUE #9990: perf(graph-iso): replace the Nat bitset with packed words above 63 vertices
## Current state

`canonicalize` runs nauty's search tree node for node (conformance pins the node counts), so the hex/nauty ratio X is a per-node constant factor. The sweep recorded at fingerprint `152d306db788` (chungus2, 2026-09-05, `reports/bench-results/`) shows X is flat at 5 to 9 up to 64 vertices and then grows: per family, X ~ n^0.3 to n^0.8 beyond that, hex per-node cost ~ n^2.1 to n^2.7 against nauty's n^1.0 to n^1.9. Overall X is 6.4 at n ≤ 96, 20.6 at n > 96, 62 on Kneser 22-2 (231 vertices). Paley goes from 56 µs per node at n = 61 to 207 µs at n = 73.

The cause is the vertex-set representation. `HexGraphIso/Nauty/Bits.lean` models a vertex set as a `Nat` bitset. Below 2^63 that is an unboxed scalar and every set operation is one machine word; above it every `land`, `shiftRight`, `% 256`, `testBit` allocates a GMP bignum. A `perf` profile of `hexgraphiso_cactus` (whole corpus, hex and nauty in one binary) attributes 76% of the hex time to GMP and the allocator (`__gmpz_init_set`, `lean_nat_big_shiftr`, `lean_nat_big_mod`, `__gmpn_divrem_1`, `malloc`/`free`), 14% to the Lean search code itself and 10% to the Lean runtime; nauty's own symbols are 0.7% of the run. The byte-chunked loops `popCountGo`, `toListByteGo`, `lowBitGo` (a `% 256` and `>>> 8` per byte, each a bignum op on an n-bit set) and the refinement helpers `cellEnd`, `splitCellLoop`, `nontrivialCellFast`, `trivialCell`, `ntcPass` are the visible callers. nauty keeps sets as `setword` arrays of `m = ceil(n/64)` words and loops over words with no allocation.

The Bits.lean docstring already anticipates this: "The packed-word representation may replace this one later only with a proof that no observable result changes."

## Starting point

#9987 (refactor(graph-iso): collapse the checked tier onto the fast surface) and #9988 (docs(graph-iso): refresh the manual's timings after the tier collapse) have both merged to `main`: the `Checked` tier is gone, the cactus data has no `checked_ns` column, the sweep at fingerprint `152d306db788` on which the measurements above were taken is recorded, and the manual's Performance section is the one deliverable 4 extends. Start from `main`.

## Deliverables

1. A packed-word vertex-set type (e.g. `Array UInt64` or `ByteArray`, `m` words, vertex 0 at the most significant bit of word 0 as in nauty) with `lowBit`, `popCount`, `nextElem`, `toList`, `rowCmp`, `permset` and the set algebra `refine` needs, each proved equal to the current `Nat` model (`Nat.testBit`-level correspondence), so that every theorem stated on the `Nat` model transports. Keep the `Nat` model as the specification the proofs mention.
2. `rowsOf`, `Ctx.g`, `RefineSt` and the refinement/search code in `Refine.lean` and `Search.lean` on the packed type, with the fast paths (`popCountFast`, `lowBitFast`, `toListFast`, `refineTrivialFast`, `nontrivialCellFast`) rewritten as word loops. Observable results (labels, forms, node counts) unchanged: the conformance fixtures stay byte-identical and the campaign agrees with nauty.
3. Side-by-side profiling as a repeatable script (`scripts/bench/graphiso_perf_side_by_side.sh` or similar): `perf record` of `hexgraphiso_cactus`, a report split into hex-Lean / GMP+allocator / Lean runtime / nauty-C categories, and the per-family log-log fit of per-node cost against n from the sweep data. Land it with a `reports/` note giving the before/after numbers.
4. The asymptotic evidence, so this issue also answers "does X stop growing":
   - `scripts/bench/graphiso_pernode_fit.py`: reads a recorded sweep, computes per-node cost for hex and nauty per family (Kneser and Johnson have node counts that grow with n, so normalise), fits log-log exponents, and prints the table plus X in the n ≤ 64 and n > 64 slices. Run it from the sweep script and store the output next to the sweep data.
   - A required check beside `check_graphiso_sweep_freshness.py`, in the existing single CI job: on every family with at least five sizes the hex per-node exponent may exceed nauty's by at most a stated margin (0.2 to start). It fails loudly if a later change reintroduces an elementwise loop.
   - The manual's Performance section gains the model in a few sentences (same tree, per-node factor, exponents per family, X in the two slices, machine and date), and the SPEC's benchmarks section names the check.
   - If 255 vertices (the nauty comparator's bound) is too small to see the asymptote after the change, extend the largest families to the bound and note the series break.
5. Cactus sweep refresh and the before/after comparison comment per the SPEC's benchmarks section.

## Context

The kernel-facing checker (`Cert.lean`, `CanonForm.lean`) also consumes `Nat` rows; leave its representation alone in this issue (the kernel has GMP-accelerated `Nat` ops and no fast arrays), and make the boundary between the compiled search's packed sets and the checker's `Nat` rows an explicit conversion with a proof. The tactic's kernel cost is a separate issue.

Run the fit on the current sweep first (fingerprint `152d306db788`) so the "before" row exists. Target: X flat in n. nauty's per-node cost is about n^1.85 on the dense families; the packed representation should give the same exponent, so X should return to the 5 to 9 range seen below 64 vertices at every size in the corpus. Anything above n^2 per node after this change points at a remaining elementwise loop.

## Verification

`lake build` green including `bench/` and `conformance/`; conformance fixtures byte-identical (`hexgraphiso_emit_fixtures` diffed against `conformance-fixtures/HexGraphIso/graphiso.jsonl`); campaign run against nauty; `scripts/bench/graphiso_cactus_sweep.sh` refreshed and `check_graphiso_sweep_freshness.py` green; the per-family fit shows hex per-node exponents within 0.2 of nauty's; no `sorry`, `axiom`, `native_decide`.

🤖 Prepared with Claude Code

kim-em:
This PR replaces the `Nat` bitset the nauty-compatible search kept per vertex set with `Nauty.VSet`, an array of 63-bit limbs (each an unboxed `Nat` below `2^63`), so every set operation costs one word operation per 63 vertices instead of a GMP bignum call above 63 vertices. `rowsOf`, `Ctx`, `RefineSt`, `SearchSt` and the whole refinement and search tree run on the packed type, the vertex count is an index of every search type, and the `Nat` bitset survives only as the kernel-facing specification: `NodeLit`'s literal tower consumes packed rows through `Ctx.toL` and `KeyL` with a correspondence proof per operation. The conformance fixtures are byte-identical and the 32,798-case campaign agrees with nauty 2.9.3. The measurement side adds `scripts/bench/graphiso_perf_side_by_side.sh` (a `perf` category table of the cactus binary) and `scripts/bench/graphiso_pernode_fit.py` (per-family fit of per-node cost against `n`), runs the fit from the sweep script, makes `--check 0.2` a required check in the single CI job beside the sweep freshness check, names it in the SPEC's benchmarks section, adds the per-node model to the manual's Performance section, and records the before/after profile and fit in `reports/hex-graph-iso-packed-sets.md`. `HexBasic.OfFn` gains `Hex.Array.zipWith'`, the kernel-reducible twin of core `Array.zipWith` in the mould of `Array.map'`, `Families.choose` becomes multiplicative because the Pascal recurrence was the largest symbol in the cactus profile, and the cactus driver materializes each instance's adjacency before timing so the hex column no longer pays the family generators that nauty never paid. The packed kernel checker that landed on `main` meanwhile (`NodePacked`, `Nauty/Packed`, `SeparatorPacked`) is ported onto the guarded single-`Nat` operations and `KeyL`, so the tactic's packed routes run on the packed-set search unchanged.

Per-node exponents: hex went from `n^0.3` to `n^0.8` above nauty per family to at or below nauty on all nine families. The hex/nauty ratio is 7.7 below 64 vertices and 7.1 above (before: 5.8 and 18.9). Instances above 64 vertices canonicalize 2.6 times faster (geometric mean); instances up to 63 vertices run at 0.74 times the old speed, because a one-limb array allocates where an unboxed `Nat` did not. That trade-off is recorded in the report rather than hidden behind a small-graph twin.

Closes #9990

🤖 Prepared with Claude Code


kim-em:
Before/after comparison per the SPEC's benchmarks section: sweep `d78dade3633a` (last on `main`, which already carries the packed kernel checker of #10006) against `ac1dfad4a453` (this branch at 6390dc3f616d3305b567cf18afcad180a19f5245, after merging `main`), both on chungus2, same corpus.

before/after cactus

|---|---|---|---|
| canonical hex canonicalize | 0.79x | 869.2ms -> 242.3ms | 211.4ms -> 38.0ms |
| graph_iso tactic | 1.03x | 140.01s -> 133.60s | 73.61s -> 63.97s |
| pairs hex canonicalize | 1.20x | 33.3ms -> 30.4ms | 13.8ms -> 8.3ms |

The canonical row's 0.79x geometric mean is the small-graph slowdown (0.74x up to 63 vertices) averaged against the large-graph speedup (2.6x above 64: Kneser 22-2 goes from 211 ms to 38 ms, Paley 229 from 29 ms to 6.7 ms), the trade-off the description states. The tactic row is unchanged within noise; its kernel checker keeps `Nat` rows converted at the boundary.

Per-node fit (`scripts/bench/graphiso_pernode_fit.py`), hex exponent minus nauty's, before and after:

| family | before | after |
|---|---|---|
| circulant-12 | +0.32 | -0.08 |
| circulant-1248 | +0.54 | -0.08 |
| grid | +0.38 | -0.18 |
| hypercube | +0.29 | -0.05 |
| johnson | +0.62 | -0.06 |
| kneser | +0.69 | -0.00 |
| latin | +0.78 | -0.09 |
| paley | +0.69 | -0.10 |
| random | +0.68 | -0.03 |

Overall X (hex/nauty, geometric mean): n ≤ 64 5.75 → 7.67, n > 64 18.85 → 7.08. Full tables in `reports/hex-graph-iso-packed-sets.md`.

PR #10089, opened 2026-09-07
docs: specify exact lattice enumeration and minima

kim-em:
Specify exact Fincke-Pohst enumeration with Schnorr-Euchner coefficient ordering, all shortest nonzero vectors, and all closest vectors to a rational target. Rectangular independent integer bases are supported, including the target component orthogonal to their span.

The design gives exact integer interval formulas, complete and budgeted result contracts, checked LLL basis transformations, and exhaustive certificates that establish minimality and all ties. The Mathlib companion interprets the lattice as an integer span and derives packing radius and kissing number. Conformance, benchmark families, and the proved-mode fplll comparison are specified.

Register the two libraries as planned and replace the future-work sketch with SPEC links. This PR is independent of the rational-function and permutation-group proposals and adds no Lean implementation.

Validation:
- Relative links, Markdown fences and `git diff --check`
- Exact interval equivalence and Schnorr-Euchner ordering checked on 4,000 rational cases
- Exact search recurrence compared with independent coefficient-box enumeration on 60 ball queries, including rectangular bases, off-span targets, negative radii and rank zero
- A2 example returns six shortest vectors of squared norm 2

The proposed Lean declarations and theorem statements have not been elaborated.

### DIFF
--- a/SPEC/Libraries/README.md
+++ b/SPEC/Libraries/README.md
@@ -47,6 +47,7 @@
 - **hex-berlekamp**: Berlekamp factoring, distinct-degree and equal-degree factorization (Cantor-Zassenhaus), and the Rabin irreducibility test over any `F_q`; the `factor_poly` / `irreducibility` tactic drivers (native `FpPoly p` arms plus extensions for other input types)
 - **hex-hensel**: Hensel lifting from `mod p` to `mod p^k`
 - **hex-lll**: LLL lattice basis reduction
+- **hex-lattice-enum** (planned): exact closed-ball enumeration, all shortest and closest vectors, and checked completeness certificates
 - **hex-berlekamp-zassenhaus**: complete factoring of `Z[x]`; the `Hex.ZPoly` extension for `factor_poly` / `irreducibility`
 - **hex-summation**: certificate-checked hypergeometric summation: Gosper, Zeilberger, and Petkovšek's Hyper, as untrusted searches whose certificates are verified by `MvPoly` identity checkers
 - **hex-conway**: Conway polynomial database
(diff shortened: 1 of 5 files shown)


kim-em:
Implementation premise check: the specified proof placement needs a prerequisite that is absent from the computational dependencies at `817c0d1cd159ac9dab0c04e1f5256db69899ea3e`.

The exact Gram–Schmidt section requires `HexLatticeEnum` to use `Hex.GramSchmidt.Int.data` and prove its preparation correct using the existing integer correspondence theorems, while remaining Mathlib-free. Those theorems are not all available on that side of the dependency boundary:

- [`gramDetVec_eq_gramDet`](https://github.com/kim-em/hex-dev/blob/817c0d1cd159ac9dab0c04e1f5256db69899ea3e/HexGramSchmidt/Int/Combination.lean#L1872) requires `hquot : StepWitness b`. The [`StepWitness` documentation](https://github.com/kim-em/hex-dev/blob/817c0d1cd159ac9dab0c04e1f5256db69899ea3e/HexGramSchmidt/Int/Canonical.lean#L920) explicitly says that concrete providers are constructed in the Mathlib companion and the computational layer only consumes the abstraction.
- The general constructor [`StepWitness.ofGram`](https://github.com/kim-em/hex-dev/blob/817c0d1cd159ac9dab0c04e1f5256db69899ea3e/HexGramSchmidtMathlib/Int/Augmented.lean#L855) lives in `HexGramSchmidtMathlib` and uses the augmented Bareiss/Desnanot determinant proofs.
- The required norm-ratio theorem [`basis_normSq`](https://github.com/kim-em/hex-dev/blob/817c0d1cd159ac9dab0c04e1f5256db69899ea3e/HexGramSchmidtMathlib/Int/GramDet.lean#L1034) and scaled-coefficient correspondence [`scaledCoeffs_eq`](https://github.com/kim-em/hex-dev/blob/817c0d1cd159ac9dab0c04e1f5256db69899ea3e/HexGramSchmidtMathlib/Int/Augmented.lean#L1156) also live exclusively in that Mathlib layer.

The existing Mathlib-free orthogonality and row-reconstruction lemmas concern the semantic basis; they do not supply these missing identifications for `data`. In particular, adding an executable triangular recurrence alone does not close the proof obligation. Importing the current correspondence proofs would violate the specified dependency boundary; adding a `StepWitness` input would change the required public input contract.

This boundary is deliberate: [`LLLState.ofBasis_valid`](https://github.com/kim-em/hex-dev/blob/817c0d1cd159ac9dab0c04e1f5256db69899ea3e/HexLLLMathlib/State.lean#L24) documents the same reason for living in Mathlib. The [Bareiss proof-placement policy](https://github.com/kim-em/hex-dev/blob/817c0d1cd159ac9dab0c04e1f5256db69899ea3e/HexBareiss/SPEC/hex-bareiss.md#L314) explicitly requires relocating the consumer's bridge theorem or redesigning its proof surface, rather than reproving the determinant infrastructure in the computational layer.

There is a computable rational alternative: the semantic wrappers call executable `GramSchmidt.basisMatrix` and `coeffMatrix`. However, using those kernels would depart from the mandated `data` preparation, and still would not supply the missing Mathlib-free implication from the current determinant-based `Matrix.independent` predicate to positive orthogonal norms.

This is an infrastructure/placement gap, not a counterexample to Fincke–Pohst enumeration or the exact interval formula. The SPEC needs an explicit boundary revision: for example, expose search/checker correctness over verified preparation data in the computational library and prove automatic preparation from the current `Basis` input in the Mathlib companion. Checking preparation at runtime alone does not solve the unconditional contract: one must still prove that preparation succeeds for every accepted independent input. That decision affects the headline theorem placement and must be reflected in the SPEC, rather than silently adding hypotheses or an extra failure case.

Validation: `lake exe cache get` and `lake build HexLLL.Lattice` succeeded (251 build jobs). A temporary Lean import probe built with `lake build` confirmed that the three declarations above are absent from the `HexLLL.Lattice` / `HexGramSchmidt` environment and printed the `StepWitness` argument on `gramDetVec_eq_gramDet`. The probe was removed afterward; the tracked worktree is unchanged.

Following the repository's premise-check rule, I have stopped before adding a partial implementation, weakening theorem statements, or introducing proof placeholders. No decomposition issue has been filed.

Issue #10082, opened 2026-09-07
feat(mv-factor): complete verified multivariate integer factorization and its Mathlib API


kim-em:
library: HexMvFactor

## Current state

The executable integer multivariate factorization stack exists, but the verified, Mathlib-facing library specified by the project is incomplete. This issue tracks completion of that existing scope, not a new factoring algorithm. Keep it open until the acceptance criteria below are met; coherent implementation PRs can land incrementally.

Source audit at main commit `8dd0f8e155ebd9c49cb08728dde3051fb8fec4a5`:

- `HexMvFactor` implements structural reductions, squarefree/content recursion, evaluation-point and leading-coefficient search, EEZ lifting/recombination, bounded partial results, Kronecker fallback, and `factor?` / `complete?`. The implementation landed in #9442; Kronecker replay was strengthened in #9567.
- Conformance drivers, committed fixtures, SymPy oracles, and `HexMvFactorizationTests` exist. All three computational libraries have `done_through: 3` in `libraries.yml`; that records conformance completion, not proof completion.
- There are **36 literal `sorry` occurrences** in their Lean sources: **18 in HexMvGcd, 17 in HexMvHensel, 1 in HexMvFactor**. These are not confined to auxiliary results. `HexMvFactor/Irred.lean:85` leaves `checkIrred_sound` unproved, and `checkComplete_sound` depends on it. GCD gaps include certificate soundness, Gauss/content/PRS results and eight squarefree contracts. Hensel gaps include shift identities, univariate and multivariate correction, lifting, uniqueness/completeness, and irreducibility transfer through specialization.
- `checkComplete` alone does **not** establish irreducibility. Its soundness theorem correctly requires the univariate `obligations` separately. The producer-provenance theorem discharging those obligations is specified but absent.
- `HexMvGcdMathlib`, `HexMvHenselMathlib`, and `HexMvFactorMathlib` are absent from source and the library registry. Consequently the specified end-to-end Mathlib theorem, uniqueness/decidability API, and multivariate tactic extensions are missing, despite README text describing them.
- `bench/HexMvHensel/` and `bench/HexMvFactor/` are absent. GCD benchmarks and a performance report exist, but its phase counter is still 3. None of the three libraries is in `scripts/release/released.yml`.

[...]

PR #10091, opened 2026-09-07
fix(mv-gcd): replay rational certificates with canonical arithmetic

kim-em:
A caller-assembled rational-lift certificate could capture unlawful source `GcdOps`, causing the checker to accept noncoprime integer inputs. The counterexample on #10082 uses `2x + 2` and `2x + 4`, an identity embedding, and source operations that call every coefficient a unit.

Fix the certificate boundary by fixing the coefficient ring throughout recursive evidence and replacing the generic lift with canonical integer models, rational scales, and ordinary integer certificates. Rational representations carry proved ring-isomorphism laws; integer replay uses canonical operations and cannot contain nested lift leaves. The producer extracts ordinary evidence before building the rational wrapper. Update the SPEC, README, and replay probes for this API.

Validation:
- Kernel regressions reject the reported source certificate, malicious local dictionaries, repeated primitive factors, invalid scales, invalid recursive evidence, and nested optional leaves; valid rational lifts succeed. Guarded axiom audits contain no `sorryAx`.
- GCD, Hensel, factorization, the multivariate test target, and kernel replay probes build.
- Fresh fixtures match all committed fixtures; SymPy checks pass for all 86 cases.
- All 19 GCD benchmark smoke checks pass, including FLINT/Singular comparators.
- Whole-graph build (10,744 jobs), including `HexManual`, passes on the head rebased onto #10079.
- DAG, file-size, copyright, phase 4/7, release-manifest checks, and `git diff --check` pass.

This is a prerequisite repair for #10082. Existing checker/Gauss/PRS/squarefree proof gaps remain; no axioms or `sorry`s are added and no phase counters advance. The tracker stays open.

### DIFF
--- a/HexMvFactor/SPEC/hex-mv-factor.md
+++ b/HexMvFactor/SPEC/hex-mv-factor.md
@@ -1401,7 +1401,10 @@ with what those call: `MvPoly` multiplication, powering, and equality,
 `imageAt`, `lcIn`, `toUnivariate`, `constIn`, `MvPoly.eval`,
 `polyNormalize`, `scalarContent`, hex-mv-gcd's `checkContent`,
 `DensePoly` content, primitive part, degree and equality, and `List`
-length and indexing. Each is `@[expose]`.
+length and indexing. Each is `@[expose]`. The shared GCD content checker
+also includes canonical rational-leaf replay in its transitive dependency
+closure, although `RatModel.not_int` rules out that leaf for these integer
+coefficients; see hex-mv-gcd's kernel-exposure contract.
 
 Nothing in the pipeline is in that closure: point search, prime search,
 `witnessOf?`, the lift, and the recombination enumeration are search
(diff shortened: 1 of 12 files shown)


kim-em:
Independent Claude Opus review found no remaining way for caller-assembled certificates to substitute arithmetic outside the checker hypotheses. It also confirmed that fixing the ring parameter closes the analogous operation-dictionary mismatch in `split`.

Follow-ups implemented in 81a826384:
- Proved `checkOps_eq` to tie public replay to the generic core.
- Added direct extraction/replay probes for both modular and Bézout splits, plus the counterexample's missing axiom audit.
- Removed duplicate raw image evaluation; producers and replay now use the law-bearing `imageAt`.
- Clarified the generic leaf callback's separate soundness obligation, the representation-mapping step, and the complete GCD/factorization replay dependency closure.
- Added the direct field import and documented why integer extraction cannot encounter a rational leaf.

The compatibility projection names remain to preserve the existing public API. Converting all integer producers and consumers to the leaf-free types would be a broader API change; the present shared integer specialization is safe because `RatModel.not_int` excludes rational leaves. Its syntactic dependency closure is now explicit in both SPECs.

The revised head passes the whole-graph build, the complete multivariate stack and kernel probes, all 86 oracle cases, all 19 GCD benchmark smoke checks, and the structural/phase/release checks. No axioms or `sorry`s were added. CI is rerunning on the revised head.

Issue #9619, opened 2026-08-25
perf(int-factor): close divisor-enumeration evidence gap and recertify the pair


kim-em:
libraries: HexIntFactor, HexIntFactorMathlib

`Hex.Nat.divisors` is implemented and proved, but has no accepted performance-evidence registration. Close that specific gap and reconcile the pair's readiness status. This is benchmark/measurement work with a final integration audit, not a new factorization implementation.

## Verified starting point

- Both libraries currently have `done_through: 7` in `libraries.yml`. That counter does not discharge the missing Phase-4 evidence. Core/bridge implementation, conformance and local Phase-5–7 work already landed in #9882 and #9887; primality prerequisites are complete (#9767, followed by #9938). Do not restart those completed projects.
- [`HexIntFactor/Divisors.lean`](https://github.com/kim-em/hex-dev/blob/main/HexIntFactor/Divisors.lean) implements public `divisors` as generated divisor products followed by `List.mergeSort` and array conversion. Its [SPEC](https://github.com/kim-em/hex-dev/blob/main/HexIntFactor/SPEC/hex-int-factor.md) retains the `O(τ log τ)` contract, where `τ = ∏(eᵢ+1)`. Existing sigma, square-divisor and kernel-replay measurements do not measure this public operation.
- `bench/HexIntFactor/Bench.lean` has no accepted `runDivisors` registration. The [bridge SPEC](https://github.com/kim-em/hex-dev/blob/main/HexIntFactorMathlib/SPEC/hex-int-factor-mathlib.md) transports divisor enumeration and names the core as its computational performance owner.
- [`reports/hex-int-factor-performance.md`](https://github.com/kim-em/hex-dev/blob/main/reports/hex-int-factor-performance.md) still places expected informational-comparator differences in `Concerns`. Preserve that evidence but move it to the comparator trend narrative, as required by the benchmarking SPEC. Actual unresolved evidence gaps remain Concerns until resolved.
- The pair is not in the release manifest. Publication is outside this issue.

## Prior experiment: retain as diagnostic evidence

Preregistration commit `8b54275eb34be820b3aa416306e58d9fb0fe41a5` used squarefree products of the first 6, 8, 10, 12, 14 and 15 table primes: `τ = 64, 256, 1024, 4096, 16384, 32768`. Subjects and divisors fit below `2^63`, avoiding a changing big-integer-width cost. The declared model was `τ * log₂ τ`.

[...]

PR #10099, opened 2026-09-07
perf(int-factor): preserve divisor evidence and record blocked acceptance

kim-em:
Public `Hex.Nat.divisors` had no operation-specific performance evidence, and unsuccessful collection deleted the raw export. Preserve every collector outcome, register the public operation on the controlled squarefree family, validate complete outputs outside timing, and retain an inclusive profile with reproducible artifacts.

This is a partial resolution of #9619. Both preregistered timing attempts returned consistent slopes but failed the independent contamination gate (4.8737% and 5.1226%, versus 0.2%). The two-attempt budget is exhausted. Both attempts remain rejected, the original contrary experiment is documented, and both libraries are rolled back to Phase 3. The remaining dependency is an admissible designated-host measurement campaign with a newly committed protocol. #9619 remains open after this partial change.

The profile attributes 76.58% to sorting and 16.33% to divisor generation. It passes the existing calibration and sensitivity checks after independently matching all 2,981 imported timestamps to raw perf timestamps and correcting their common clock origin. Reprocessing has a complete hashed provenance record; ELF symbol sizes bound attribution, and the unresolved leaf share is 0.03%. Raw captures, failed calibration, normalized profile, symbols, commands, hashes, and diagnostics are retained. Expected external comparator differences move to the trend narrative; the actual evidence blocker stays in Concerns. The core/bridge API mapping is explicit. Library implementations and proofs are unchanged.

Validation:
- `lake build HexIntFactor HexIntFactorMathlib HexConformance HexManual hexintfactor_emit_fixtures` passed (11,031 jobs).
- `lake exe hexintfactor_bench verify`: all 41 cases pass.
- Fresh IntFactor fixtures match exactly; PARI checks all 415 cases with no failures.
- 24 collector/profile tests pass, including inconclusive exports, malformed JSON, subprocess failures, timeouts, complete-output validation, and clock-normalization rejection.
- Phase 4, DAG, Phase 7, performance freshness, and Mathlib-free bench checks pass.

Existing local Phase-5–7 library source and documentation are unchanged from their accepted join and build successfully; recertification remains blocked by Phase 4. No release manifest or publication changes are included.


kim-em:
The requested independent Opus review is complete. I verified its substantive findings and updated the implementation:

- GMP-ECM now uses subprocess-only elapsed time, excluding provenance hashing and evidence writes.
- Rejected divisor collection ingests available export/telemetry and records validation before propagating the original failure. Separate hashed rechecks document validation of both immutable rejected attempts.
- Profile reprocessing now runs end to end from the retained `perf.data.gz`, with input/output/source hashes, commands, exits and profiler-checkout state. The clock correction preserves every relative timestamp and corrects the common wall-clock origin.
- ELF symbol bounds leave gaps unresolved; the bounded reprocessing preserves the headline shares and reports 0.03% unresolved leaf samples.
- Validation pins the one-rung verdict trim and trial indices. The report explicitly limits the verdict range to τ=256..32768 and retains τ=64 as diagnostic data.
- Added gap/clock/model/failure tests, archived integration logs, and documented the gates made dormant by rolling the pair back to Phase 3. All 24 tests and local integration checks pass.

Two suggested changes were deliberately limited: raw failure logs remain complete rather than truncated, and the Lean driver's default input remains available for the harness's `verify` calls at parameters 0 and 1. The profile runner has an explicit registered-parameter guard. Neither rejected measurement was promoted, and no new timing campaign ran.

PR #10092, opened 2026-09-07
feat: implement canonical rational functions and Mathlib correspondence

kim-em:
Implements the libraries specified in #10087 through Phase 3.

Includes canonical normalization and uniqueness, cancellation-first field arithmetic, partial evaluation at poles, normalization certificates, derivatives, polynomial-part decomposition, and Mathlib RatFunc correspondence with canonical numerator/denominator agreement. Adds compiled manual examples, QQ/GF(2)/GF(7) conformance fixtures, SymPy and FLINT checks, and benchmark anchors.

Validation against current main:
- Combined library, conformance, benchmark, fixture-emitter and manual build: 11,126 jobs passed.
- Fresh fixtures match byte-for-byte; 1,169 SymPy and 242 FLINT full-result comparisons pass.
- All 35 frozen benchmark hashes pass.
- Structural checks, kernel axiom audits, and performance-artifact freshness checks pass.
- No sorries or new axioms in the new libraries.

CI and a fresh Opus review are required before merge. Phase 4 remains explicitly open: operation-specific scaling evidence, profiles, and larger matched-input FLINT measurements will follow separately. The Mathlib companion documents the bundled Field instance compatibility requirement.

PR #10100, opened 2026-09-07
perf(rational-fn): complete Phase 4 benchmarks and kernel replay

### LINKED ISSUE #10096: HexRationalFn: late-mismatch benchmark changes the first visited coefficient
library: HexRationalFn

## Current state

Phase 4 is in progress; done_through remains 3. The query anchor and initial scientific different registration change the highest coefficient, although the specification asks for a difference at the last compared coefficient.

## Symptom

Declared expectation: different n => n, based on visiting all n coefficients before finding a mismatch.

Observed: the clean e12e88614 sweep on chungus2 CPU 3, degrees 128,256,512,1024,2048,4096,8192,16384, reports inconclusive, beta approximately -0.999, with approximately 37 ns at every rung. Raw samples are retained in reports/bench-results/hex-rational-fn-query-e12e8861-chungus2-cpu3.json on the Phase 4 branch.

Root cause: RationalFn equality calls Array.instDecidableEqImpl on the numerator. Array.isEqvAux checks index i before recursing to i-1: highest coefficients are compared first, and the constant coefficient is last. Adding monomial n 1 creates an immediate mismatch, not the advertised full traversal. This is a benchmark fixture bug, not a rational-function arithmetic defect.

## Deliverables

- Keep done_through at 3; no rollback PR is needed because Phase 4 was never claimed and the finding invalidates no completed correctness phase.
- Change the mismatch to the constant coefficient without changing the independently derived linear model or schedule.
- Assert equal lengths, equal nonconstant coefficients, and distinct constant coefficients before timing.
- Retain the failed sweep, rerun the scientific ladder, and report both the diagnosis and corrected evidence in the performance report.

## Verification

lake build hexrationalfn_bench; sizes fixture validation; verify; scientific run at the unchanged schedule, with the full raw export retained.

### LINKED ISSUE #10097: HexRationalFn: complete timed-region profiles for fixed workload families
library: HexRationalFn

## Current state

The library remains at done_through 3. Phase 4 adds scientific query and witness-size replay ladders, but normalization, addition, multiplication, coefficient-height, and calculus currently have fixed latency anchors only.

## Symptom

SPEC/profiling.md requires a timed-region profile per input family, excluding preparation and result hashing. The pinned LeanBench.Profile.runProfile resolves only parametric registrations; runFixedChildMode does not emit a timed-region sidecar. The fixed anchor families therefore cannot use the mandated filter today. Upstream feature request: https://github.com/kim-em/lean-bench/issues/73.

## Deliverables

- Keep done_through at 3 while profile and scientific operation coverage remain incomplete; no completed correctness phase is invalidated.
- Once fixed profiling is supported, or genuine operation-specific parametric registrations replace the anchors, collect representative filtered profiles for the five remaining families.
- Attribute the high calculus latency with a profile before making an algorithm change; do not infer a characteristic-zero derivative optimization for all coefficient fields.
- Record filter diagnostics, source/host/toolchain provenance and dominant inclusive/leaf costs in reports/hex-rational-fn-performance.md.

## Verification

Every retained profile passes the filter confidence and sensitivity checks. No fake parameter, false complexity model, or whole-process profile substitutes for the missing fixed timed-region support.

### LINKED ISSUE #10105: Unbalanced Karatsuba block accumulation performs quadratic full-output additions
Phase 4 rational-function ladders at commit 791f17ca expose near-quadratic scaling for products with one fixed linear factor (multiply: 0.925 ms at degree 32, 7.76 s at 4096), although blockwise multiplication should be linear in the long operand. A sampled diagnostic at degree 1024 attributes 97.26% inclusive cost to Karatsuba.Raw.blocksFrom and 87.21% to Raw.addShift. The accumulator constructs a fresh array across the entire growing output on every short block, performing rational additions even outside the block support. This is shared HexPolyFast infrastructure, not coefficient growth in the rational-function fixture. Fix the accumulation loop to update only the shifted block support, with an extensional proof and regression evidence for short-by-long products. The failed scientific export will be retained on PR #10100; the model must not be weakened to quadratic.

### LINKED ISSUE #10106: Shared-denominator scaling fixture couples degree to rational coefficient height
The Phase 4 addShared fixture at 791f17ca uses 1/(h X) + 1/(h (X+1)). Its second gcd is gcd(h, 2X+1); evaluation at -1/2 creates rational denominators of order 2^degree(h). Therefore the adjacent bounded-coefficient linear model is not sound for this input family, independently of the observed slope. Replace the degree-isolated control with 1/(h X) - 2/(h (X+1)), whose second gcd is against 1-X and whose partial sums stay machine-word-sized over the declared ladder. Keep the failed original export and record rational-height growth in the separate height family. This is a fixture-premise correction, not a fitted complexity change.

kim-em:
## Summary

Complete Phase 4 for the rational-function library implemented in #10092 / #10087. Both `HexRationalFn` and its correspondence-only Mathlib bridge advance to `done_through: 4`.

- Cover every advertised compiled operation with 40 scientific registrations: 35 passing two-sided models and five documented passing one-sided upper bounds. Keep 35 fixed full-output hash/comparison anchors separate from operation coverage.
- Fix three exposed performance defects with proofs: avoid rescaling monic inverses; dispatch short polynomial products/divisions directly; accumulate general unbalanced Karatsuba blocks in one owned output array instead of repeatedly copying growing suffixes.
- Exercise degree, cancellation, long/short ratio, coefficient height, output size, long Euclidean chains, and certificate witness size independently. Include odd-sized blocks, genuine second-gcd addition cancellation, and total cancellation.
- Record full internal comparisons, 270 matched FLINT results and representation sizes, and eleven calibrated operation-only profiles covering every input family and the worst comparator gap.
- Add literal certificate proof probes and a release-quality six-round fresh-module sweep with import/null controls, artifact sizes, axiom sets, resource counters, and strict shared-host interference admission. All four replay/rejection probes stay below 1.08 seconds against their preregistered five-second budgets.
- Complete the five-section headline report with all raw evidence and no unresolved Concerns. Extend the existing CI/build checks; add no workflow or top-level job.

### Performance rationale

The two-sided degree families deliberately keep coefficient arithmetic machine-word-sized and control quotient-chain length or operand shape; their linear or Karatsuba scaling is family-specific, not a generic rational bit-complexity claim. The long finite-field Euclidean chain uses the published O(M(n) log n) upper bound, and fixed-degree coefficient-height families use the published quadratic arithmetic upper bound because GMP's data-dependent paths do not justify a tight wallclock model. The raw harness's faster-than-bound `inconclusive` wording is translated only as the documented one-sided rule allows. FLINT is informational, with different integer-pair representation and specialized packed-integer multiplication; its large/diverging balanced-product gaps are reported, not hidden or treated as same-class wins. Kernel replay uses separate absolute fresh-module budgets, never compiled timing or a synthetic combined verdict.

### Validation

- Full `lake build`: **10762 jobs passed**, including aggregate, manual, and Mathlib bridge.
- Affected computational, Mathlib-test, conformance, and fixture-emitter targets pass, including the new arbitrary-fuel/block-size allocation guards.
- `hexrationalfn_bench list`, `verify`, `sizes`: **75 smoke checks passed**, frozen hashes and independent fixture invariants intact.
- Scientific runs: **40 passing registrations, 960 successful signal-eligible rows**, unchanged two-second target / three trials / slope and noise policies.
- All **12 internal comparison groups** agree on complete-output hashes, including the eight-rung unbalanced ratio comparison.
- **1169** native conformance fixtures agree with SymPy; **120** polynomial-backend fixtures agree with the FLINT/exact-Python oracle; **270** matched performance fixtures agree with FLINT.
- **11** kernel-region profiles pass calibration/sample/sensitivity checks and classify at least 98% of leaf samples.
- Fresh kernel sweep: **complete, release_quality=true, no validity exceptions**, all budgets passed; only standard `[propext, Classical.choice, Quot.sound]` dependencies.
- Phase-4, DAG, Mathlib-free bench/probe, copyright, source-size, factor/graph performance-freshness checks and conversion/DAG unit tests pass.

The LeanBench operation-only profiling support was independently fixed and merged upstream in kim-em/lean-bench#75; this PR pins that revision. No axioms, `native_decide`, or new sorries.

All rational-function headline measurements use clean source `83d220225762a783750c110949b37fb80479ba90`; subsequent commits change only tests, documentation, and evidence. All 63 fresh-proof source fingerprints match. The accumulator uses a tighter preallocation and a list-free tail-recursive update loop; FLINT conversion and checkout-provenance tests run in the existing CI step. Eight affected polynomial-backend scaling registrations and all 64 backend smoke checks also pass. Independent Opus review and the focused follow-up review are complete; substantive feedback is addressed (see the review record below).

Closes #10096. Closes #10097. Closes #10105. Closes #10106.


kim-em:
Independent Opus review is complete, including a fresh focused review of the accumulator follow-up. Opus found the algorithm, capacity argument, unconditional semantic proofs, division dispatch, models, and kernel protocol sound. Addressed the substantive feedback: tighter storage, list-free updates, empty-tail guard, exact-capacity regressions, conformance coverage documentation, scoped quotient-prefix cost statement, conversion/provenance tests in CI, and refreshed affected-backend evidence. Final rational-function evidence uses clean 83d220225: all 40 models, 960 eligible rows, 12 comparison groups, 270 exact FLINT cases, 11 qualified profiles, and all fresh-proof budgets pass. The subsequent test/documentation changes preserve all 63 measured proof-import hashes. All 64 polynomial-backend smoke checks and eight focused backend scaling checks also pass. Review differences: the empty-tail issue also existed in the old implementation (fixed nevertheless); untracked files intentionally count as dirty, and informational FLINT exports do not themselves grant release-quality status. No further runtime change was made for the cosmetic cutoff-name suggestion.

PR #10104, opened 2026-09-07
perf(int-factor): fix telemetry ownership and retry controlled collection

kim-em:
The divisor telemetry monitor could count its own newly forked benchmark children as foreign work because it scanned ancestry before scanning threads. Read ownership from the same task stat record as state/CPU, in a dedicated session. Preserve partial telemetry on errors and interruption, clean up descendants on every exit, and preserve runner exit codes. PID handles and bounded rescanning cover grandchildren and live worker threads whose main thread has exited. Collector timeouts allow a short evidence-writing grace period before forcibly stopping the group.

Preregister a new finite campaign at `36e2a3d1fa8fe9e1f49067246c58f98aaba8d4a1`: CPU 62 / SMT sibling 14, unchanged two-sided τ log₂ τ model, seven trials per rung and 0.2% interference ceiling. All 150 two-second preflight windows failed, so no new acceptance timings were collected and no retry followed. Archive every observation, source/binary hash, command/log and the host dependency. The report explains the coarse gates: this preflight effectively required zero busy ticks, and the timed foreign-task score generally requires zero sightings. Further collection needs a new resolution-aware preregistration with justified host controls. A further untimed 30-second survey of 47 physical-core pairs found none under the same ceiling; the best maximum sibling activity was about 0.33%, with 0.033% tick resolution. Its generator and raw counters are retained. No subsequent timing campaign was registered. Prior rejected and contrary evidence stays intact.

#9619 remains open after this partial change. Both libraries remain at `done_through: 3`; publication is outside this PR. The current benchmark binary exactly matches the existing inclusive profile.

Validation: full `lake build HexIntFactor HexIntFactorMathlib HexConformance HexManual`, all 41 bench verification cases, 41 affected collector/telemetry/profile tests, matching fresh fixtures, 415 PARI oracle cases, phase/DAG/freshness checks and the Mathlib-free scan (43 executables, 191 proof probes) pass. Logs are retained in `reports/bench-results/intfactor-campaign-2-integration.json` and its adjacent directory, with final infrastructure checks in `intfactor-campaign-2-cleanup-validation-4.json`.


kim-em:
Claude Opus reviewed the implementation. The ownership field indices, fork/thread inheritance, collector timeout group, retained artifacts and honest Phase-3 status checked out.

Addressed findings: monitor errors and explicit interruption now persist partial telemetry and terminate the entire group; nonzero runner exits also clean up grandchildren. Added subprocess regressions for standalone and collector launch modes, recorded owned runnable tasks, handled missing timing files in `--audit-attempt`, guarded/restored affinity, and corrected the report's quoted duration. All 40 tests pass; the first cleanup-test run exposed an ESRCH race in the test observer, which is fixed with the failed log retained.

The report now explicitly describes resolution: the two-second/100-Hz preflight effectively requires zero ticks, and the timed foreign-task score effectively requires zero sightings. I agree these are coarse and poorly matched to the survey's busy core; I do not agree that zero-tick windows are impossible by construction. No criterion was loosened or applied retroactively. The exhausted campaign remains rejected, and any further campaign requires a new resolution-aware preregistration with justified host controls. No new timing was taken.

CI is running on the updated head, with a focused follow-up review of the cleanup changes in progress.

kim-em:
The focused follow-up review's must-fix points are addressed:

- Cleanup requires a dedicated session; an existing pipeline group is rejected before spawning the runner. The collector already launches a dedicated session.
- Ordinary cleanup now preserves runner exit codes and runs on successful exits too. It uses PID handles and bounded rescanning, including for a zombie main thread with live worker threads. The monitor stays in the collector's group so forced timeout cancellation still reaches the whole tree; moving the benchmark into a separate group would have weakened that boundary.
- Telemetry writes are atomic, repeated interruption cannot truncate the rescue write, and timeout cleanup allows five seconds for evidence writing before a final group kill. `WNOWAIT` retains the leader PID until that kill. Observer affinity is recorded explicitly, and missing exports now produce a clear diagnostic.
- The report scopes the test count to the affected collector/telemetry/profile modules.

All 41 affected tests pass, including actual subprocess regressions for pipeline launch rejection, ordinary failure, interruption, leaked descendants on success, live worker threads after leader exit, and timeout output preservation. The rejected campaign and its thresholds remain unchanged; no further timing was taken. CI is running on the final update.

PR #10107, opened 2026-09-07
feat(graph-iso): prove and adopt the structured search

### LINKED ISSUE #10043: feat(graph-iso): prove the structured search by one policy-parameterized recursion (Fable)
libraries: HexGraphIso, HexGraphIsoMathlib
depends-on: #10027
depends-on: #10041
depends-on: #10035

Design note: `reports/hex-graph-iso-rewrite.md` (until merged, https://github.com/kim-em/hex-dev/pull/10026) (the phases, layout,
deletion lists and decision criteria). Read it first; this issue is one
step of that plan.

model: fable

Kept `blocked` on the release issue and on the spike; a Fable session
claims it by number after removing the label, once the spike has passed.
If the spike failed, this issue runs as Phase C: the literal port is
proved equal to the generic recursion at the nauty policy by unfolding,
then the same soundness theorem.

## Goal

Target: `canonSpecKey_eq_tracedKey` restated with `runColoredTraced` and
`tracedKey` (`CertReplay.lean:1060`) re-pointed at the engine, so
`certifyCanon?_isSome_of_keyEq`, `CertTotal`, `CanonForm`, and
`Ops.canonicalize_eq_certifyCanon` are untouched.

### B1. Architecture: one recursion, one policy

The ladder `Nauty/Model/` cannot be the target: its prunes are per-node,
while nauty's code 1 to 4 returns abort sweeps at several levels at once,
`(fix, mcr)` pairs prune far from the admitting leaf, and the engine
descends dominated subtrees when `eqlev_first == level` to harvest
generators. A rung with non-local exits has the engine's recursion shape,
so the two designs converge:

- `Generic.node`/`Generic.sweep`: the engine's recursion over a
  `Policy σ n` typeclass (`visit`, `descend?`, `leaf`, `skipChild?`,
  `afterChild`, ...). The engine is `Generic.node (σ := Search n)` at the
  nauty instance; write the engine directly first (the spike), then
  refactor and re-run the sweep to confirm the compiler specialises the
  typeclass away (`trace.compiler.ir.result`). If it regresses, keep the
  direct engine and prove `Engine.node = Generic.node` by unfolding.
- A policy that never prunes reduces `Generic.node` to `specNode`:
  `generic_trivial_eq_specNode`, with sufficient node and sweep fuel.
  Its state retains the labelling, partition, path codes, and incumbent.
  Only additional pruning bookkeeping is trivial. Exhausted generic
  calls stop the traversal, so equality with the specification's
  truncated whole-tree fold is not asserted for insufficient fuel.
- Generic soundness, in the shape of `searchNode_eq` and `searchNodeG_eq`:
  under `SoundPolicy`, `best' = some (incMax best (prefixKey cs
  (specNode ..)))`, where `best : Option (Key n)` is `none` before
  the first leaf. Carry the semantic incumbent through the temporary
  `compCanon = 1` overwrite window, and read it from the executable
  state when the comparison is stable. The invariant is preserved,
  and `unwind t` below the
  level carries `Witness t σ'` (the current child subtree at level `t+1`
  is bounded by the incumbent), which intermediate loops transport and the
  loop at `t` turns into ordinary coverage.
- `SoundPolicy` obligations are local per decision and reuse existing
  lemmas: `skipChild?` via `childKey_of_orbPruned`, `longprune_carried`,
  `shortprune_carried`, `childKey_of_carried`; code prune via
  `codeInv_keyCmp_lt`, `specNode_keyLe_of_code_lt`; leaf verdicts via
  `tied_full_keyCmp`, `leafEvent_faithful`, `updatecan_inv`; unwinds via
  `cellStab_of_scatter`, `orbjoin_orbSound`, the frozen machine for
  `eqlevCanon`, and the small-cell subtree for the cheap return
  (the subtree starts at `noncheaplevel`, not at the preceding
  failed-guard level; reuse `CheapDesc` and the small-cell theorems);
  store validity for the restored admission test requires new run-history
  bookkeeping: retain `SubtreeOk` at the `gcaFirst` ancestor, construct
  `DescPath`s from that frame to the first and current leaves, and prove
  their target-position paths equal. The existing `CheapDesc` only
  describes the current node and `Guide.refReach` only records cell
  permutation, so neither supplies this history. With these facts,
  `descPath_leafRows_all`, `leafRows_eq_of_descPaths`, and
  `checkAutom_scatter_of_descPaths` justify the cheap admission.
  Code 2 uses `checkAutom_scatter_of_leafRows_eq`; invariant maintenance
  via `CodeCmpInv`, `FirstCodeInv`, `recover_machines`,
  `otherNodePrep_frames`, `firstterminal_*`, `AutosOk`/`PairOk`,
  `OrbSound`, `refine_equitable`. Keeping `recover`, `firstterminal`,
  `compareCodes`, `pushAuto` as the same functions on a record with the
  same field names makes those lemmas transfer by renaming.
- The invariant is one record of about ten clauses (sizes and reach,
  `CodeCmpInv`, `FirstCodeInv`, `CanongInv`,
  `GenTraceOk`, `AutosOk`, `OrbSound`, `gcaFirst/gcaCanon < level`,
  `noncheaplevel ≤ level`, `cosetindex < n`) instead of 68 types, because
  unwinding and sweep coverage live in the theorem's conclusion.
  Do not require every stored generator to stabilize the current
  partition: this fails on reachable states, already for the empty graph
  on four vertices. Generator validity is global. Cell stabilization is
  required for the particular carrier or composed orbit witness at the
  frozen ancestor partition where it is consumed. The invariant and
  soundness contracts must retain the ancestor frame and prior-child
  coverage needed to construct and interpret `Witness`; these facts
  cannot be inferred from comparison counters or automorphism validity
  alone. Use `RunInv.otherLeaf`, `Guide`, `Anchor.ofCellCarrier`, and
  `Unwind.Located` as the reference for those obligations. The clause
  count is an estimate, not a requirement to omit necessary hypotheses.
- Reach facts consumed by `CertTotal` (`canonlab_cellsReach`,
  `runColoredTraced_result`) re-proved on the generic recursion as
  policy-independent lemmas from `refine_reachAt`/`breakout_reachAt`.

### B3. Size, risks, deletions

New: generic recursion plus trivial instance 0.5k; generic soundness 3-4k;
nauty policy invariant and transitions 2-3k; decision justifications
2-3k; reach lemmas 1k. Total 9-12k. Deleted on completion:
`Nauty/Correct/` (27.8k), `Invariant/Incumbent`, `Coverage`,
`TraceAgree`, most of `Invariant/Reach`, the `SearchSt`-specific half of
`Domination`, the literal `Search.lean` (the engine takes its place and
name, keeping the correspondence table). Net about 20k lines fewer, plus
5.5k from B2.

Risks in order: the cheap-return witness (code equality across a cheap
subtree; confirm the reusable statement early from `Correct/Exit`
and `SmallCell/Transitive`); typeclass specialisation of the mutual recursion
(mitigated by direct-then-refactor); `Exit.fuel` unreachability (small,
the level-versus-fuel bound); reach re-proof (mechanical).

Phase C fallback (spike fails): same generic recursion and policy, with
the literal port proved equal to `Generic.node` at the nauty policy by
unfolding, then the same soundness theorem.

## Acceptance

- `canonSpecKey_eq_tracedKey`, `certifyCanon?_isSome`, and the public
  theorems of `Ops.lean` hold with the engine as the search.
- `Nauty/Correct/` and the literal `Search.lean` are deleted; the engine
  takes the `Search` name and keeps its correspondence table.
- The oracle, the campaign twin (now trivially one engine) and the sweep
  are unchanged from the spike's measurements.
- No `sorry`, no `axiom`, no `native_decide`.

## Constraints every PR must respect

- Any non-comment change under `HexGraphIso/`, `HexGraph/`,
  `bench/HexGraphIso/Cactus.lean` or `scripts/plots/hexgraphiso-cactus.py`
  requires `scripts/bench/graphiso_cactus_sweep.sh` on chungus2 and the
  data, manifest and figures committed with it
  (`scripts/bench/check_graphiso_sweep_freshness.py`, `ci.yml:103`). A file
  move is a delete plus an add, so move PRs need a sweep; batch them.
- `HexManual/Chapters/NautyAlgorithm.lean` cites about 40 `Nauty` names by
  Verso role and `HexManual/Chapters/HexGraphIso.lean` cites
  `Pairwise.search`; `lake build HexManual` is a required check on every
  rename or deletion.
- `scripts/release/released.yml` lists `HexGraphIso.TacticTests` as the
  test module; the manifest must follow any test-target change.
- Style: `SPEC/writing-style.md` and `.claude/CLAUDE.md` (short verb-noun
  names, no process narration, no `axiom`, no `native_decide`).
- A peer session `hex-graph-iso-perf-fix` is active; any change it lands
  under `HexGraphIso/` moves the sweep baseline and must be rebased onto.

## Out of scope

Further performance work beyond the spike's engine.

kim-em:
The structured engine now drives the public canonicalization, certification, generator and orbit APIs. Its complete key and generator-completeness theorems follow from actual generic-recursion calls and proved local rules, without legacy traversal correctness assumptions. Empty-graph certification and the public theorem statements are preserved.

The engine takes `Nauty/Search/Search.lean` and retains its nauty correspondence table. This removes the literal traversal, `Nauty/Correct/`, the old incumbent/closure proof, and recursive reach proof. Reusable primitive-state and reference/stabilizer mathematics remain in `Search/State`, `Invariant`, and `Generation`. The direct hot loop retains its proved equality with the policy recursion. Canonicalization does not import the full generation proof.

Validation:

- Both libraries, registered tests, conformance and `HexManual`: successful build after rebasing onto current `main`.
- All 6,233 fixture records unchanged byte for byte; nauty agrees on those and the 32,798-case campaign. Full wrapper twin passes. Negative oracle controls reject altered rows, labels and node counts.
- Axiom guards cover engine and public correctness; dependency audit finds no legacy traversal proof. No added `sorry`, `axiom`, or `native_decide`.
- Chungus2 cactus data, manifest, figures and tactic timings committed under `4fa1958f160d`; freshness and the existing 0.2 exponent check pass.
- All 98 benchmark node counts match the spike. Current/archive engine-time geometric mean is 1.0012 (worst instance 1.0243), comparing one current run with the archived three-run median.
- A temporary newest-32-pair filter changes a normal-root traversal from 168 to 712 nodes, preserving key and 11 generators with local proof changes. This demonstrates locality parity with the legacy proof, not an exclusive advantage or speedup. Production retains the original full scan.

The dependency checker now includes `import all` edges in build reachability, with a regression test; sealed-import restrictions are unchanged. The library umbrella also includes the standalone trace proof, and the policy entry point includes the first-root proof.

Optimization scope: changes inside policy operations have local obligations; changes to recursion shape must maintain direct/generic agreement and may require new return and generation arguments. The primitive `SearchSt` view remains a maintenance cost.

Proof structure and validation details: `reports/hex-graph-iso-obligations.md`; locality evidence: `reports/hex-graph-iso-locality.md`.

Closes #10043.


kim-em:
The independent Opus review found no correctness defect or vacuous assumption in the completed engine/generic equivalence, checked-admission, reachability, exhaustive-policy, or local leaf-maximum proofs. Its medium findings concern the still-missing whole-search coverage argument.

Implemented follow-up:

- `pruneReturn_target` proves the arithmetic split between a target at or above `eqlevCanon` and the cheap-boundary target.
- `Comparison.prune_witness` and `Comparison.prune_result` give actual code-prune exits a `Generic.Result` carrying semantic ancestor coverage. The witness uses `cs.take (t + 1)`: the review's sample `cs.take t` loses the first unequal code when `t = eqlevCanon`, so that proposed statement was not used.
- `SubtreeOk.node_key` and `SubtreeOk.path_key` prove that any complete selected descent in a small-cell subtree realizes its entire specification key, including codes. The underlying child-key theorem is now outside `Correct/`.
- Removed the redundant positive-size premise from `Trivial.Complete`, documented its immediate-parent stack discipline, clarified that generic sweep entries may represent vertices or offsets, and removed the dummy empty graph context in the depth proof.

The review's proposed stack redesign is unnecessary for the exhaustive instance: `recover_child` and `node_eq` establish the current balance and immediate-parent return. The generic module contains the executable recursion, so keeping the direct engine's reducible `Exit`/`Leaf` aliases there does not introduce a dependency on proof modules. The matching engine/legacy result names remain namespace-qualified during the migration; their public replacement and deletion must be atomic once maximum soundness is available.

Still unresolved: the full pruning invariant (`AutosOk`/`PairOk`, orbit witnesses, and frozen ancestor frames), binding path codes to those frames through the actual recursion, and composing all cheap and automorphism returns into whole-search coverage. Checked automorphism validity alone does not establish the ancestor stabilization needed by the pair ledger. Consequently the public search switch and legacy deletions are not ready, and this PR remains a draft rather than claiming #10043 is complete.

kim-em:
Follow-up to the independent Opus review:

- Extracted `FixedCells`, `LocalAutos`, `PathStab`, and the reusable `PathOk` transitions into `Invariant/PathStab.lean`. The legacy root initializer remains with the literal search. `NodePre`, `SweepPre`, and `FirstPre` now carry current-partition stabilization; the initial state and both actual call paths establish it. `SweepPre.local_pairs` combines this path evidence with the root ledger at the consuming partition.
- Proved `node_fixed` through `Generic.CallPolicy`, including first-path and off-path calls, nonlocal exits, and fuel exhaustion. Target-cell singleton evidence proves the chosen vertex fresh, and child cleanup restores the exact parent fixed set before recovery. This supplies the missing premise of the path-stabilization return transition.
- Added `WorkspaceOk` to the persistent and first-entry invariants. Explicit/implicit admissions preserve positive capacity and the array bound, and the actual first descent establishes the invariant.
- Added `Covers.keysMax`, `Result.node`, and witness implication transport. The saved first-reference history now records its actual code array contents; `FirstRef.code_eq` connects those contents to the semantic comparison codes, and `pathKey_codes` identifies the key's real-code prefix.
- The claim that colour stabilization had no consumer was incorrect: `PairsOk.admit` consumes it for every explicit admission. A separate public colour/`IsIso` bridge was optional, and main now supplies its own `AutTrace` bridge through #10108. I did not reorder `CheapOk`'s existing array arguments: the explicit instantiations are correct, and that stylistic change is not needed for correctness.

The fixed ancestor code binding, cheap/automorphism unwind coverage, short-prune fix-test receipt, and whole-engine maximum remain unfinished. Witness implication transport alone does not establish the missing frozen-frame binding. The public search migration and legacy deletion are still gated on those proofs. The new public generation-completeness dependency from #10108 is documented separately on #10043; it also needs a transfer that canonical-key equality does not supply.

kim-em:
Follow-up to the independent Opus comparison review:

- The full-window carrier lemma is now named `SweepPre.window_carriers`. `FilterCover.lean` instead proves long-filter coverage for an arbitrary mutable live subset using `ChildCover.filterDesc`, the actual pair-drop theorem, and vertex-indexed specification keys. It carries earlier coverage through incumbent growth, cursor advancement, and parent recovery. The short-filter counterpart retains the receiving fix-test premise explicitly; integrating that producer and these lemmas into the whole maximum contract remains unfinished.
- `recover_filters` proves both filters read the same workspace before and after recovery. `pruneReturn_short` and `leafExit_cheap_short` derive implicit admission from the actual short flag. `Generic.ShortPolicy` instantiates the common recursion to transport leaf-emission properties. Its nauty instance supplies `node_origin` and `sweep_origin`; `LeafReturn.admission` identifies the newest pair at a returned state using capacity supplied there.
- Factored the executable post-child control into `Generic.advance` and `Generic.resume`. Reach, fuel, fixed-point, safety, comparison, and short-return contracts now state their local theorems over those operations. The direct-engine equality still builds.
- Moved `Comparison.prepare` into its sole consumer, `CodeCalls.lean`, and removed the comparison import from `Prepared.lean`; pared the safety imports accordingly.
- I disagree that discarding the explicit `Grows` result loses the first-leaf lower bound: `ReturnCodes.lower` already states `keyLe (incKey ctx fs st.firstlab) (incKey ctx bs st.canonlab)` and survives `firstPath_codes` and the real root initializer. No redundant induction was added for that bound.

The remaining coverage contract will retain frozen ancestor information separately from the settled comparison receipt. The draft remains incomplete and is not ready to merge.

kim-em:
Addressed the fifth Opus review's concrete findings:

- Short-return admission now uses the receiving state's fields and proves the return target is at most `noncheaplevel - 1` for implicit pairs, or equals `gcaCanon` for explicit pairs.
- The implicit fix test is proved for an actual returned child, using its partition effect, restored fixed points, and frozen singleton cells. It does not assume the conclusion as a filter precondition.
- Removed unused workspace wrappers, renamed the new coverage relation to `CellCover`, aligned short-return binder order with the executable operations, and added cursor exhaustion and representative-skip rules.
- Added canonical-ancestor bounds to actual node and sweep preconditions. `CanonOut` now tracks reference installations and both ancestor bounds through the common recursion; `canon_old` identifies the retained parent reference on an upward return, and `canon_guide`/`canon_locate` connect it to covered ancestor children.

The review found no correctness defect in the original short-return origin or mutable coverage proofs. Its remaining integration concerns are still applicable: the explicit pair's receiving fix test, actual first-path orbit-skip justification, and the whole-engine maximum are unfinished. The new guide restoration consumes child coverage as the intended induction hypothesis; it does not itself establish whole-search coverage. The PR remains a draft.

kim-em:
Addressed the canonical-reference review in the short-filter batch:

- The generic sweep now proves its unconditional strict unwind bound through `SoundPolicy`. Both actual node branches have strict bounds, and `child_target` identifies a received child target with the parent level.
- Canonical-reference transport and pair validity now have partition-based entry points that apply before any sibling has returned. `child_canon_guide` includes both first-child and fixed-point cleanup; `first_return_pair` proves validity after the actual leftmost child's cleanup without a later-sibling precondition.
- Short-return provenance retains the actual classifier. `LeafReturn.admission` transports its scatter relation and recorded generator; `return_pair` derives local validity for both explicit and implicit pairs. `SweepPre.return_cover` applies that result to the actual short filter. The obsolete, weaker `short_admission` wrapper is removed.
- `CanonGuide.mem` proves membership in the original target window, which can contain vertices removed from the mutable target set.
- Shared canonical-control lemmas replace the duplicate admission, prune, and recovery proofs. `FixedCells.ofSearchOut` delegates to `ofEffect`. `Result` imports the new proof layer without publicly re-exporting it.
- All new proof modules are committed and reached by the library build. The small-cell shape is also carried by the actual node and sweep preconditions, including first-child recovery.

The dummy target-level argument remains in the policy-independent child proof, consistent with the other local-policy instantiations. No executable behavior was changed. The whole-engine maximum, nonlocal coverage witnesses, first-path orbit coverage, generation completeness transfer, public migration, and deletions remain unfinished; the PR remains a draft.

PR #10120, opened 2026-09-08
Prepare the Mathlib-free polynomial gcd-domain lift

kim-em:
The proof-only polynomial gcd-domain lift was using its single `sorry` for all four class fields, and the SPEC's intended arity induction lacked a Mathlib-free multiplicative recursive view. This slice proves the divisibility, nontriviality, and zero-product fields directly, adds `toUnivariate_mul` at the first variable, and leaves only `gcd_exists` inside the existing gap.

It also makes the executable `LawfulGcdOps` domain proof independent of the remaining gcd-existence gap. Exporting an existing `HexPoly` convolution proof changes its source fingerprint but no executable definition, so the PR records the exact proof-only blob transition required by the benchmark freshness guard. This is an incremental prerequisite for #10082; it does not close the tracker or reduce the remaining literal `sorry` count.

Validation:
- `lake build HexMvGcd HexMvHensel HexMvFactor HexMvFactorizationTests` (619 jobs)
- `lake build HexMvGcd.Normalize HexMvGcd.Gauss HexMvGcd.Gcd` after review cleanup
- 75 benchmark-freshness unit tests
- `git diff --check`
- `#print axioms Hex.MvPoly.toUnivariate_mul`
- `#print axioms Hex.MvPoly.zero_product`

Both audited theorems depend only on `propext`, `Classical.choice`, and `Quot.sound`.

### DIFF
--- a/HexMvGcd/Gauss.lean
+++ b/HexMvGcd/Gauss.lean
@@ -451,7 +451,24 @@ variable [Lean.Grind.CommRing R] [DecidableEq R] [BEq R] [LawfulBEq R]
 /-- Gauss's lemma lifts proof-only gcd-domain structure through every finite
 multivariate arity. -/
 theorem gcdDomainLaws : GcdDomainLaws (MvPoly n R cmp) := by
-  sorry
+  refine {
+    dvd_iff := ?_
+    one_ne_zero := ?_
+    no_zero_div := ?_
+    gcd_exists := ?_ }
+  · intro a b
+    constructor
+    · rintro ⟨q, hq⟩
+      exact ⟨q, hq.trans (MvPoly.mul_comm q a)⟩
+    · rintro ⟨q, hq⟩
+      exact ⟨q, hq.trans (MvPoly.mul_comm a q)⟩
+  · intro hone
+    have hcoeff := congrArg (coeff (Mono.zero : Mono n)) hone
+    rw [coeff_one, coeff_zero, ite_eq_left rfl] at hcoeff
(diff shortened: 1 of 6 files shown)


kim-em:
I ran the requested independent Claude Opus review. It found no soundness blocker and confirmed that the new recursive-view multiplication and zero-product proofs are structurally independent of the remaining `gcd_exists` proof.

I addressed its concrete cleanup concerns:
- Removed the unused private classical division scaffolding, avoiding a future competing-`Div` problem because `Fraction` is indexed by the division instance.
- Renamed the public domain lemma to the short, unambiguous `zero_product`.
- Generalized the private fold helper over its element universe.

I retained `toUnivariate_mul` in `HexMvGcd.View`: a minimal build importing only `HexMvPoly.Recursive` cannot synthesize the `MvPoly` commutative-ring instance required by the convolution proof. `View` is the first existing Mathlib-free module with both the recursive definitions and packaged ring laws.

I also left the pre-existing polynomial `GcdDomainLaws` instance priority unchanged. Lowering it would not create a clean alternative yet: the executable `LawfulGcdOps` construction still has other proof fields that were elaborated through the Gauss-law path. That broader instance cleanup should accompany closure of `gcd_exists`.

kim-em:
The superseded CI run caught the benchmark freshness fingerprint for `HexPoly/Euclid/MulRing.lean`. I added the required exact proof-only transition record. The factorization freshness checker and its 75 unit tests now pass locally; a new CI run is starting on the amended head.

PR #10124, opened 2026-09-08
prove multivariate content multiplicativity

kim-em:
`contentIn` and scalar `content` could be computed, but their multiplicativity contracts were still admitted. This proves both from the executable recursive views and the proof-side Gauss content theorem, and generalizes recursive-view multiplication from variable zero to any selected variable.

The proof keeps certificate production unchanged. It identifies each recursive content with the scalar content up to normalized associates, applies lower-arity multiplicativity, and uses canonical normalization to obtain equality.

Validation:
- `lake build HexMvGcd.View`
- `lake build HexMvGcd.Gauss`
- `lake build HexMvGcd.Gcd`
- `lake build HexMvGcd.Content`
- `git diff --check`

Part of #10082.

### DIFF
--- a/HexMvGcd/Content.lean
+++ b/HexMvGcd/Content.lean
@@ -7,6 +7,7 @@ Authors: Kim Morrison
 module
 
 public import HexMvGcd.Cert
+public import HexMvGcd.Gauss
 
 @[expose] public section
 set_option backward.proofsInPublic true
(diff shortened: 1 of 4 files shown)


kim-em:
Independent Opus review found no mathematical correctness or trust-boundary defect. I verified its maintainability findings and updated the PR:

- replaced the older 46-line zero-degree split permutation proof with the new general permutation theorem;
- hoisted pairwise divisibility and normalized-associate equality into shared Gauss helpers, removing duplicate copies from `Gcd.lean` and `Content.lean`;
- shortened the content-local helper names, exposed `content_C`, and reused the constant-multiplication lemma in `primPart_mul`.

I kept the constant-embedding multiplication lemma in `HexMvGcd.Content` rather than moving it into `HexMvPoly.Basic`: this slice needs it, while changing the already released lower-level `HexMvPoly` API would widen the release boundary for no proof benefit. All four affected targets rebuild after the cleanup.

Issue #10141, opened 2026-09-09
spec: ordered real algebraic number library


kim-em:
## Objective

Write `SPEC/Libraries/hex-real-algebraic.md` for real algebraic numbers as an
executable ordered field, together with its Mathlib companion
`hex-real-algebraic-mathlib`. This is the "ordered real algebraic number layer
with exact comparison" asked for in response to the algebraic numbers
announcement
(https://kim-em.github.io/blog/2026-9-8-exact-algebraic-numbers-and-number-fields-in-lean/).

## Current state

hex-number-field already has the exact primitives; what is missing is the
type and its instances.

- `AlgebraicNumber.isReal` is exact at the stored separation precision, with
  the companion theorem `isReal_iff`.
- `AlgebraicNumber.realCompare` is an exact total order on real canonical
  numbers: structural equality first, then ball centres at
  `separationPrec (a.p * b.p)`. The companion proves it is the order of the
  real parts.
- `ZPoly.algebraicRoots` returns real roots first in increasing order, as a
  theorem about the values.
- `AlgebraicNumber` has a law-bearing `Field` in the companion, structural
  equality, `Repr` round trips, and arithmetic that exactifies on every
  operation.

There is no type of real algebraic numbers, no `LT`/`LE`/`Ord` instances, no
ordered-field structure, and no real-closed-field statement.

## Required design

- Represent real algebraic numbers as the subtype of canonical
  `AlgebraicNumber` by `isReal = true`, so equality stays structural and every
  existing operation (arithmetic, exactification, `algebraicRoots`, `conj`,
  `approx`, `Repr`) is reused rather than re-implemented. Decide in the SPEC
  whether this is a new library or an extension of hex-number-field; the
  default is a new library, since the release manifest is per-library and
  hex-number-field has shipped.
- Decide how closure of the subtype under the field operations and casts is
  carried. The recommended default is to recheck `isReal` on each result (one
  disc test at stored precision) behind a `panicWith` fallback that the
  companion proves unreachable, matching the existing `exact` / `_isSome`
  pattern. State the alternative (threading the proof) and why it was not
  chosen.
- The order: `LT`, `LE`, decidable `<` and `≤`, `Ord`/`compare`, `Max`, `Min`,
  `sign`, `abs`, with `realCompare` as the reference semantics. Provide the
  Mathlib-free instances of Lean core's order classes (`Init.Data.Order`) and
[...]

PR #10145, opened 2026-09-09
feat: implement exact ordered real algebraic numbers

kim-em:
Real algebraic values now have a canonical, executable subtype with exact comparison, arithmetic, square roots, ordered real polynomial roots, rational recognition, rounding, and certified dyadic approximations. This implements the specification introduced by #10144 for #10141.

The Mathlib-free library reuses `AlgebraicNumber` operations and checks closure at stored precision. Its proof-only law package supplies core order and `grind` instances; the companion proves the package and supplies `Field`, `LinearOrder`, `IsStrictOrderedRing`, `FloorRing`, and `IsRealClosed`. Correspondence proofs cover the real embedding and its algebraic image, root completeness and multiplicities, approximation bounds, and executable representation round trips.

Validation:

- Built both libraries, conformance and generated Repr checks, executables, and the downstream number-field and tower companions.
- Passed the compiled core checks, all 55 CI oracle cases and all 64 local-profile cases against pinned python-flint/FLINT, with no skips.
- Passed 11 oracle regression tests, including wrong answers, nonreal inputs, root identity and multiplicity, and missing capabilities; a deliberate CLI mismatch produced a replayable failure record.
- Verified fresh fixtures and generated Lean Repr expressions match the committed files; checked 151 relative Markdown links, the dependency DAG, manifest and conformance invariants, source lints, performance freshness, and `git diff --check`.

Extends the existing CI job and oracle runner, explicitly building both new libraries and the standalone check executable. The shared Python installation now pins python-flint 0.9.0 for all FLINT oracles so the test-only binding has a reproducible FLINT 3.6.0 ABI. The libraries remain unreleased; no release-manifest entry is added. Faster comparison algorithms and an unconditional Mathlib-free proof of the law package remain the separate follow-ups specified by the design.


kim-em:
Claude Opus reviewed the implementation. Addressed the confirmed findings in b485316af:

- Added both libraries to CI's explicit build/cache target list and the standalone check executable to the conformance build.
- Fixed nested `Repr` parentheses and added generated, compiled `Option` round trips.
- Removed global order-to-real simp rewrites; added notation-level power/scalar correspondence and core dictionary/data checks, including numeric literals.
- Separated general-root capability from scalar identity in the adapter, added a regression for the missing general-root case, and centralized Python version constants.
- Clarified the printed-syntax proof boundary and aligned the square-root SPEC formulas with the theorem statements.

Two suggested changes were not appropriate after checking the sources:

- The reported FLINT allocation mismatch is absent in the pinned version. FLINT 3.6.0's [fmpz_vec.h](https://github.com/flintlib/flint/blob/v3.6.0/src/fmpz_vec.h#L61-L63) declares the exact resizable-vector signatures used here; [gr_poly.h](https://github.com/flintlib/flint/blob/v3.6.0/src/gr_poly.h#L578-L584) takes `fmpz_vec_t` multiplicities. These differ from the older underscored raw-vector API.
- Matching inside the square would assume a stronger enclosure than Hex certifies. `RefinedIsolation.root_mem_closedDisc` guarantees the circumscribed disc. The oracle retains that enclosure and requires unique matching plus a bijection; ambiguous matches fail. Added a comment explaining the argument.

Scalar checking retains the pinned wheel's public C qqbar contexts because python-flint has no supported transfer of an arbitrary selected polynomial root into a Python `_gr` value. The algorithms remain FLINT's independent exact arithmetic; the Python scalar interface is separately probed. The SPEC now explains that boundary. Allocation ownership is bounded by the fixture run (55 CI / 64 local records).

`done_through: 1` records the admitted lifecycle phase, not a count of implemented features; this PR does not claim the separate Phase-2 scaffolding audit or later performance/polishing admission.

Revalidation passed: both libraries and companion proofs, core and nested Repr compilation, all 55 CI oracle records without skips, and 11 oracle regression tests. The earlier 64-record local profile also passed. CI is rerunning on the updated head.

kim-em:
CI attempt 1 on b485316af passed the full build, all conformance oracles, and every benchmark correctness check. The benchmark step failed only its combined wall-clock limit: 416 seconds against the 360-second hard cap. Largest contributors were hexroots (96s), hexgf2 (78s), hexnumberfieldtower (63s), hexnumberfield (45s), and hexmodular (44s). The failed run and its logs remain available at https://github.com/kim-em/hex-dev/actions/runs/34328590607/attempts/1.

Rerunning the unchanged failed job once to determine whether this operational timing failure repeats. No benchmark corpus, cap, or correctness check has been relaxed.

kim-em:
The unchanged rerun again passed the build, all oracles, and every benchmark correctness check, but took 375 seconds against the 360-second cap. Both attempts are retained; no further unchanged rerun is being requested.

c8bf1c406 reduces the concurrent oracle pool and emitter worker limit from two to one. Two emitters with two workers each can consume the four-vCPU runner alongside the benchmark tail. The new limits reduce that contention while retaining the two verification tails, every fixture and benchmark, and the existing hard cap. This changes scheduling only; conformance may take longer. YAML validation confirms the single-job structure and required-oracle gate remain intact.

Issue #10147, opened 2026-09-09
spec: direct certified radicals and cyclotomic fast paths


kim-em:
## Objective

Write the thorough SPEC for direct certified radicals of `AlgebraicNumber`, including cyclotomic fast paths. Specify the complete executable algorithm and its Mathlib proofs, not merely an approximation shortcut or a wrapper around `AlgebraicPoly.roots`.

## Current state

The principal `sqrt`/`nthRoot` API has complete Mathlib correspondence. Its general implementation solves `X^n - a` with the algebraic-coefficient solver, which constructs a common field, a norm eliminant, and an evaluation eliminant. Isolation-based branch selection is being added separately. The canonical constructor still isolates every root of a selected minimal polynomial. `hex-cyclotomic` has an existing SPEC but no implementation.

## Required design

- Construct the annihilator `p(X^n)` directly from the input minimal polynomial. Eliminate common-field discovery and generic norm/evaluation resultants from this route.
- Give certified principal-root enclosures, a square-root specialization, branch-cut handling on and near the negative real axis, and input-computable bounds proving termination. State zero/index-zero/index-one conventions explicitly and retain the existing Mathlib branches.
- Certify the correct embedding and irreducible factor without exactifying all candidates. Reuse factorization and isolation work through final construction. For primitive roots of unity, coprime powers have the same minimal polynomial: specify how to reuse it and certify the selected embedding, avoiding redundant minimal-polynomial computation.
- Address the canonical constructor's all-roots isolation cost with a concrete design that reduces it while preserving structural equality, conjugation, canonical provenance, and Repr round trips. Name the necessary representation changes and proof obligations; do not silently assume a certified arbitrary enclosure is canonical.
- Cover rational radicals, perfect-power/binomial cases, composite root indices, exact recognition of roots of unity, and cyclotomic construction using the existing `SPEC/Libraries/hex-cyclotomic.md`. Make the relationship to the rational-angle roots-of-unity constructor explicit.
- Specify computational and Mathlib APIs, all missing bridge lemmas, failure behavior, bounded search, dependency graph, and compatibility/migration.
- Include the canonical integer-root construction of `1099511627776*X^2 + 1099513724929` as a high-height stress case. It exhausted an 8 GiB process cap before comparison in local testing (bug #10156); account for isolation and factorization memory as well as runtime.
[...]

PR #10168, opened 2026-09-10
spec: refine direct radical certification, costs, and migration

kim-em:
The direct-radical SPEC's local canonicalization could require roughly 150,000 exact checks per value, and its evidence migration omitted an executable producer and a computational consumer of the old Boolean field. Tighten the fixed grid and search window to fewer than 10,000 candidates, stop at the first lexicographic success, and require constructor performance evidence before migration. Specify retained factor-output evidence, a primitive field for Mathlib-free consumers, and the resulting cyclotomic dependency and delivery order.

Also make the Taylor coefficient denominator correction explicit, bound composite plans, preserve the current `rootNear` printer and zero exception, require generated fixture regeneration, and document the unity fallback's precision and allocation limits. Update the high-height quadratic requirements to reference the fix and measurements now on main. This follows #10164 and #10147; no Lean implementation changes. An exact runtime-neutral blob exemption repairs the base branch's benchmark-freshness failure for the shim docstring update.

Validation: dependency DAG, manual split, release manifest, relative links and anchors, `git diff --check`, and exact rational checks of the revised Pellet/grid constants. Opus reviewed the merged design and independently verified the corrected mathematics and executable design. Its remaining wording and schematic type corrections are included. Both factorization and graph-isomorphism freshness checks pass locally.


kim-em:
The requested Opus review completed successfully. A second focused review verified the revised Taylor/Pellet inequalities, tighter grid and lexicographic early exit, zero exception, retained-factor evidence, primitive field, and migration obligations. Its final sentence and schematic type corrections are included.

Verified findings addressed:
- Reduce the canonical search from roughly 150,000 candidates to at most 92², stop at the first success, and require constructor performance evidence before migration.
- Specify the retained factor array and erased equality proof, migrate the tower's Boolean-field consumer to explicit primitivity, and make the cyclotomic dependency and delivery order explicit.
- Document index-dependent precision and the unity fallback's practical resource limits; require allocation checks in budgeted APIs.
- Require generated Repr/fixture regeneration, preserve zero and existing `rootNear` round trips, clarify prime-power bounds, reuse `xPowSubOne`, and require a cyclotomic manual chapter.

One factual review claim was rejected: the shipped unity route already factorizes through `RootSelection.maximum?` → `root.exact`. Replacing the complete fallback with a supposedly complete index-certificate search would also assume infrastructure that does not exist. The complete fallback remains, with explicit resource limitations.

The quadratic memory report now supplies the fixed baseline. CI additionally exposed an unrelated missing freshness exemption for a docstring-only change on main; the exact blob transition was inspected and recorded. Local freshness checks pass. Final CI is running.

PR #10241, opened 2026-09-12
feat(mv-poly): add canonical kernel list form

### LINKED ISSUE #10229: feat(mv-poly): canonical list form of MvPoly arithmetic for kernel reduction
## Objective

Give `Hex.MvPoly` a canonical list form for kernel reduction, with its denotation theorems, so that certificates over polynomial entries can be checked by the kernel under the discipline of `SPEC/matrix-tactics.md`. This is the shared prerequisite that the symbolic `rank` arm (https://github.com/kim-em/hex-dev/issues/10223), the `rank_locus` tactic (https://github.com/kim-em/hex-dev/issues/10226) and the symbolic `det` tactic all name. It is one issue for the SPEC section in `HexMvPoly/SPEC/hex-mv-poly.md` and its implementation, since the design is fixed by its consumers.

## Current state

- `Hex.MvPoly n R cmp` is a structure over `Mono n := Vector Nat n` exponent vectors with a fixed monomial order. `HexMvPoly/KernelTests.lean` checks that the reference operations reduce in the kernel on small closed examples, but `Vector` and `Fin` are on that path, which `SPEC/matrix-tactics.md` §Kernel discipline rules out for certificate checking (the reference `checkRank` on `Hex.Matrix` took 371 s at `16 × 16` where the list form takes 115 ms).
- The consumers need exactly this (recorded in `SPEC/Libraries/hex-generic-rank-mathlib.md` §Prerequisite changes and `HexDeterminantalIdealMathlib/SPEC/hex-determinantal-ideal-mathlib.md` §Kernel route): canonical term lists, addition, multiplication, negation, scalar multiplication, zero test and equality by structural recursion over `List Nat` exponents and a kernel-cheap coefficient representation, with a denotation into `MvPoly` and the theorems that make dropping zeros and duplicates transport.

## Required design and work

- `HexMvPoly/Kernel.lean`: `Term := List Nat × κ` and `PolyList := List Term` for a coefficient representation `κ` (`Int` first; `Rat` as a reduced numerator-denominator pair; residues as `Nat` with the modulus a parameter), canonical when the exponent lists are strictly decreasing in the fixed order and no coefficient is zero. Operations `add`, `mul`, `neg`, `smul`, `isZero`, `beq`, and `evalAt` at a point in `κ`, all `@[expose]`, structural recursion only, arithmetic through `Nat.add`/`Nat.mul`/`Int.add`/`Int.mul` and comparisons through `Nat.blt`/`Nat.beq`/`Int.decEq`; no `Array`, `Vector`, `Fin`, `dite` or well-founded recursion on the path.
- Denotation `denote : PolyList → MvPoly k C cmp` (Mathlib-free, into the reference type) with `denote_add`, `denote_mul`, `denote_neg`, `denote_smul`, and the canonical-form theorems the consumers require: `isZero_iff : isZero l = true ↔ denote l = 0`, `beq_iff : beq l l' = true ↔ denote l = denote l'` for canonical inputs, and that every operation returns a canonical list from canonical inputs. `evalAt` commutes with `MvPoly.eval`.
- A conversion `toList : MvPoly k C cmp → PolyList` with `denote_toList`, so a compiled producer can quote its result as a term list and the kernel never sees the reference structure; and the identification of a `Hex.Matrix (MvPoly …)` quoted by term lists with its `rowLists`, as `ofLists` does for integer literals.
- Companion (`HexMvPolyMathlib`): the denotation composed with `HexMvPolyMathlib.equiv` into `MvPolynomial`, so the consumers' Mathlib statements read off the same lists.
- SPEC section in `hex-mv-poly.md` (and the companion SPEC) stating the representation, the canonical-form invariants, the theorems, and the kernel discipline it satisfies; `KernelTests.lean` gains list-form examples with timings.
- No `sorry`, no `axiom`, no `native_decide`.

## Verification

- `lake build`, `scripts/check_dag.py`; the kernel tests replay a `4 × 4` polynomial certificate identity through the list form in well under a second.
- `beq` is shown to be an equivalence on canonical lists and `isZero` to agree with `denote` on the fixtures.

kim-em:
Symbolic matrix certificates need polynomial arithmetic that reduces without reference trees, vectors, or finite indices. This adds canonical `List Nat` term lists for that replay path, with proved denotation into `Hex.MvPoly` and Mathlib's `MvPolynomial`.

- Exposed structural arithmetic: linear-merge addition, balanced-row multiplication, negation, subtraction, scalar multiplication, identity, zero/equality checks, and evaluation.
- Arity-aware canonicality and preservation proofs, semantic zero/equality laws, and producer conversion round trips. `toList` has a checked linear path for the default lexicographic order.
- `denote_toRows` identifies nested quoted polynomial lists with their original rows; consumers instantiate it at `rowLists P` without adding a matrix dependency to `hex-mv-poly`.
- Int and Rat fixtures, plus a non-diagonal 4×4 adjugate identity checked against independent coefficient literals in both multiplication orders. A corrupted determinant is rejected.

Canonicality requires every exponent list to have the declared arity; otherwise denotation into a fixed-arity polynomial is not injective. The list order is fixed lexicographic order, independently of the reference storage comparator. The coefficient operations are generic; positive-characteristic reflection providers remain a separate prerequisite in the consumer SPECs.

Validation: `lake build` passed all 13,738 jobs; the focused core/test/Mathlib targets, `python3 scripts/check_dag.py`, and diff hygiene checks also passed. The 4×4 certificate's kernel check measured 77 ms on the shared host. The scoped profiler reports this timing but does not enforce a wall-clock limit; existing reference/proxy benchmarks do not establish a general PolyList scaling budget.

Closes #10229

### DIFF
--- a/HexMvPoly.lean
+++ b/HexMvPoly.lean
@@ -14,6 +14,7 @@ public import HexMvPoly.Eval
 public import HexMvPoly.Structural
 public import HexMvPoly.Recursive
 public import HexMvPoly.Ring
+public import HexMvPoly.Kernel
 
 public section
 
--- a/HexMvPoly/Kernel.lean
+++ b/HexMvPoly/Kernel.lean
@@ -1,0 +1,1230 @@
+/-
+Copyright (c) 2026 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
+-/
+
+module
+
+import HexBasic.Fold
+public import HexMvPoly.Ring
+public import HexMvPoly.Structural
+
+@[expose] public section
+set_option backward.proofsInPublic true
+
+/-!
+Canonical list arithmetic for kernel certificate replay.
+
+Unlike the reference `MvPoly` representation, this module never traverses an
+`Array`, `Vector`, or `Fin` while computing. Exponents are ordinary lists and
+terms are kept in descending lexicographic order. The coefficient operations
+are deliberately the ordinary operations on the representation type: at
+`Int` they reduce to `Int.add`, `Int.mul`, and `Int.neg`; `Rat` is already a
+reduced numerator--denominator representation; a residue provider may use its
+canonical `Nat` representative.
+-/
+
+namespace Hex.MvPoly.Kernel
+
+open Hex
+open scoped Hex
+
+universe u
+
+/-- A list-form term: an exponent list and a coefficient. -/
+abbrev Term (κ : Type u) := List Nat × κ
+
+/-- A sparse polynomial represented by a canonical list of terms. -/
+abbrev PolyList (κ : Type u) := List (Term κ)
+
+/-! # Kernel primitives -/
+
+/-- Equality of exponent lists, using only `Nat.beq` and structural recursion. -/
+def expBeq : List Nat → List Nat → Bool
+  | [], [] => true
+  | a :: as, b :: bs => Nat.beq a b && expBeq as bs
+  | _, _ => false
+
+/-- Lexicographic comparison of exponent lists, using only `Nat.blt`. -/
+def expCmp : List Nat → List Nat → Ordering
+  | [], [] => .eq
+  | [], _ :: _ => .lt
+  | _ :: _, [] => .gt
+  | a :: as, b :: bs =>
+      if Nat.blt a b then .lt
+      else if Nat.blt b a then .gt
+      else expCmp as bs
+
+/-- Pointwise addition of exponent lists. Canonical inputs have equal length;
+the trailing clauses make the operation total on untrusted data. -/
+def addExp : List Nat → List Nat → List Nat
+  | [], bs => bs
+  | as, [] => as
+  | a :: as, b :: bs => Nat.add a b :: addExp as bs
+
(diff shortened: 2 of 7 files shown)


kim-em:
Independent Claude Opus review completed twice. The follow-up found no substantive correctness blockers and confirmed the structural merge budgets, linear translated rows, denotation proofs, and independent 4×4 tridiagonal certificate.

Addressed findings include linear-merge addition and balanced multiplication; `denote_one`/`denote_sub`; nested-row quotation; stronger Int/Rat fixtures; literal certificate validation and rejection of corruption; an axiom guard (`propext` only); and a theorem tying public addition to every sufficient merge budget. Corrected the benchmark claims and the determinantal-ideal SPEC's denotation type.

The scoped profiler remains intentionally enabled on the requested timing example; it reports timing without claiming timeout enforcement. Grevlex producer conversion can still use the documented quadratic normalization fallback. The positive-characteristic residue reflection provider remains the separate prerequisite named by the consumer SPECs; it is not implemented by this PR.

The final commit passed `lake build` (13,738 jobs), focused kernel/Mathlib builds, the DAG checker, and diff hygiene. The revised 4×4 kernel checks observed 63–77 ms on this host. Waiting for CI on the final commit before merge.

Issue #10333, opened 2026-09-20
Implement ordered-domain pseudo-division for real-closure queries


kim-em:
> Superseded by #10375, the consolidated implementation assignment following merged SPEC PR #10365. Computational, companion-proof, conformance and performance obligations carry forward there. This issue is closed as superseded, not as implemented.

library: HexPoly

## Architecture hold

Correction #10365 now separates executable representations from semantic field proofs. Revert #10366 is already merged. Keep implementation paused until the updated correction passes fresh Opus review and CI, merges, and the coordinator releases this issue. The earlier benchmark-policy exception proposal is withdrawn; preserve Mathlib-free benchmarks. Opus review currently needs refreshed Claude CLI authentication. The superseded fallible PolyOps requirements are not authorized.

## Governing contract

[HexPoly/SPEC/hex-poly.md](https://github.com/kim-em/hex-dev/blob/main/HexPoly/SPEC/hex-poly.md). Read the corrected version before starting. The family retains exact arithmetic, user-supplied transcendental approximations and infinitesimal/algebraic towers. HexInterval and HexIntervalMathlib are out of bounds.

## Executable and proof boundary

Follow the [shared execution contract in correction #10365](https://github.com/kim-em/hex-dev/blob/issue-10143-user-oracles/SPEC/real-closure-execution.md), once reviewed and merged. Computational representation coefficients have ordinary total operations, canonical zero and honest structural DecidableEq. They reuse DensePoly kernels without a field instance on raw syntax. Semantic equality/replay identities use zero differences. Companions prove operation/sign preservation and zero reflection into a lawful semantic field; this representation map need not be injective. Quotient-field proofs do not enter executable algebraic constructors. No benchmark import exception, PolyOps, fallible arithmetic or HexInterval work is authorized.

## Deliverables

The existing total DensePoly field division, gcd, xgcd, xgcdLeft and monicize APIs already provide field arithmetic. The rejected PolyOps implementation #10348 was reverted by #10366. Do not restore it or invent a replacement coefficient record.

[...]

Issue #10332, opened 2026-09-20
Prove shared Sturm–Tarski and integer query correspondence


kim-em:
> Superseded by #10375, the consolidated implementation assignment following merged SPEC PR #10365. Computational, companion-proof, conformance and performance obligations carry forward there. This issue is closed as superseded, not as implemented.

library: HexRealRootsMathlib

## Architecture hold

Correction #10365 now separates executable representations from semantic field proofs. Revert #10366 is already merged. Keep implementation paused until the updated correction passes fresh Opus review and CI, merges, and the coordinator releases this issue. The earlier benchmark-policy exception proposal is withdrawn; preserve Mathlib-free benchmarks. Opus review currently needs refreshed Claude CLI authentication. The superseded fallible PolyOps requirements are not authorized.

## Governing contract

[HexRealRootsMathlib/SPEC/hex-real-roots-mathlib.md](https://github.com/kim-em/hex-dev/blob/main/HexRealRootsMathlib/SPEC/hex-real-roots-mathlib.md). Read the corrected version before starting. The family retains exact arithmetic, user-supplied transcendental approximations and infinitesimal/algebraic towers. HexInterval and HexIntervalMathlib are out of bounds.

## Executable and proof boundary

Follow the [shared execution contract in correction #10365](https://github.com/kim-em/hex-dev/blob/issue-10143-user-oracles/SPEC/real-closure-execution.md), once reviewed and merged. Computational representation coefficients have ordinary total operations, canonical zero and honest structural DecidableEq. They reuse DensePoly kernels without a field instance on raw syntax. Semantic equality/replay identities use zero differences. Companions prove operation/sign preservation and zero reflection into a lawful semantic field; this representation map need not be injective. Quotient-field proofs do not enter executable algebraic constructors. No benchmark import exception, PolyOps, fallible arithmetic or HexInterval work is authorized.

## Deliverables

Real.instIsRealClosed is already implemented by #10347 at 825fea66c757744a080f0e90cd5034beaa70d88c. Preserve that independent result. No PolyOps interpretation work is required or authorized.

For the actual #10335 API, reuse ordinary DensePoly/Polynomial correspondence and prove pseudo-remainder scaling, degree, endpoint and literal-replay bridges over an ordered commutative domain with an order-preserving embedding into an arbitrary ordered real closed field. A Field hypothesis on the coefficient domain is forbidden; Int is an instance. Own the shared query theorem and integer specialization here, with no reverse HexSturm dependency.

[...]

Issue #10331, opened 2026-09-20
Implement the real-closure family through verified APIs and performance evidence


kim-em:
## Objective and scope

Complete the computational and proof implementation envisaged by #10143: ordered extensions, real-closure towers, exploration, real-valued rcf integration, conformance and performance evidence. The eight original SPEC directives are design provenance, not implementation completion.

PR #10365 is merged as ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c. The [shared execution contract](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/real-closure-execution.md) and owning SPECs govern implementation. Use ordinary total operations, canonical-zero representations and noninjective semantic interpretation; no field instance on raw representatives or fallible PolyOps framework. #10366 reverted the rejected PolyOps implementation. Computational libraries and benchmarks remain Mathlib-free.

Approximation is caller-supplied, with separate containment and requested-width proofs; universal sign progress also uses relative transcendence. HexInterval, HexIntervalMathlib, bundled pi/e providers and their analytic/performance work are out of bounds. Previously cancelled interval work stays cancelled.

## Experiments, specifications and implementation

E1 compared storage retention, smaller defining polynomials and known-irreducible zero tests using existing division/gcd kernels at one and two algebraic levels. All 360 paired arm samples were retained; independent python-flint checks agreed. E2 checked persistent refinement, a changed dependent defining coefficient, live-value transport and stale binding rejection. See the [experiment report](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/reports/real-closure-storage-experiments.md) for evidence and limits.

The merged SPECs adopt exact remainder retention for monic clean definitions, preserve the general non-monic clean path, and require explicit context transport and binding checks. These experiments establish bounded feasibility, not universal selected-root proofs, a general tower implementation or Phase-4 performance.

## Consolidated implementation assignments

| Assignment | Issue | Dependency ordering |
| --- | --- | --- |
| Effective polynomial/Sturm–Tarski computation and algebraic proofs | #10375 | Existing infrastructure; completion does not wait for Tau Ceti |
| Sturm–Tarski root semantics and dependent proof evidence | #10389 | Actual required Tau Ceti signed-remainder/Cauchy-index theorem, plus merged effective APIs; not the whole roadmap |
[...]

PR #10347, opened 2026-09-20
feat(real-roots): share the real closed field instance

kim-em:
The shared real-root companion now exports `Real.instIsRealClosed`, so downstream proofs can use Mathlib's real-closed-field API over ℝ without importing the real-algebraic companion. The existing odd-root proof moves into `HexRealRootsMathlib.RealClosed`; the real-algebraic instance reuses `IsRealClosed.exists_isRoot_of_odd_natDegree`.

This is tranche 1 only. Refs #10332; coordinator #10331. The issue stays open for the fallible operation-record interpretation, shared Sturm–Tarski/replay soundness, integer specializations, and their regressions. Those require the API milestones from #10333/#10335 and actual proved Tau Ceti foundations; roadmap PR TauCetiProject/TauCetiRoadmap#420 is not proof delivery.

Validation:
- `lake build HexRealRootsMathlib HexRealAlgebraicMathlib HexRealRootsMathlib.ProofProbe.RealClosed HexRealRootsMathlib.ProofProbe.RealClosedBaseline` passes.
- Instance synthesis and arbitrary odd-degree root existence are checked through the lower umbrella; guarded axiom sets are exactly `propext`, `Classical.choice`, and `Quot.sound`, including the downstream instance.
- Import-DAG/umbrella checks, probe-manifest validation, existing probe-manifest tests, and `git diff --check` pass.
- Four alternating fresh-module pairs, clean source provenance, artifact sizes and raw samples are in `reports/hex-real-closed-proof.md`. All completed samples are retained; no incremental-cost claim or phase-counter change.

The local dependency cache needed a full official refresh before the broader build; `lake build --no-build Mathlib` then confirmed all Mathlib targets were up to date.


kim-em:
Fresh Claude Opus review found no Lean soundness or dependency-placement defect. I verified its findings against the source and addressed both medium concerns:

- Updated the real-algebraic SPEC dependency diagram and proof route, corrected its Mathlib pin link, and updated both companion READMEs.
- Recollected the four alternating fresh-module pairs at published clean commit af51d5a6d88abf95650c59e021df16d382c70258. The current report links that commit and raw artifact. The earlier samples remain archived with their provenance limitation stated explicitly.

The report now emphasizes that this is an umbrella/export and ordinary theorem-use measurement, dominated by importing dependencies. It does not claim to measure construction of the real instance or future literal-query replay.

I retained the focused-module axiom guard and the unchanged private helper name: both keep the moved proof and its trust check easy to compare with the downstream original. The shared harness validates the new manifest, and the existing probe-manifest tests pass; I did not add tests that merely restate the single fixed pair or expand CI for them. A future Mathlib instance will require removing this local declaration rather than renaming it; there is no duplicate at the current pin.

Local verification also completed the full benchmark import check: 54 benchmark executables and 2,465 build-only proof probes pass. No Lean source changed after the successful affected-library build.

Issue #10342, opened 2026-09-20
impl: hex-interval-mathlib certified π and e point providers for ordered-field extensions


kim-em:
library: HexIntervalMathlib

## Current state

This is the evolving implementation directive for the interval companion's #10143 prerequisite slice, coordinated by #10331. It does not request completion of the entire interval solver.

Existing public modules already prove real containment for checked addition, subtraction, multiplication, intersection, regularization and the supported reciprocal/division slice. Reuse them. HexOrderedFnMathlib owns Horner enclosure composition/convergence and oracle-context registration for its rational functions.

The missing part is reusable certified named-constant point providers. The interval companion SPEC's “Certified point bounds” prescribes reduction, convergent exact approximation, outward remainder, reconstruction, and outward rounding. “Exponential and logarithm” says “Nullary π propagation returns successively tighter dyadic enclosures.” Existing `Experiment/PntPiPoint` authenticates a fixed `315/100` bound using `Real.pi_lt_d2`; it is not an arbitrary-precision provider. `PntExpPoint` and `PntExpNegative` are fixed source families, explicitly not precision-driven generated-cut algorithms. A generic callback parameter or replay assuming its desired source proposition does not deliver the promised π/e constructor.

## Deliverables

1. Audit and expose the existing containment API needed by the family without duplicating it. Independently identify the smallest reusable point-certificate format and ordinary-kernel analytic proof route for π and `Real.exp 1`, following the SPEC's exact arithmetic/remainder discipline. Check relevant pinned Mathlib facts and existing source-family proofs before deriving new lemmas; do not rely on a `proof_wanted` Chudnovsky identity.
2. Coordinate bounded exact runtime approximation and certificate/checker production with the existing HexInterval implementation directive. Those algorithms must remain Mathlib-free and terminate/resource-check on arbitrary inputs. This companion supplies the analytic meanings and replay theorems in `Elementary/Constants.lean` and the relevant `Elementary/ExpLog.lean` slice. Initial useful certified bounds may land first, but do not close this issue on fixed-precision fixtures.
3. Prove each accepted point certificate contains the exact named real, including outward rounding, reduction/reconstruction, exact subject and precision claims. Authenticate source-rule identities; unchecked runtime endpoints cannot provide their own proof. Supply kernel proof reconstruction with no compiled trust. Provider registrations into an OrderedFn context belong downstream and must bind these exact source theorems.
[...]

Issue #10334, opened 2026-09-20
impl: hex-interval readiness and bounded π/e providers for ordered-field extensions


kim-em:
library: HexInterval

## Current state

HexInterval is implemented extensively but `libraries.yml` still records `done_through: 0`. Existing public `Interval.lean`/`Multiplication.lean` expose resource-checked addition, multiplication, intersection, outward regularization and basic reciprocal/division. The umbrella explicitly identifies later default-policy/terminal layers, and the SPEC preserves unsupported operations as planned surface. HexOrderedFn needs bounded dyadic enclosures and deterministic coefficient/constant refinement, which it owns itself; it does not need the entire experimental interval frontend or a general solver.

No currently open implementation issue covers this prerequisite. This directive audits/finishes the owning interval SPEC's admission and the exact computational slice needed by #10143; it does not authorize broad unrelated solver development.

## Deliverables

First tranche: audit the existing public API against the SPEC and Phase-1 criteria, recording implemented declarations and remaining mandatory gaps on this issue. Build the library and characterize existing arithmetic/sign-cut/resource semantics. Establish whether the Phase-1 admission can honestly advance to 1. If yes, make the minimal readiness change with evidence; if not, identify precise missing mandatory API and the smallest correct implementation, and report before expanding scope. Never bump merely because a consumer is waiting.

Verify the ordered-fn prerequisite slice is usable through public imports: constructing outward finite dyadic intervals for exact rational coefficients, intersection with preserved endpoint openness, add/multiply Horner arithmetic, regularization, comparison against zero and preflighted resource failures. Generic oracle scheduling/provenance remains HexOrderedFn's job. Expose existing supported primitives only where the owning SPEC permits, and fix any actual in-scope bug found with direct regression coverage.

Keep follow-on review/conformance/benchmark admission work in this same owning issue if required for the family phase gates; tie each requested advancement to actual Phase criteria. Do not represent this scoped prerequisite audit as completion of every unrelated numerical frontend/provider in the 6000-line interval SPEC.

## Named constant providers

[...]

Issue #10143, opened 2026-09-09
spec: real closures of ordered fields with infinitesimals and transcendentals (umbrella)


kim-em:
## Objective

Design, as a family of SPECs, computation in real closures of `ℚ` extended by
computable transcendental constants and by infinitesimals, following
de Moura and Passmore, "Computation in Real Closed Infinitesimal and
Transcendental Extensions of the Rationals" (CADE 2013),
https://www.cl.cam.ac.uk/~gp351/infinitesimals.pdf, implemented as Z3's
`RCF` module. This umbrella holds the family-level design: what exists, what
is missing, the proposed division into libraries, the user-facing surface,
and the proof-side risks. Its deliverable is the family section of
`SPEC/future-work.md` plus one `spec:` directive per library, which are the
issues that write the individual SPECs. Implementation is a further layer
behind all of them.

## What the paper does

A tower `ℚ ⊂ transcendentals ⊂ infinitesimals ⊂ algebraics`, in that
mandatory order. Transcendental and infinitesimal extensions are rational
function fields `K(τ)`, `K(ε)` in coprime numerator / monic denominator
normal form. Sign in `K(τ)` comes from interval refinement using a
user-supplied approximation procedure for `τ`; sign in `K(ε)` is the sign of
the lowest-order nonzero coefficient, with each new infinitesimal below every
positive element of the previous field. Algebraic extensions use a squarefree
but not necessarily irreducible defining polynomial; a root is identified by
an interval with possibly infinite endpoints plus a Thom-style sign
assignment on some derivatives; the sign of `q(α)` is computed by
Ben-Or–Kozen–Reif sign determination built on Tarski queries; inversion is
extended gcd with dynamic splitting of the defining polynomial when the gcd
is nontrivial; root isolation bisects while the Cauchy-type bound is finite
and falls back to sign determination otherwise; and a "clean representation"
discipline (signed pseudo-remainder Sturm sequences, no reduction by
non-monic defining polynomials) is the optimisation that matters (0.28 s
versus a 30-minute timeout on the paper's `tower8` example).

## What Hex already has

- hex-rational-fn is exactly the paper's `K(τ)` normal form, generic over any
  `Lean.Grind.Field` with decidable equality. Only the ordering is missing.
- `DensePoly K` has gcd, extended gcd, Yun decomposition and division over an
  arbitrary field, which tower-level Sturm chains and inversion need.
- hex-real-roots has Sturm chains, sign variations, counts at `±∞`, and
[...]

PR #10349, opened 2026-09-20
feat(interval-mathlib): certify arbitrary-order pi and exp-one source bounds

kim-em:
The interval companion lacked reusable analytic source bounds for π and exp(1). Add ordinary-kernel bounds at every approximation order: Machin's formula with a geometric arctangent remainder, and the factorial series with Mathlib's exponential remainder. Both sources prove that order `k+4` leaves half of the requested width for outward rounding.

The public umbrella exposes the two elementary modules. Public-import probes certify both dyadic cuts at 2, 8 and 16 bits, derive π−4<0 and 2<exp(1)<3, demonstrate why exp order zero must be rejected, and audit all new theorem axiom sets. The containment audit and exact certificate/proof handoff are in `reports/hex-interval-mathlib-constants.md`; #10334 has agreed to the same formulas.

Validation: `lake build HexIntervalMathlib HexIntervalMathlibReplayProbe`, DAG check, `git diff --check`, and all six fixtures independently enclosed by python-flint 0.9.0 / 256-bit Arb. Four-round fresh-module AB/BA measurements passed all 24 builds from clean commit `0dd499e6`; the largest probe maximum was 8.755 s against its 60 s cap. Raw samples, artifact sizes, source/toolchain/host provenance and Arb output are committed under `reports/bench-results/`. Existing CI build and conformance paths cover the additions. The shared oracle registry classifies oracle-only edits correctly; all 14 classifier/runner regression tests pass, including failure propagation.

This is a complete analytic source/progress tranche, not a completed runtime provider. The bounded Mathlib-free generator/checker in #10334, its exact casts and rounding bridge, literal reconstruction, resource-success schedule and adversarial certificate tests remain open. No library phase counters advance.

Refs #10342. Coordinated by #10331; paired runtime work in #10334.


kim-em:
Fresh Claude Opus review checked the pinned identities, arbitrary-order error/width algebra, fixtures, trust audits, and the agreed runtime formula boundary. I separately ran the Lean, Python, DAG and measurement checks; the reviewer used read-only source and retained artifacts.

Addressed the substantive CI finding: the Arb source-proof oracle is now a `HexIntervalMathlib|-|...|-` entry in the shared oracle registry. The bespoke workflow block is removed. Regression tests prove that an oracle-only change selects the companion, no emitter build occurs, and oracle failure fails the runner. All 14 classifier/runner tests and the actual Arb runner pass.

The report now distinguishes requested width `k` from positive Core grid precision `k+2`, and explains that `k+4` is a sufficient termination witness, not a mandated planner order. This was shared with #10334. I did not adopt the suggested `floor(k/4)+1` π schedule: at k=3 it chooses n=1, whose analytic width exceeds 4/15 rather than being at most 1/16. A 16-based radius bound would instead justify `ceil((k+4)/4)`. The current conservative schedule is sound and accepts smaller planner-selected orders through the same source theorems.

The existing common-order payload and `Hex.Interval.Pi` namespace remain aligned with the agreed runtime interface; separate Machin orders and a wider arctangent API are later optimizations, not correctness blockers here. The report also clarifies that the measurement parser records the first axiom line, while the guarded Lean checks enforce every source and fixture theorem's axiom set. Fixed 2/8/16-bit fixtures continue to use the documented 256-bit Arb precision. No proof source changed during these review fixes, and all 51 recorded measurement-source hashes still match.

kim-em:
Merged current main at 91e6b0e35 without conflicts, including its existing cache fix for the slow Mathlib-free proof-probe lint. The PR-specific diff is unchanged. Local public library/probe build, DAG check, 14 classifier/runner tests, bench dependency lint (55 executables / 2477 proof probes), and diff check all pass. The refreshed CI run is 35512376329; I am retaining a single gh run watch monitor at 120-second intervals.

kim-em:
The user explicitly removed transcendental support from the #10143/#10331 family. This interval/provider work is no longer required by that effort. Closing without merge; preserve branches and local proof/measurement artifacts. This is a scope cancellation, not a correctness judgment. Do not resume under the family coordinator.

Issue #10352, opened 2026-09-20
Complete hex-rank Phase-4 benchmarks and performance evidence for certified row selection


kim-em:
library: HexRank

## Current state

HexSignDet's merged SPEC consumes HexRank for certified row selection and nonsingular minors, so the #10143 family tracked by #10331 needs this existing prerequisite to finish Phase 4. This issue completes HexRank's already-specified performance obligations; it does not add BKR logic or a new rank algorithm.

At current main `041918abcb653281aa32c28c24974a2eff17f5c6`, HexRank has `done_through: 3`. Every direct prerequisite satisfies the Phase-4 admission gate: HexBareiss 4, HexDeterminant 7, HexMatrix 7, HexArith 7, HexBasic 7, HexPoly 4. No open owning rank implementation/performance issue was found; #10341 and #10346 are downstream consumers, not substitutes.

The compiled driver `bench/HexRank/Bench.lean`, `hexrank_bench` Lake target and existing CI bench list/verify route already exist. It has 18 integer parametric registrations and 11 polynomial fixed registrations. This is not a blank benchmark implementation.

Concrete missing evidence/gaps:

- `reports/hex-rank-performance.md` is absent. The existing `reports/hex-rank-carriers-performance.md` measures fresh-module tactic proofs in the companion; it does not establish HexRank's compiled Phase-4 verdicts or profiles.
- The bench file expressly says the SPEC's external comparators are “not yet wired”. There are no persistent FLINT integer/rational and SymPy polynomial benchmark adapters or recorded comparator curves in this driver.
- The existing integer ladders stop at 64 (dense/variable-rank) or 128 (fixed low rank), while the SPEC/manifest family contract names 16 through 256. Polynomial support includes DensePoly Rat up to 12 but multivariate fixed cases only through 8. Justify actual complete scientific schedules against the SPEC; the small CI smoke range is not the scientific range.
- The polynomial block does not yet supply all three separate producer-first-pass, complete-certificate and checker measurements on both full-rank and deficient cases as required by the SPEC's per-family attribution contract. Its generic six-second timeout has no recorded comparator-derived operation-specific performance budget.
- The module introduction still labels variable-rank deficient scaling `n³`/mode 1, while the actual declarations use the one-sided Hadamard bound. Reconcile stale descriptions and audit model assumptions against actual generators before measurements. For example, dense inputs are L*U products rather than matrices whose entries are individually bounded by 8; the literal `log B≤3` commentary does not describe those product entries.

[...]

PR #10354, opened 2026-09-20
bench(rank): separate polynomial preparation and complete operation coverage

kim-em:
Polynomial checker benchmarks reconstructed their certificates inside the timed callable despite describing that work as preparation. Store validated polynomial matrices and certificates in a per-child runtime cache initialized by `warmupFirstIter`; timed checks now reuse the certificate. Complete the first-pass/certificate/checker grid for both carriers, full and deficient rank, at dimensions 4, 8 and 12, preserving existing case names and adding expected-result hashes.

Add the headline report's API/evidence inventory and explicit outstanding gates. Correct stale integer model descriptions without changing their declarations. The existing integer families still require a strongest-mode audit: triangular/identity pivot structure can make the generic Hadamard bound an unjustified choice of mode. Polynomial fixed cases remain registration coverage until operation-specific SymPy budgets are established; the child timeout is only an operational safeguard. HexRank remains at phase 3.

Validation: `lake build hexrank_bench`; `hexrank_bench list` and `verify` (30 smallest-rung smoke cases; full 54-case verification also retained); copyright/line-count and DAG checks; Mathlib-free import check (54 executables and 2463 proof probes); `git diff --check`. Retained isolated timing checks expose the six-second cap's inclusion of preparation at multivariate dimension 12; the 30-second check passes all repeats, and the final operational cap is 60 seconds for preparation/warmup headroom, without discarding the original samples. No production code, proofs, SPEC, AGENTS.md, PLAN.md or workflow changes.

Refs #10352

The report explicitly distinguishes direct polynomial first-pass compilation (array implementation) from the public producer wrappers (reference implementation), whose missing compiler-replacement import is a remaining #10352 runtime obligation. Default CI checks the smallest polynomial rung while all dimensions remain available through explicit selection.


kim-em:
Fresh Opus review completed. Accepted the compiled-path finding and corrected the report: public producer wrappers use reference elimination, while direct polynomial first-pass targets use ReduceImpl. The missing import remains explicitly tracked in #10352; its isolated AB/BA experiment retains all six pairs and establishes no speedup claim. Accepted the smoke-cost and cap concerns: default verify now selects dimension 4 for each polynomial carrier/rank/operation (30 total checks pass), while explicit polynomial verification retains all 36 scientific cases; the final operational cap is 60 seconds. Clarified which source hash covers the 30-second measurements and disclosed the unrecorded CPU number in the six-second diagnostic. No Phase-4 completion is claimed.

PR #10345, opened 2026-09-20
feat(interval): admit the checked dyadic API through Phase 1

kim-em:
Admit the existing checked dyadic HexInterval API through Phase 1 and add public-import conformance for the exact-coefficient consumer slice. Rational singleton division, intersection, Horner add/multiply, regularization and zero-cut inspection are usable without an experimental backend. The new tests also distinguish precision, endpoint, alignment and product-growth refusals.

The [supported-contract audit](https://github.com/kim-em/hex-dev/issues/10334#issuecomment-5749565268), [declaration inventory](https://github.com/kim-em/hex-dev/issues/10334#issuecomment-5749585795), and [coordinator admission decision](https://github.com/kim-em/hex-dev/issues/10334#issuecomment-5749600437) establish the Phase-1 disposition. `PLAN/Phase1.md` permits correct implemented bodies with explicit owned deferrals. Only `HexInterval.done_through` advances from 0 to 1. This PR does not claim Phase-2 review, full-library conformance/performance, companion admission, or implementation of the following deferred contracts.

| SPEC contract | Current implementation / deferral | Follow-up owner |
| --- | --- | --- |
| Required provider boundaries; large-argument elementary functions; reusable series: bounded precision-driven Machin π and exp(1) generation, literal replay and outward width guarantees | Fixed source-family experiments exist; reusable bounded providers remain absent from the supported API | #10334 runtime; #10342 real soundness/effective convergence |
| Release default policy: deterministic bounded scores, canonical choices, fairness and observations | Generic `Policy` and authenticated `RuntimeController` are supported; staged/adaptive defaults remain experiments, with no production default selected | #10334 retains deferral accounting; no implementation issue owns this later design layer, and the coordinator owns any broader dispatch |
| Search-selected terminal/split correlation and retained proof integration | Generic bounded `Search.Result` and typed runtime events exist; production autonomous split/terminal integration remains deferred | #10334 retains deferral accounting; no separate runtime/companion implementation owner is assigned for this later layer; coordinator dispatch |
| Concrete fact/equality/policy/derivation representations and measured storage | `FactRef`, `EqualityOrigin`, `PolicyFeature`, `PolicyEvent`, `ProofPlan` and related SPEC sketches are explicitly nonfrozen or represented by current namespaced generic APIs; experimental layouts are not promoted | #10334 tracks deferral accounting; these are not artificial missing stubs |
[...]

PR #10357, opened 2026-09-20
feat(interval): add bounded pi/e generators and literal replay

kim-em:
Stacked on #10345 for review; retarget to upstream `main` after that admission tranche merges. The working branch remains `issue-10334`; `issue-10334-providers` is its review branch.

Named-constant consumers currently have only fixed source-family experiments. Add reusable Mathlib-free π and exp(1) generators and literal certificate replay through the public `HexInterval` umbrella.

`Hex.Interval.Constants.generate` accepts an approximation order; `enclose` supplies the deterministic `k+4` schedule. `check` authenticates the requested source and width, recomputes the exact rational center/remainder and outward dyadic cuts, and checks the actual width ≤ 2⁻ᵏ. Order, integer bits/work, allocation, endpoint/precision and replay limits are explicit; insufficient order or resources return failures. Source identities are `piMachinV1` and `expOneTaylorV1`.

The formulas match #10342: Machin's identity with geometric arctangent radii, and the factorial sum with symmetric radius `(n+1)/(n!*n)`. Rounding uses Core precision `+(k+2)`, whose grid size is `2^(-(k+2))`. The source/remainder helpers are exposed for the companion's proofs, while the checked entry points own resource admission. Computed witnesses are checked against the bit charge and reversed cuts are rejected after comparison preflight. Core proofs authenticate source, precision, positive order, exact approximation, ordered cuts, requested dyadic width, exact interval view and retained arctangent power; real containment and effective convergence remain companion-owned. A constructor-view theorem in `Canonical.lean` lets replay expose its literal cuts through ordinary public imports without unfolding the sealed constructor.

Conformance covers both sources at seven precisions through 1000 bits, malformed certificates and independent resource failures. The compiled fixture stream is cross-checked with Python Fraction/direct-power/factorial computations; independent π containment uses `4*(atan(1/2)+atan(1/3))`. Extend the existing oracle and benchmark CI paths, with no additional jobs. Four fixed 1000-bit producer/checker hash anchors and all raw samples are included in [the report](reports/hex-interval-constants.md); they are acceptance evidence, not a Phase-4 complexity claim.

Validation: builds of `HexInterval`, public arithmetic/constant/core conformance, `hexinterval_emit_constants`, and `hexinterval_constants_bench`; the filtered CI oracle and bench scripts; 44 CI-infrastructure tests and three oracle mutation tests; DAG, copyright, line-count, Mathlib-free bench and diff checks.

No phase counter advances here. #10334 remains open for formal core resource/rounding contracts, applicable full-library review/conformance gates, and Phase-4 complexity/comparator/attribution evidence. #10342 separately owns real soundness, effective progress and ordinary-kernel proof measurements. Generic registry scheduling, Horner ownership and unrelated solver/provider development remain outside this change.

Refs #10334


kim-em:
Fresh independent Claude Opus review found no landing blocker and confirmed the source formulas, conservative bit estimates, independent enclosure oracle and CI wiring. The follow-up in 8759539a0 addresses the substantive defensive findings:

- Preflight and reject reversed cuts, prove `checked_ordered`, and strengthen `checked_view` to exact finite bounds.
- Check computed rational witness sizes against the declared bit charge; test negative radii, oversized computed witnesses and one-unit budget deficits.
- Document work/allocation values as conservative predetermined charges, and `limitsFor` as a request-scaled convenience requiring independent caps for untrusted requests.
- Move the stdlib oracle to the Fraction group and record reproducible measurement provenance. bbbc615db retains a new acceptance run of the corrected executable and preserves all initial samples.

The remaining outward Rat inequalities are explicitly retained as core proof work on #10334. Exact replay intentionally recomputes the source and makes no claim to be cheaper than generation. The versioned fixture identifier is retained to make schema identity explicit. The unchecked arctangent helper remains restricted by its documented source contract (5 or 239); generalized analytic lemmas must assume q≥2. These two low-priority suggestions do not identify a defect in either checked source.

The focused correction review is running. Relevant builds, all 14 independent oracle fixtures, mutation tests, DAG and diff checks pass after the changes.

kim-em:
The focused correction review confirmed the preflighted ordering defense, exact-view proof and unchanged valid outputs. Its medium proof-surface concern is now addressed: checked_width proves upper-lower ≤ Dyadic.ofIntWithPrec 1 bits, following finished_width, the authenticated endpoint equalities, exact singleton views and view_subWithin_ready. All relevant public-import builds pass without proof placeholders. Helper admission limits are now explicit in the SPEC/docstring, and the resource regression comment no longer claims to observe execution timing. Other suggestions were checked against evidence: the existing billion-bit test already uses fixed limitsFor 32; the oracle completes in about one second locally, so speculative scheduling changes are unwarranted; and the reconstruction patch was applied in a fresh temporary directory and its exact SHA-256 was verified. The shared integerBits error intentionally covers bit-budget violations, and the small local consistency proof is retained. Outward Rat-cut inequalities and formal resource adequacy remain clearly tracked follow-on obligations.

kim-em:
Synced the admission CI correction and moved ConstantsConformance to an appended Lake glob, preserving all upstream registrations. The timeout test now accepts both documented terminated process states (Z and X); it continues rejecting live descendants. Provider builds, DAG and diff checks pass. The only additional provider edit clarifies the benchmark hash payload: it includes precision, order, rational witnesses and cuts, while each registration fixes the source identity. No measured executable definition changed.

kim-em:
The user explicitly removed transcendental support from the #10143/#10331 family. This interval/provider work is no longer required by that effort. Closing without merge; preserve branches and local proof/measurement artifacts. This is a scope cancellation, not a correctness judgment. Do not resume under the family coordinator.

Issue #10358, opened 2026-09-20
Implement rcf with real algebraic and certified constant coefficients


kim-em:
## Outcome and current state

Complete the optional real-coefficient rcf adapter under #10331 and #10143. Follow the [owning SPEC](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/HexRCF/SPEC/hex-rcf.md#planned-real-coefficient-extension) and [shared execution contract](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/real-closure-execution.md), revised by merged PR #10365.

The base handler registry (#10359), source-schema tranche (#10368), and shared RealFormula frontend (#10338) are already merged. Preserve their typed recognition, transactional dispatch, original-target checks and valuation-parametric equivalence. Registration and schema preparation do not themselves implement a numerical real-coefficient solver. Authentication, coefficient specialization, numerical production, replay, soundness and final evidence remain below; do not repeat delivered frontend work.

Representations use ordinary total operations, canonical zero, structural DecidableEq and explicit sign, with noninjective interpretation into lawful semantic fields. No fake field instance, fallible arithmetic record, per-operation certificate framework or benchmark import exception. Mathematical equality uses zero differences; literal context and certificate bindings remain exact. Persistent refinement transports dependent contexts and requested live values explicitly and invalidates stale bindings.

The caller supplies each approximation procedure and separate containment and width proofs for the same subject. Universal total search additionally requires relative transcendence over the predecessor field. Finite checked successes provide per-input termination only; finite semantic soundness still requires valid bounds. HexInterval, HexIntervalMathlib, bundled pi/e providers and their analytic/performance work remain out of bounds.

No worker dispatch or monitoring is authorized by this update. Consume actual merged API/proof tranches, not merely closed SPEC or roadmap issues.

## Later milestone gates

| Tranche | Required merged artifacts |
| --- | --- |
[...]

PR #10359, opened 2026-09-20
feat(rcf): add checked transactional coefficient handlers

kim-em:
`rcf` now dispatches optional coefficient handlers only when rational reification returns a typed unsupported-closed-coefficient result. Existing rational proof construction and its false/replay/resource failures remain terminal.

The base `@[rcf_handler]` registry checks monomorphic meta signatures and uses deduplicated `Name.lt` order. Handlers explicitly decline, fail, or propose a proof. Successful attempts retain auxiliary proof declarations; decline/failure restores the full state. Each attempt restores its metavariables, and the base checks proposed proof terms against the original target without specializing existing metavariables. Failed tactics restore the goal list, including on runtime exhaustion.

Direct explicit local equalities (in either orientation) can supply checked closed coefficient witnesses without cancellation; arbitrary parameters and nonclosed/cyclic aliases remain rejected. The original target remains unchanged. Proposed proofs are checked for transitive axioms outside the standard three.

Regressions cover explicit and compound aliases, retained zero-divisor syntax, auxiliary proof declarations, rejected sorry proof candidates, existential/Ioc dispatch, imported registrations, ordering/deduplication, malformed signatures, decline/failure restoration, wrong and ill-typed proofs, unresolved proof metavariables, target mutation, terminal false/exhausted/replay failures, rational-normalizer exhaustion, and rational compatibility. Fresh-module generated theorems have only the standard `propext`, `Classical.choice`, and `Quot.sound` axioms.

Validation: `lake build HexRCF HexRCF.ReifyTests HexRCF.HandlerTests HexRCF.LintTests HexRCF.Conformance HexRCF.CertificateTests HexRCF.DecisionTests`; `lake build HexRCFRealFormula` against the merged shared frontend; DAG/umbrella checks and their 12 tests; release manifest/trust-surface checks and 108 release-manifest/sync tests; copyright/line-count checks; 18 relative link targets; `git diff --check`.

This is the complete base-registry tranche. It adds no real-coefficient solver or family import; the optional adapter and its remaining prerequisite gates stay on the open issue.

Refs #10358. Coordinated by #10331.


kim-em:
Fresh Opus re-review confirmed the earlier correctness fixes. Follow-up changes at 734a050b3:

- Fix let-bound alias abstraction with an explicit real lambda; add a direct regression.
- Add a fresh `Elab.async true` module exercising auxiliary proof declarations and the transitive axiom walk.
- Reject-test both an auxiliary theorem hiding `sorryAx` and an unapplied compiler-trusting axiom candidate, without native proof computation. Diagnostics identify the constant introducing the forbidden dependency.
- Cover aliased non-rational endpoints remaining terminal.
- Document that successful Core environment/log/info/kernel-task changes survive, while failed/declined attempts roll back.

I retain the `Expr → MetaM HandlerResult` interface. The reviewer inferred that adapters would have to duplicate alias recognition, but `Reify.closeCoefficient?` is already public; the authorized source-schema continuation calls it directly. A handler must scan the entire original goal for all coefficients/guards anyway: passing only the first recognition failure would not supply that inventory. The SPEC now explains this reuse explicitly.

Validation: affected Lake handler/rational modules passed, including async ordinary-kernel theorem and axiom audits; DAG, release manifest/trust audit and 37 structural tests passed. CI runs on the final revision. No real-coefficient solver is claimed by this tranche.

PR #10360, opened 2026-09-20
bench(rank): wire exact-domain comparisons and separate producer stages

kim-em:
HexRank's driver used identity-block fixtures for some declared growth families, omitted producer/checker combinations, and had no persistent external comparison path. The public producer also compiled without the existing array implementation's compiler replacement, while direct polynomial benchmarks used it.

This tranche fixes that import, validates ranks/certificates/shifted pivots during prep, extends integer schedules through 256, and adds separate certificate-assembly, checked-producer and integer-witness targets. FLINT Int/Rat and exact-domain SymPy polynomial comparisons share actual Lean-generated inputs; preparation stays outside timing, replies are checked, and six alternating AB/BA blocks retain every result. Default CI smoke selects the smallest canonical cases. No workflow or phase bump is introduced.

The report retains partial scientific evidence, the complete four-family profile analysis, the import-only AB/BA comparison (no speedup claimed), protocol overhead, registration checks, and the original signal-floor failure plus its passing longer-batch follow-up. Remaining gates are explicit: complete/consolidate scientific and comparison schedules, establish polynomial operation budgets, and cover native polynomial-quotient witness production. HexRank remains at Phase 3.

Performance rationale: bounded random dense entries replace the unit-triangular product; dense deficient factors replace identity pivot blocks. Hadamard minor bounds plus quadratic arithmetic give the independent one-sided n^5(log n+3)^2 / n^5(2 log n+5)^2 bounds for growing rank. Fixed rank/bits retain the stronger two-sided n^2 model; isolated second-pass work is constant in n. Polynomial anchors do not yet claim absolute-budget verdicts.

Validation: `lake build hexrank_bench HexRankMathlib hexrank_emit_fixtures`; 84 default smoke checks; 24 additional polynomial attribution checks; 28 named external anchor checks; persistent-protocol regression tests; all 282 emitted conformance records accepted by the existing carrier oracle; dependency DAG, copyright/line-count checks and `git diff --check`. Timed-region profiles classify over 90% of samples and pass calibration/sensitivity. CI and fresh Claude Opus review are required before squash/auto merge.

Refs #10352


kim-em:
Addressed the fresh Claude Opus review in 257071955:

- Fail on malformed flattened certificates or false native checker results; reuse the certificate validated during prep rather than recomputing it.
- Validate parametric rank/Boolean output hashes and retain malformed/missing exports as failed command records while continuing the schedule; add regression tests.
- Extend the existing CI job with rank protocol tests and a generated-C check that public producers call the proved array reducer; extend persistent-comparator warmup coverage to the rank adapter.
- Preserve raw metadata/source patches byte-for-byte and provide a reproducible collector with separate retention manifests. Verified all historical source hashes against their revision blobs; added the omitted Lake-file hash supplement.
- Document the paired polynomial compiler-route slowdown (all six after samples slower), wrapper reconstruction in producer/checker ratios, native witness self-check costs, exact external verification subset, and subprocess CPU placement. Keep the slowdown and remaining evidence as explicit Phase-4 concerns.
- Record environment overrides in future profile manifests and support invocation from bench/.

One review inference was incorrect: `hexrank_attribution_bench` is a renamed byte-identical copy of the normal `hexrank_bench` target, not an uncommitted build target. Both files hash to `499475eb6ee476c23ff9855d422fd9b5fec05576996307d43c9ca01eabc8932f`.

Validation: `lake build hexrank_bench HexRankMathlib` (3151 jobs), all 84 smoke registrations, comparator protocol and runner/lint unit tests, 643 persistent warmup registrations, generated-C route check, dependency DAG, copyright/line-count checks, and `git diff --check` passed. Scientific schedules remain retained and ongoing; no Phase-4 bump or issue closure in this tranche.

PR #10369, opened 2026-09-20
bench(rank): retain comparator curves and operation budgets

kim-em:
Refs #10352. Coordinator: #10331.

Completes the retained integer and polynomial-ring performance evidence: all 48 integer paths satisfy their declared evidence modes, all 60 polynomial fixed cases meet operation-specific SymPy-derived absolute ceilings, and all 156 exact-domain rank comparisons have six adjacent alternating AB/BA blocks. The report includes every rung, ten censored pairs, per-worker protocol controls, coefficient-growth records, and four reproducible comparator plots. All comparators remain informational.

Adds eight compiled quotient-witness targets with isolated rational preparation, modular completion, and checking. Seven large-ladder verdicts are consistent with their independently declared cubic model. The full-rank checker remains inconclusive slower; its profile and sole unchanged rerun (also inconclusive) are retained. **HexRank stays at Phase 3 and this issue stays open.** No rank algorithm or theorem is changed in this tranche.

Extends the existing CI job's focused tests and preserves the existing smoke route. The quotient helper is explicitly build-only and Mathlib-free. Raw exports, commands, versions, source reconstruction patches and losslessly compressed stdout are retained with hashes.

Validation: `lake build hexrank_bench HexRankMathlib hexrank_emit_fixtures`; all 92 native smoke cases and 24 new external stage cases pass; 39 focused Python tests pass; complete comparison/output/budget analysis passes; DAG, Phase-4 manifest and release-manifest checks pass; generated compiler-route guard and `git diff --check` pass. The full Mathlib-free scan passes: 55 bench executables and 2473 proof probes. Fresh Opus review follows immediately while CI runs.


kim-em:
Addressed the substantive fresh Opus review findings in `eed641caf` and `a594570de`:

- Re-anchored every archived snapshot to merged main `ff87b54e3`, added `b3360aca1` and `655570d58` (the three resolution verdict rows), and included the input-capture snapshots. `rank_reconstruct.py` applies the recorded source scope and verifies per-run hashes. Both missing follow-up sources were reconstructed successfully. All frozen executable aliases are documented as copies of the same Lake target.
- Added `--budgets` enforcement, exact checks against the retained reference derivation, and a test that changed live references cannot move the frozen ceiling. Added deterministic `--check` paths for analysis, tables and figures, wired into the existing CI freshness step.
- Recorded source/binary hashes, exact commands, and timing for final list/verify outputs: 92 native smoke cases in 1.437 s, and all 24 external stage checks. The earlier outputs remain, with their missing original binary hashes stated explicitly.
- Hardened resume against an unterminated frozen journal and changes to comparator source. Corrected deficient-polynomial support comments, recorded numeric budgets beside registrations, distinguished the weakest three-rung integer evidence, and removed duplicated quotient model prose.

On the checker finding: the report now explicitly compares `finishFull` (which includes checking) with standalone `checkFull`, and records their discrepancy and concurrent-run context. I agree this prevents attributing the inconclusive result to an established super-cubic algorithmic path. I do **not** treat separately scheduled wall times as a guaranteed containment inequality: fresh and cached witnesses can differ in layout/ownership, and host conditions differ. Therefore the containing target does not supply a passing standalone checker verdict, and no completed sample is discarded. Phase 3 and the open Concern remain.

The optional augmented-rank observation is documented: its expected rank is not independent certificate validation; preparation already checks the original rank and exact certificate identities. The representation concern does not affect the current protocol: both carrier constructors build dense DomainMatrix values, and all measured operations preserve that representation. I left the measured timed path unchanged.

Local verification: Lean build, native/stage verification, focused tests, source reconstruction/hash checks, exact budget analysis, four figure freshness checks, table freshness, and diff checks pass. CI is running on the updated head. The separate tail-accumulator experiment is not included in this PR.

Issue #10375, opened 2026-09-21
Implement effective Sturm–Tarski computation and algebraic correspondence


kim-em:
## Outcome and completion boundary

Finish the effective shared polynomial and Sturm–Tarski computation, its algebraic correspondence, and validation that can be completed with the existing foundations. **Completion of this issue must not wait for Tau Ceti.** The root-sum semantic bridge and only the results genuinely dependent on it are owned by #10389. Both issues remain under #10331 / #10143; together they preserve the original assignment's obligations.

Mathlib companion work is not automatically deferred. Keep all available algebraic proofs, finite-sum lemmas and kernel evidence here. The SPECs retain their full mathematical requirements; this issue has a narrower delivery boundary, not weaker correctness claims.

Governing contracts, revised by PR #10365 (merged as ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c):

- [Shared execution contract](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/real-closure-execution.md).
- [hex-poly](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/HexPoly/SPEC/hex-poly.md) and [hex-poly-mathlib](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/HexPolyMathlib/SPEC/hex-poly-mathlib.md).
- [hex-real-roots](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/HexRealRoots/SPEC/hex-real-roots.md) and [its companion](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/HexRealRootsMathlib/SPEC/hex-real-roots-mathlib.md).
- [hex-sturm](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/Libraries/hex-sturm.md) and [hex-sturm-mathlib](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/Libraries/hex-sturm-mathlib.md).

## Merged inputs and ongoing work

[...]

PR #10379, opened 2026-09-21
feat(poly): interpret operations through noninjective coefficient maps

kim-em:
Noncanonical coefficients can have distinct stored representatives with the same value. The shared polynomial path needs interpretation of its actual algorithms without a field instance or an injectivity assumption on those representatives.

Add `Hex.DensePoly.Interpret.map` with a proved `@[csimp]` replacement using `Array.map`, and operation-only transfer for arithmetic, derivative, Horner evaluation, division, gcd, xgcd, xgcdLeft, powers, and monicization, preserving the existing size-derived bounds. Compose these proofs with `HexPolyMathlib` to identify both division outputs, the associated gcd, and the actual Bézout identity. Zero-difference checks characterize semantic polynomial equality. Ordinary-kernel and compiled regressions use a zero-normalized representation of `a+bX` at `X=1`, including unequal representatives, leading cancellation, noncanonical nonunit divisors, zero division, and semantically monic coefficients that are not literal one.

This is the polynomial interpretation slice of #10375. Pseudo-division/gcd generalization and transfer, the shared Tarski producer/replay, both query frontends, backend/denominator agreement, and their validation remain under the issue. Abstract query semantics also require the actual Tau Ceti foundation; roadmap targets are not imported proofs. No Phase-4 completion is claimed.

Validation: full `lake build` (14,509 jobs), `lake build HexPoly.Conformance`, and generated C confirms compiled clients call the array implementation; targeted core and companion builds and proof probes; DAG, release-manifest, benchmark-import and whitespace checks; 108 release-tool and 80 freshness unit tests. Factorization and graph-isomorphism freshness checks pass; exact-blob exemptions document why the added interpretation API and tests do not change the measured factorization executable. The tests are registered in `HexReleaseTests` and the split-release manifest; published CI explicitly builds their generated test targets.

Refs #10375.


kim-em:
Claude Opus review completed using the second-opinion skill. I checked the findings against source and made these changes:

- Published hex-poly and hex-poly-mathlib CI now explicitly builds their generated regression targets.
- Added direct negation, quotient and modulus transfer lemmas, plus the companion modulus and negation bridges.
- Added a general semantic monicity theorem and documented the distinction from the literal `Monic` witness required by `divModMonic`/`modByMonic`.
- Added ordinary-kernel and compiled division probes with a noncanonical, nonunit leading coefficient, expanded axiom guards, and removed an unused fixture lemma.

Two suggestions did not warrant the proposed change. The map deliberately retains its zero-reflection proof so it can preserve the stored length without normalizing through `ofCoeffs`; compilation uses the proved `Array.map` replacement. Also, the review grouped `xgcdLeftMonicAux` with structurally monic division, but that helper actually uses general `divMod` and takes no `Monic` witness. The documentation therefore identifies only the APIs with that literal requirement.

This review covered the polynomial interpretation slice. The remaining pseudo-remainder, query/replay, frontend, semantic foundation and Phase-4 obligations remain tracked in #10375.

PR #10381, opened 2026-09-21
feat(poly): share scaled pseudo-division and plain pseudo-gcd

kim-em:
Signed remainder chains need fraction-free polynomial division with an explicit multiplier and a plain gcd over the coefficient domain. The existing pseudo-division implementation lived under HexResultant, outside the shared polynomial owner.

Move that exact array recurrence and its reconstruction/size proofs to `HexPoly.PseudoDiv`, retaining `HexResultant.Basic` as a compatibility import. Add recorded-scale division, positive sign correction of all three outputs using leading-coefficient sign and exponent parity, and plain pseudo-gcd with proved termination by the second input's stored size. No coefficient division, caller budget, Bézout accumulator or `Field Int` instance is introduced.

Generic noninjective interpretation follows the actual power table and cancellation, quotient and remainder folds. The companion proves reconstruction, nonzero/positive multipliers, strict remainder bounds, agreement of both outputs with scaled field division including zero divisors, and the common-divisor characterization of pseudo-gcd in the semantic field. Integer and noncanonical-coefficient probes exercise the production APIs in the ordinary kernel and compiled code; kernel proofs rewrite array folds using core lemmas without changing the executable loops.

Validation: full `lake build` (14,516 jobs); targeted core/companion builds and axiom guards; HexPoly and HexResultant conformance; 57 independent FLINT results (36 new pseudo-arithmetic results), with deliberate multiplier corruption rejected; cases include negative leading coefficients, zero/constant inputs, `(2,X)`, exact division, defective degree drops, repeated factors, and rational backend agreement; DAG, release-manifest, copyright, file-length, benchmark-import and whitespace checks; factorization and graph-isomorphism freshness checks. Exact-blob exemptions cover only the existing factorization measurement, whose code does not call these new APIs.

This is the shared pseudo-arithmetic slice of #10375. The Tarski producer/replay, query frontends and their semantic theorems and backend agreement remain outstanding, along with the required Phase-4 measurements. No issue completion or new performance result is claimed.

Refs #10375.


kim-em:
Claude Opus reviewed the pseudo-arithmetic slice independently and found no correctness or soundness bugs. I verified the findings against the source and made these changes:

- Positive sign correction now tests the leading coefficient only when the recorded exponent is odd. Its companion proves that this branch is equivalent to negativity of the interpreted multiplier. This avoids a sign query on the computed power while retaining the existing pseudo-division recurrence.
- Reconstruction, remainder correspondence and both-output division correspondence now include zero divisors. The universal integer and noninjective proof probes also cover this case.
- Added 36 independent FLINT pseudo-arithmetic comparisons, including exact multipliers, both quotient/remainder outputs, zero inputs, both parity cases for a negative leading coefficient, defective degree drops, and nonconstant fraction-field gcds. All 57 emitted results pass; the oracle rejects an intentionally corrupted multiplier and checks integrality without truncation.

I retained the existing pair-returning recurrence and its compatibility import. Reusing its internal power table would require changing that interface/proof boundary; the implemented sign optimization addresses the expensive sign query now. The duplicate scalar power computation remains an optimization opportunity, with no performance claim made. I also retained the actual-fold interpretation proofs: their higher elaboration budget is local, and they do not add runtime computation.

The original issue remains open for the query/replay and frontend work, semantic proof gates and required Phase-4 evidence. Runtime oracle agreement is not a substitute for companion proofs.

kim-em:
The focused Opus follow-up found no correctness or soundness bugs and confirmed that the three targeted findings are resolved. Full local rebuild also passed (14,516 jobs).

The optional remaining suggestions are recorded rather than claimed complete: sharing the internal power table is still a possible performance optimization; the positive-multiplier theorem retains its nonzero-divisor premise, which matches every signed-chain division; and the oracle's impossible non-integral expected-output path fails loudly before the normal mismatch-record writer, as existing malformed-oracle paths do. None changes the proved correspondence or the accepted executable inputs. New Phase-4 performance evidence is still outstanding on #10375.

PR #10385, opened 2026-09-21
feat(sturm): add ordered-field queries and domain correspondence

kim-em:
This adds the ordered-field frontend to the shared Sturm–Tarski kernel. Queries support finite endpoints and both infinities, reject exactly the mathematical domain failures, and normalize remainders by a positive absolute leading coefficient. A private `Prepared` value retains the validated squarefree chain so repeated queries reuse it. Noncanonical representations use ordinary operations and interpretation laws, with no field instance on their storage.

The companion proves exact domain equivalence, exact prepared input bindings, prepared-domain validity, ordinary/prepared certificate acceptance, certificate/query value agreement for arbitrary literal contexts, and exact bindings and domain validity of accepted replay. Positive denominator clearing is exposed from the existing integer-polynomial implementation, with exact polynomial correspondence; integer/dyadic endpoint guards now have their complete mathematical interpretation. Universal rational/integer validity agreement after positive clearing is proved by `query_rat_domain`, including invalid heads and root endpoints on finite ordered dyadic intervals. Whole-Option value agreement is exercised by conformance, but its universal successful-value theorem remains open.

The benchmark cost-model derivation uses quadratic summed dense work for normal Chebyshev chains and their endpoint/replay passes. Initial constant-query reduction and fixed-denominator clearing are linear; for a fixed quadratic head, the shared pseudo-division recurrence has at most two correction summands per output coefficient, giving linear work in query degree. These are coefficient-operation models, not uniform bit-complexity promises. For Fibonacci pseudo-gcd, the recurrence performs 3*m+7 scalar multiplications per nonconstant divisor degree m, plus six in the final constant step; their sum is 3*n*(n+1)/2+7*n+6. The extended ladder distinguishes that lower-order term without changing the quadratic model.

Validation includes:

- Full local `lake build`: 14,546 jobs passed, plus the complete `HexConformance` target (13,808 jobs) and focused field/integer companion builds.
- Infinities, invalid domains, common factors, negative leading coefficients, prepared reuse, noncanonical coefficients, forged squarefree witnesses, wrong initial quotients and exact context bindings; 36 positive-clearing whole-Option comparisons.
- Fresh ordinary-kernel acceptance, corrupted-identity and stale-context rejection, plus domain interpretation. These modules use the existing `HexConformance` target; no additional proof-test library is introduced. Only standard Lean axioms occur.
- Thirteen Mathlib-free query-stage registrations and two registrations in the existing polynomial benchmark, independently checked exact fixtures, four-trial scientific schedules, adjacent alternating backend comparisons, and profiles restricted to operation regions. All raw evidence is retained in `reports/hex-sturm-performance.md` and its linked artifacts.
- Eleven of fifteen new scaling registrations are now consistent. The fixed-field pseudo-gcd mismatch is resolved by widening its degree ladder through 1024, preserving all original rungs, its quadratic model and the algorithm. Four query registrations remain unresolved after the single unchanged rerun (one has mixed pass/inconclusive evidence). Retained diagnostics identify the actual 31-bit small-Int boundary and coefficient/identity products crossing it. The unresolved cases are explicit remaining performance obligations, not passes; no phase advancement or complete Phase-4 claim is made.
- Existing polynomial and query regression modules are now wired into their standard conformance modules; no new test library is added.
- DAG/source/release checks, existing performance-data freshness and the benchmark smoke gate. CI extends the existing single job.

Refs #10375. This is a further implementation slice. Actual Tau Ceti polynomial IVT/Rolle and signed-remainder/Cauchy-index delivery is still needed for the abstract root-sum and semantic replay theorems. Those proofs, count/singleton/sign/bound results, universal whole-Option backend agreement, the four unresolved query scaling checks and the remaining Phase-4 axes remain open. Runtime agreement does not discharge companion proofs.


kim-em:
Fresh Claude Opus review and focused follow-ups found no correctness or soundness defect. The final implementation is on `256c251ec1fbabbabc513c3c8b7629583823d4fd`.

Findings addressed:
- Shared the literal fixture in a Mathlib-free conformance module; added independent false-squarefree, wrong-initial-quotient and genuine nonconstant-terminal-gcd rejection tests. The last test passes every chain identity and isolates the constant-tail guard.
- Added exact preparation bindings, prepared-domain validity, prepared-certificate acceptance, certificate/query value agreement for arbitrary literal contexts, and all six accepted-replay bindings. Ordinary-kernel axiom guards cover the frontend proof API.
- Proved universal rational/integer `isSome` agreement after positive denominator clearing on finite ordered dyadic intervals. Opus checked the coefficient interpretation, unit scaling, endpoint conversion and invalid-head/root-endpoint cases. Successful-value agreement remains unproved.
- Corrected the 31-bit threshold wording and measurement provenance; aligned registered input-family names with the SPEC and emitted fixtures; documented denominator-clearing ownership and the prepared certification API.
- Retained every completed measurement. The fresh proof measurements explicitly identify their earlier source/import graphs; the current companion umbrella reaches the pre-existing full Mathlib import through `ChainCorrespond`. No current-import-cost or improvement claim is inferred from those runs.

Review suggestions assessed but not adopted:
- Moving unrelated SPECs or narrowing the pre-existing `ChainCorrespond` import would broaden this PR. The import consequence is documented.
- The real-cast clearing specialization is now private to the frontend composition instead of adding another public helper to an earlier library.
- The integer certificate/query value link already exists in `HexRealRootsMathlib.Query.integer_certify_checks`; the new arbitrary-context frontend links need no expansion of the earlier libraries.
- Equal build-job counts do not establish identical source graphs or identify the cause of a timing difference. The report makes no causal performance claim from the two fresh-module snapshots.
- The measurement harness already rejects missing/mismatched axiom output. It was left unchanged; guarded CI prints and deliberate bare measurement prints serve their separate checks.

Local full build (14,546 jobs), complete HexConformance (13,808 jobs), focused axiom checks, DAG/source/Phase-4 checks, Mathlib-free bench checks and 24 independent exact FLINT fixtures pass. CI is running on the final head. Existing phase records remain unchanged, and no new Probe library target was introduced. The remaining semantic, successful-value/backend transport and Phase-4 obligations stay in #10375.

Issue #10320, opened 2026-09-20
perf(poly-det-mathlib): make symbolic det faster than norm_det by removing the non-certificate costs


kim-em:
## Objective

Make the symbolic arm of the `det` tactic faster than `norm_det` where its certificate is small, by removing the costs that are not the certificate. On the `N4K2D2S4` probe (`bench/HexPolyDetMathlib/ProofProbe/N4K2D2S4Hex.lean`: 4×4, two atoms, degree-2 entries with four terms each) the full dispatched tactic has a median of 1.50 s against 1.10 s for `norm_det` followed by `ring` (`reports/hex-poly-det-mathlib-performance.md`). Profiling the tactic on that case (`lean --profile -D profiler.threshold=0`, with spans added around the frontend's steps, and re-checking the components of the emitted proof term separately in the kernel) gives this attribution of Hex's own work, beyond the ~0.7 s both arms spend elaborating the theorem statement:

| stage | time (two runs) | note |
|---|---:|---|
| `mkAppM ``Polynomial.identify #[A, B, conjunction]` in `HexPolyDetMathlib/Frontend.lean` | 400 to 635 ms | Meta unification only; the kernel repeats the same unfolding |
| kernel check of the whole auxiliary theorem | 480 to 980 ms | split below |
| producer | 50 to 77 ms | real work |
| reify, convert, quote, entry identification, node count | about 70 ms | |

and of the kernel check (`result._proof_1`, 262,247 nodes):

| component | time | note |
|---|---:|---|
| `checkDetPolyPacked … = true` (the certificate) | 155 to 284 ms | 46,323 nodes |
| the 16 entry proofs (`HexReflectMathlib.Kernel.eval_checked`) | 150 to 224 ms | about 11 ms each, of which about 8 ms is the kernel computing `beq (ringList k e) p` |
| the target entry proof | 106 to 160 ms | the kernel multiplies four quadratics with list arithmetic to compare with the witness value |
| `beq left right` on the value lists | 5 to 10 ms | |

[...]

PR #10386, opened 2026-09-21
perf(poly-det): integrate rational factors and bounded structural proofs

kim-em:
Symbolic determinants with common rational row expressions or sparse structure previously expanded those expressions before proving the determinant identity. This implements the structural routes specified in #10384: generic rational row-factor transport, triangular/zero identities over arbitrary commutative rings, and sparse cofactor proofs with an actual expansion bound of 64 scalar-product leaves. Failed target comparisons restore the original goal and retain the existing certificate/Mathlib fallback.

The public `det`, `det%`, and `Hex.normPolyDet` interfaces remain opt-in. Forced certificate arms bypass the shortcuts. Regression coverage includes different numeric matrices and denominators, opaque factors, arbitrary universes, positive characteristic, row/column permutations, false targets, local operation overrides, and exact budget boundaries.

Validation: full `lake build`; determinant kernel tests; unchanged conformance fixtures and 42 exact oracle records; DAG, Phase-4, release-manifest and Mathlib-free benchmark checks; seven resource-limit harness tests. Proof audits permit only `propext`, `Classical.choice`, and `Quot.sound`.

Related to #10320. The original general tree-certificate attribution targets remain open; these structural wins do not establish them. The SPEC change was merged separately in #10384.

Six adjacent AB/BA pairs on nine fixed inputs, with quiet fresh modules and separate route/axiom audits, show 12.875 s → 3.379 s on the larger rational case and 22.736 s → 7.636 s on the larger sparse case (import-subtracted build medians). All six larger inputs favor Hex in all six pairs; three tiny comparisons remain inconclusive. The entire bounded campaign took 25.95 minutes, including final checks after the operation-admission fix. All samples, including one SIGTERM-interrupted attempt and its bounded recovery, are retained in `reports/bench-results/hex-det-tree/structural/`; see `reports/hex-poly-det-structural-performance.md`.

The final-code six-pair checks retain the gains: 1.619 s → 0.600 s for quadratic rational row factors and 2.305 s → 0.832 s for the sparse example, each with six Hex wins. Tiny triangular timings remain inconclusive.


kim-em:
Independent Opus review found no soundness issue. Addressed its substantive findings in a3772fce9:

- Authenticate rational operation instances before extraction and reject mixed operand types before numeric certification. Added the specific `HDiv Nat Rat Rat` regression (without an `HMul Rat Nat Rat` instance). Kernel/assembly errors remain visible rather than being indiscriminately swallowed.
- Thread the shared reflection configuration through structural assembly/finalization; tiny proof-node budgets now test both decline paths. Restore saved state on finalization decline.
- Add a kernel-checked 6×6 case with two nonzero cofactors and two expansion levels, reversed rational/sparse equalities, sparse term/simproc coverage, and a characteristic-three fallback whose proof is checked to use the residue certificate.
- Remove the namespace-wide heartbeat increase: all regression tests pass at the default limit.
- Clarify separately recorded initial hashes, retained script versions, and failure-path costs in the report. Raw compressed samples were independently checked programmatically for source hashes, pair counts and audit coverage.

The full 14,540-job local build passes after these changes. The review's remaining refactoring suggestions concern duplication and small admission overhead; they do not block this bounded implementation. The general tree-certificate targets remain open.

PR #10387, opened 2026-09-21
feat(sturm): prove rational query agreement by signed remainder comparison

kim-em:
Prove that rational and integer/dyadic Tarski queries return the same whole `Option` after positive denominator clearing. Previously only domain agreement was proved; successful values were covered by fixtures.

The proof extracts remainder equations from the checked initial, recurrence and terminal identities. It shows that arbitrary accepted chains for positively scaled inputs have equal lengths and corresponding entries related by positive factors, then proves equality of finite endpoint signs and variation values. Singleton chains, zero initial remainders and nonconstant terminal gcds are included, with no injectivity assumption on coefficient interpretations and no root-sum theorem.

Audit names and ownership across the shared kernel, frontends, correspondence and clients. The polynomial data are `RemainderStep` and `SignedRemainderChain`; query evidence is `TarskiCertificate`, specialized as `IntTarskiCertificate`. `EndpointSigns.intDyadic` owns integer/dyadic endpoint operations outside `ZPoly`, alongside `EndpointSigns.ofSign` for coefficient-field endpoints. The terminal predicate belongs to `SignedRemainderChain.lastIsConstant`. Modules and correspondence use `Tarski`; helper names now describe their operations (`normalizeContent`, `subIsZero`, `lastIsConstant`, `checkEndpoints`, `nonvanishing`). The field frontend exposes `PreparedDomain` and `Sturm.check`. Current callers, documentation, manifests and proof guards use these names. Recorded historical measurement artifacts remain unchanged.

Validation: full local `lake build` (14,548 jobs), then aggregate libraries, manual, complete `HexConformance` and the Sturm benchmark after review fixes (14,790 jobs); guarded ordinary-kernel axiom checks, dependency-DAG, conformance-target and file-size checks pass. No new axioms, sorries, runtime algorithms, libraries or phase changes.

This supplies whole-Option rational backend agreement and value agreement of arbitrary accepted certificates (`check_rat_value`). Literal certificate translation in both directions remains separate work. Root-sum semantics and their consequences still require the shared foundation; this PR does not claim those or Phase-4 completion.

Refs #10375.

PR #10388, opened 2026-09-21
feat(sturm): transport certificates and prove backend agreement

kim-em:
Supplied Sturm–Tarski certificates can now be translated between integer/dyadic and rational representations without rerunning polynomial division or chain production. Mathlib-free `HexSturm.Transport` clears all literal scales and quotients at finite dyadic endpoints and embeds integer evidence into rationals, retaining infinities. Companion proofs establish acceptance with the corresponding exact input/context bindings and the same value.

Adds whole-Option query agreement under noninjective representations into a common ordered field, positive input scaling, and matching finite/infinite endpoints. Arbitrary accepted certificates may use different valid chains and positive normalizers. These are algebraic results; no root-sum semantics are assumed or claimed.

Validation: full local build/manual/conformance/benchmark build, ordinary-kernel proofs and axiom guards, noncanonical representation tests, independently supplied fractional literal evidence, and computational cases for zero-initial/singleton chains, nonconstant terminal gcds, nontrivial denominators, infinity embedding, and invalid context/value/interval/scales. Fresh Claude Opus review and follow-up found no mathematical correctness defect; placement, documentation and coverage findings were addressed.

Refs #10375. Based on main after #10387; no phase changes or Phase-4 completion claim.


kim-em:
Fresh Claude Opus review of the transport implementation and a follow-up review found no mathematical correctness defect. The follow-up verified that denominator clearing may recompute signs at its explicit finite dyadic interval: the acceptance theorem requires the original exact endpoint bindings, so it does not assert unconditional value preservation for a mismatched interval.

Addressed findings: moved executable translations to Mathlib-free `HexSturm.Transport` under their owning types; kept acceptance proofs in companions; fixed theorem-table and ownership/scope documentation; added computational transport cases for singleton and common-factor chains, nontrivial head/query denominators, infinity embedding, and rejected context/value/interval/scale changes. Companion and computational conformance builds pass. Local universal proofs retain the audited standard axiom set.

The finite dyadic restriction is explicit for rational-to-integer clearing; integer-to-rational embedding retains infinities. Micro-optimization of repeated clearing and the already-namespaced proof helper names are not correctness gates. No semantic root-sum or Phase-4 completion is claimed.

Issue #10389, opened 2026-09-21
Prove Sturm–Tarski root semantics after the Tau Ceti foundation


kim-em:
## Outcome and start gate

Complete the Hex-specific semantic bridge from the shared signed-remainder computation to sums of signs over roots in an arbitrary ordered real closed field. This is the deferred part of #10375 under #10331 / #10143. #10375 can close independently once its effective computation, algebraic proofs and immediate validation are complete.

Start this integration when an actual proved Tau Ceti signed-remainder/Cauchy-index result with the required hypotheses is available for import, together with any foundation it requires. #10300 coordinates the roadmap; closing that issue or approving the roadmap is not this gate. We need the particular theorem, not the whole roadmap, Thom/BKR development, or a constructed real-closure tower. Record the exact upstream module/declaration, revision and relevant hypotheses before implementation.

The contracts remain [hex-real-roots-mathlib, shared foundation and proof ownership](https://github.com/kim-em/hex-dev/blob/main/HexRealRootsMathlib/SPEC/hex-real-roots-mathlib.md#shared-foundation-and-proof-ownership) and [hex-sturm-mathlib](https://github.com/kim-em/hex-dev/blob/main/HexSturmMathlib/SPEC/hex-sturm-mathlib.md). This split changes delivery ownership, not the statements or soundness requirements of those SPECs.

## Required foundation shape

Over an arbitrary ordered real closed field, relate endpoint variation differences to the finite sum of `sign (F.eval α)` over the distinct roots of nonzero squarefree `P` in the specified open interval. Finite endpoints are root-free; either endpoint may be infinite. The signed-index convention must give `P'/P` contribution +1 at a simple root.

The foundation must support, directly or by a Hex algebraic adapter, initial reduction of `F*P'`, positive scaling of remainder identities, zero initial remainder/singleton chains, arbitrary common factors of `P,F`, nonconstant terminal gcds, constant heads and infinite endpoints. Polynomial IVT/Rolle over the abstract field are upstream foundation obligations where needed. Existing derivative-only Sturm results over the reals do not by themselves establish this statement.

## Deliverables after the foundation lands

- [ ] Connect the already-checked recurrence, terminal, degree and endpoint-sign facts to the imported foundation, proving the shared variation/root-sum identity. Adapt an unscaled upstream statement by positive-rescaling correspondence if necessary; reuse the single existing kernel.
[...]

Issue #10376, opened 2026-09-21
Implement ordered rational-function extensions and their proofs


kim-em:
## Outcome

Implement exact ordered rational-function extensions by successive infinitesimals and caller-supplied computable real constants, together with their semantic proofs and required evidence. This is the combined computational/companion assignment under #10331 for #10143.

Governing contracts revised by PR #10365 (merged as ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c): [hex-ordered-fn](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/Libraries/hex-ordered-fn.md), [hex-ordered-fn-mathlib](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/Libraries/hex-ordered-fn-mathlib.md), and the [shared execution contract](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/real-closure-execution.md).

## Deliverables

- [ ] Reuse canonical RationalFn arithmetic over lawful prealgebraic coefficient fields. Add opt-in orders and explicit total sign without conflicting global instances; do not instantiate RationalFn on noncanonical selected-root syntax.
- [ ] Implement infinitesimal signs from the first nonzero numerator and denominator coefficients, comparisons and successive infinitesimals. Prove normalization invariance, order laws and the ordered Laurent/Hahn interpretation specified by the companion.
- [ ] Implement only the minimal exact rational bounds and Horner operations needed by the caller-approximation protocol. Keep computation `approx : Rat → Bounds` separate from requested-width and containment proof functions; both guarantees refer to the same subject, embedding and requests. The library does not supply analytic approximation procedures.
- [ ] Implement the proof-founded total sign and derived successive-level approximation searches. Accessibility/progress proofs erase; execution still performs refinement. Prove progress from containment, shrinking width and relative transcendence over the entire predecessor field, not merely Q. Formal zero remains an algebraic test.
- [ ] Keep optional finite comparisons and successful finite certificate checks separate from total field registration. A finite success witness proves only that search's termination; semantic validity still needs containment. Original expression divisor guards survive cancellation in tactic consumers.
[...]

PR #10391, opened 2026-09-21
feat(sturm): prove supported query counts and correct bit-cost models

kim-em:
The new signed-remainder certificates lacked a bridge to the existing real Sturm root-count theorem. This connects arbitrary accepted integer query-one certificates to finite dyadic-interval and whole-line counts, proves producer nonnegativity, and transports finite counts to rational polynomials. Open distinct-root sums agree with the legacy half-open counts under the checked squarefree/nonvanishing domain. The general ordered-field bridge and dependent count wrapper remain in #10389.

Adds the elementary finite-root-sum lemmas required by the revised #10375, ordinary-kernel literal count checks, and exact axiom audits. Root membership and nonzero-scalar invariance are available through companion lemmas; no replacement root solver or analytic foundation is introduced.

Corrects performance declarations using the actual arithmetic costs. Initial division at query degree m materializes Θ(m²) coefficient bits; initial reduction, integer/rational queries and replay now pass their quadratic tests on multiword inputs. Dyadic replay has Θ(n³) normalizer iterations and Θ(n⁴) binary work. The bounded cubic timing hypothesis passes at degrees 128–1024 but fails the complete extension through 2048 (residual +0.168613). Replay remains an explicit performance finding on #10375. The failed quartic wall-time test and all previous results are retained. The benchmarking-rule correction is a separate commit. No library phase metadata changes.

Validation: aggregate/manual/conformance build, ordinary-kernel checks, dependency/file-size/conformance-target and Phase-4 lints, 13 benchmark smoke checks, and 23 pinned exact FLINT axis/control fixtures. Benchmarks remain Mathlib-free. No axiom, sorry, native_decide or duplicate kernel is added.

Refs #10375. General semantic integration remains #10389; concrete tower integration belongs to #10376/#10378.


kim-em:
Claude Opus reviewed the implementation and corrected cost models through the `second-opinion` skill. The review found no defect in the count/root-sum proofs. The following changes address its substantive concerns:

- The integer and rational query-one results now identify the answer with the open distinct-root sum, using squarefreeness and upper-endpoint nonvanishing to bridge the existing half-open multiset count. Root membership lemmas replace exposing the semantic definitions.
- Query-degree declarations use the independently derived quadratic coefficient-bit volume. Integer query, rational query, initial reduction and replay all pass the same quadratic model on the final ladder. Smaller-ladder failures, the limited replay margin and the host memory ceiling remain explicit.
- Replay distinguishes cubic normalizer iterations from quartic binary work. Exact calculations extend through degree 2048. The first wider timing run hit its process cap in all four largest-rung trials; its survivor-only harness verdict is explicitly not counted as a successful extension. Only the operational cap changed for the complete follow-up.
- The benchmarking-rule correction is a separate commit. Original verdicts are qualified by their finite coefficient regime, and source/binary provenance records remain intact across commits during collection.

I did not infer a tight cubic bound for the full checker from the cubic stored input size: that size alone does not bound growing-integer products. The normalizer finding is recorded on the assigned issue rather than creating a workaround decomposition issue. No library phase changed.

The complete degree-256–2048 extension returns an inconclusive cubic verdict (residual +0.168613). All sixteen measurements finished; the degree-2048 median is 365.525 seconds. This leaves the replay performance finding open on #10375. It is not a Phase-4 pass.

Validation: the reviewed implementation passed the complete one-job CI run 35614560676. The final evidence update is receiving its own CI run. Local aggregate/manual/conformance builds, exact axiom audits, structural/Phase-4 checks, Mathlib-free import checks, thirteen benchmark smoke checks and twenty-three exact axis/control oracle fixtures passed.

kim-em:
The final Opus review independently verified the model derivations, exact work counts and every quoted final timing/result. I corrected its remaining substantive concern: the degree-512 profile identifies normalization work, but does not establish that normalization alone causes the degree-2048 residual. Recurrence verification also performs growing-integer products.

The report now places the measured growth factor **9.97** beside the independently counted **7.97** iteration factor and **16.02** division-bit-volume factor, without fitting a coefficient or exponent. Explicit sums of retained exclusive profile periods give **27.62%** for nine division/normalizer symbols and **33.85%** for seven allocation/copy symbols; the latter have no caller attribution. The sums and their exact symbol lists are retained. I checked the allocation sum directly rather than adopting the review's approximate 25% estimate.

Homogeneous integer Horner evaluation is mathematically sound as a possible local improvement, but would remove the very iteration cost used by the cubic hypothesis. A new model must account for the remaining recurrence products before timing a changed implementation. It should track a power-of-two exponent and shift coefficients, rather than materializing successive denominator powers, which would regress sparse inputs. No such implementation change is part of this model-correction PR. The replay performance gate remains explicit on #10375.

Issue #10377, opened 2026-09-21
Implement BKR/Thom sign determination and checked replay


kim-em:
## Outcome

Implement univariate BKR sign determination, Thom root identity/order and checked replay, including the companion proofs and required performance evidence. This is the combined computational/companion assignment under #10331 for #10143, not a quantifier-elimination algorithm.

Governing contracts revised by PR #10365 (merged as ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c): [hex-sign-det](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/Libraries/hex-sign-det.md), [hex-sign-det-mathlib](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/Libraries/hex-sign-det-mathlib.md), and the [shared execution contract](https://github.com/kim-em/hex-dev/blob/ad0e9f9f8e26318f8e9a79ea81acb2f1c57cbe2c/SPEC/real-closure-execution.md).

## Deliverables

- [ ] Build exact complete sign tables from the shared prepared Tarski-query API. Reuse existing rational matrix/rank/row-reduction infrastructure. Preserve fixed query/row/column order and exact integer counts, including zero and empty cases.
- [ ] Implement BKR recursive support reduction with explicit evidence that every realizable sign condition remains in the candidate support. A square invertible submatrix and M*c=t alone do not justify omitted columns. Replay includes child query evidence, support-completeness derivations and exact matrix identities.
- [ ] Implement squarefree-root descriptors, validated partial/full Thom encodings, equality/order/sign at a selected root, common-product comparisons and re-encoding. Preserve context identity and reject copied/stale derivative vectors or child certificates after refinement.
- [ ] Reuse ordinary total coefficient operations, NatCast and explicit sign on canonical-zero representations. Semantic polynomial equations use zero differences; exact integer/rational matrix identities and literal context bindings retain their actual equality checks.
- [ ] Prove finite linear algebra/support induction and correspondence of the actual algorithms with abstract root/sign semantics. Consume the required Tau Ceti Thom and BKR foundation through the companion; distinguish missing foundation theorems from executable availability.
- [ ] Expose the specified sample-point interface for the real-closure, CAD and coverings consumers, without implementing those clients here.
[...]

PR #10401, opened 2026-09-22
feat(sign-det): check reduced moments and prove algebraic correspondence

kim-em:
BKR moment queries now reduce after each indexed multiplication, using shared positive pseudo-division and normalization. Replay checks the supplied positive-scaled identities and degree bounds before checking the Tarski certificate on the reduced operand; it does not expand the full product or run the producer. The explicit unreduced mode and exponential reference remain available for comparisons.

The new Mathlib companion proves acceptance of the actual reduction producer and sign preservation for arbitrary accepted reductions at every root, including noninjective coefficient interpretations. Early shape/exponent checks reject malformed input before factor-list allocation. Conformance covers reduced/full agreement, high-degree products, malformed chains and scales, mutations of produced trees, noncanonical representations, literal context bindings, and ordinary-kernel proofs with axiom inventories. The companion target extends the existing CI job.

Validation: full `lake build`; `lake build HexSignDet HexSignDetMathlib +HexSignDet.Conformance +HexSignDetMathlib.Conformance`; dependency DAG, copyright, line-count, conformance-registration and Phase-4 metadata checks. No new sorry, axiom or native_decide. Existing unrelated proof warnings remain in the full repository build. Metadata checks and these regressions are not Phase-4 evidence.

Refs #10377. This is a partial tranche: total determine/SignTable APIs and producer completeness, tensor/basis integration, Thom descriptors and selected-root/sample-point operations, initial per-query reduction reused across moments, shared domain/nested evidence and serialization, full independent-oracle conformance and Phase-4 measurements remain. Full semantic proofs still consume #10389 and the separately required Tau Ceti BKR/Thom foundations. The reduction theorems do not claim root-sum soundness or issue completion.


kim-em:
The independent Opus review found no checker/proof soundness hole. Its substantive evidence and documentation concerns are addressed:

- Added exact, identity-consistent forgeries for a negative right scale, a zero left scale and a zero right scale. Their polynomial identities pass while the positivity guards reject the false sign claims. The sign-flipping case is also an ordinary-kernel theorem with an axiom inventory.
- Added reduced/direct comparisons for non-unit positive and negative leading coefficients, a non-unit linear head and a cubic head with high-degree queries.
- Corrected the computational/companion ownership statements and documented the unguarded reduction builder as producer-internal.
- Explicitly recorded that the current producer does not yet share an initial reduction of each query across moment rows. That format/integration work remains part of #10377 and is required before claiming the reduced-degree performance bound.

The optional quotient-size pre-check was not added here: the shared identity checker currently accepts arbitrary finite polynomial witnesses, and this tranche claims neither a certificate-size bound nor Phase-4 completion. The query pre-reduction/evidence work will account for operand and certificate growth. The repeated shape checks remain intentional standalone entry-point validation.

`lake build HexSignDet HexSignDetMathlib +HexSignDet.Conformance +HexSignDetMathlib.Conformance` passes after these changes. Full root-sum/Thom correctness and performance gates remain open as listed in the PR.

PR #10402, opened 2026-09-22
feat(sign-det): reuse reduced queries and tensor inverses

kim-em:
BKR parent nodes now combine the retained child inverses by an integer tensor product and solve with checked exact divisibility. Query polynomials are reduced once before the balanced recursion; all moment rows and child slices reuse their reduced operands and indexed witnesses. Replay checks the preprocessing against each original query list and never expands an unreduced moment on the reduced path.

The companion proves full column rank after pruning, acceptance and column order of the actual rank producer, the right-to-left inverse adapter, tensor correspondence in the literal Cartesian order, its composition with the actual child rank witnesses and retained lists (`Node.product_inverse`), and exact recovery by the scaled solver. It also proves acceptance of query preprocessing and sign preservation for every actual checked node operand under noninjective coefficient interpretation. Computational proofs verify acceptance of child slices and preservation of query lengths and operand degree bounds.

Conformance adds 59 deterministic rational cases checked independently by pinned FLINT qqbar root/sign evaluation and the existing real-algebraic API. Complete sparse-table equality checks every condition, including omitted support with a matching total. The existing CI job runs the emitter, oracle and eight oracle mutation tests. Kernel probes and replay mutations cover the new finite algebra and copied, reordered or altered preprocessing evidence.

Validation: full `lake build`; computational and companion conformance targets; the actual filtered `scripts/ci/run_oracles.sh` harness (59 cases, zero failures; eight Python tests); DAG, copyright, file-size and conformance-registration checks; `git diff --check`. Phase-4 registration checks pass, but no performance gate is claimed.

Refs #10377. Remaining gates include total producer completeness and the validated public table/descriptor APIs; shared domain replay, finite DAG serialization and nested/refined contexts; Thom completion, root order, sign-at-root, common-product comparison, re-encoding and the sample-point interface; all required semantic correspondence through #10389 and the separate Tau Ceti Thom/BKR foundations; broader descriptor/extension-field conformance; and the complete Phase-4 runtime, proof-cost and profiling evidence. Preprocessing witnesses are reused by production, but replay still repeats their checks at tree nodes. The rank performance obligation #10352 remains separate. This PR does not close the assignment.


kim-em:
Independent Claude Opus review found no soundness blockers. I checked and addressed its six findings:

- Added an explicit serialization-order check to the independent oracle, with a targeted assertion, so ordering and incorrect complete tables report different failures.
- Added `Node.basis_matrix`, `Node.basis_support`, and `Node.product_inverse`. These identify the actual retained minor and exact row/support lists used by `buildTreeFrom`, then prove its tensor witness identity using the actual child rank certificates. The theorem assumes checked child systems and their produced rank witnesses, not an accepted parent or root semantics. Its ordinary-kernel axiom inventory is checked in conformance.
- Documented the rank-column and left-inverse dependencies at the producer shortcut. I kept the existing `.system` diagnostic: splitting that internal error was optional and changes no acceptance or mathematical guarantee.
- Factored the reduction guard and documented the sharing precondition for direct `buildTreeFrom` calls. The public `buildTree` path still performs original-query reduction once.
- Corrected the reduction producer's shape documentation to reference the proved preprocessing length invariant.
- Added `ImportError` to the oracle's top-level failure handler.

The first CI attempt failed before compilation because the fixture target registration invalidated the lakefile blob exemption for historical factorization measurements. An exact blob-transition exemption now records the runtime-neutral registration and the existing native-versus-elaborator `HexBasic` distinction; the benchmark freshness checker passes locally. No new performance completion is claimed.

PR #10403, opened 2026-09-22
feat(sign-det): implement checked Thom descriptors and root operations

kim-em:
Selected squarefree roots now have checked partial/full Thom descriptors, completion, full root lists, selected-query signs, cross-polynomial comparison and re-encoding. Every descriptor reconstructs its formal derivatives and carries recursive count-one evidence bound to its sign operation, exact head, open interval and immutable context. Sparse `SignTable` lookup includes omitted conditions, and its finite correspondence theorem covers every lookup.

Completion preserves the old derivative constraints; root lists use the largest-differing-index Thom rule and preserve every input encoding row. Insertion checks a common head and canonical full derivative slots; finite sortedness is proved under explicit comparator laws, while strict real-root order remains a companion gate. Joint selected-sign replay proves that every matching finite observation has the returned signs. Comparison constructs a common product with the shared polynomial gcd/division, checks its root-union identities and squarefree domain, and re-encodes both selections over the whole line. Old defining equations, derivative signs and strict endpoint queries retain the root identity even when an endpoint is a root only of the other polynomial. Noncanonical equal-valued coefficients use zero-difference polynomial checks; context and operand bindings retain literal equality.

The companion proves correspondence of the actual formal derivatives, the root union for arbitrary accepted common-product identities under noninjective interpretation, and the actual finite-endpoint inequalities. Literal sparse-row adapters keep ordinary-kernel descriptor/sign replay independent of opaque table construction. Kernel probes accept a singleton descriptor and selected sign and reject wrong signs and stale contexts, with checked axiom inventories. A nonempty derivative-slot kernel probe evaluates both formal derivatives of a quadratic and rejects a copied sign-equivalent query certificate. Invalid zero heads report domain errors.

The finite algebra layer now proves `solveSystem_complete`: the actual rational solver succeeds on every checked integer system, preserves literal input orders and returns the same counts. The proof uses the core rank API at the executable rational field instance and establishes positivity/divisibility of the actual common denominator and the exact integer inverse identity. Shared rational list/matrix encoders have acceptance and decoding proofs, and both leaf systems are constructed from arbitrary finite observations with their inverse witnesses checked by the ordinary kernel. The parent-system theorem derives shape, distinctness, count and matrix checks from the actual child bases and independently complete child supports, without assuming an accepted parent. These theorems remove finite solver gaps without assuming that replay acceptance supplies root semantics.

These remain diagnostic constructors: internal `BuildError` is separate from malformed/absent/ambiguous input, invalid target domains and absent target roots. This PR does not claim the final total `determinePrepared`, domain-exact validation or total selected-root APIs. It does not replace the outstanding producer/semantic correctness proofs with successful compilation or fixture results.

Validation: full `lake build`; computational/companion conformance and emitter builds; the actual filtered oracle harness with 101 cases and 16 adversarial Python tests; DAG, copyright, source-size, conformance-registration and diff checks. The 59 rational table cases compare FLINT qqbar and the existing real-algebraic API; 27 descriptor, 10 comparison and five re-encoding cases use independent exact FLINT roots/evaluation and numerical root order. Mutation tests reject lexicographic root order, omitted roots, incorrect signs, unrelated common heads, unremoved common factors and false validity/re-encoding conclusions.

Refs #10377. Remaining gates include general producer completeness and domain-exact totality; full root/sign correspondence through #10389 and the separately required Tau Ceti Thom/BKR foundations; shared domain/table verification, finite DAG serialization, nested/refined contexts and supplied-proof integration; the final total descriptor/root interface for sample consumers; extension/infinitesimal conformance with the required pinned Z3 oracle; and the full Phase-4 runtime, fresh-module proof-cost, comparisons and profiling evidence. Root extraction currently repeats the whole replay and derivative construction once per root. Comparison constructs four BKR tables and four prepared domains, including for identical heads; sharing and cost attribution remain required. #10352 remains a separate rank-performance obligation. No Phase-4 gate or full issue completion is claimed.


kim-em:
Claude Opus reviewed this implementation through the second-opinion skill. I verified the findings against source and addressed the substantive correctness and coverage concerns in 861e6b51c:

- Insertion now calls the guarded full-descriptor comparator. Regressions reject both unrelated heads and mismatched partial derivative slots that previously produced a strict raw-word comparison.
- Added finite insertion/enumeration sortedness proofs under explicit comparator laws and checked their axiom inventories. The README explicitly retains the semantic strict real-root ordering gate.
- Added an ordinary-kernel literal replay probe with derivative index 2. It forces both formal derivatives and rejects a copied certificate for a sign-equivalent constant query.
- Zero heads now report domain failure before descriptor-shape diagnostics. The independent oracle checks this classification and rejects the old reason.
- Documented the exact outstanding repeated work: one whole replay and derivative reconstruction per output root, and four table/domain constructions per comparison. Sharing and the performance gates remain open under #10377.
- Clarified that partial derivative indices pair positionally with signs and may be permuted; canonical full encodings use [1,…,n].

The optional encoder-exposure suggestion does not require a change: exposing these two existing value definitions deliberately supports proofs of their actual implementation, with no runtime body/type changes. The committed exact-blob performance exemption records that narrow distinction and makes no proof-cost or SignDet performance claim. The squash PR title already uses feat.

Validation after these fixes: computational and companion builds/conformance, literal kernel probes and axiom inventories, the actual oracle harness (101 cases, 16 Python mutation tests), DAG/copyright/conformance registration and diff checks all pass locally. CI is running on the updated head. No missing Tau Ceti foundation, producer totality or Phase-4 gate is claimed complete.

PR #10404, opened 2026-09-22
feat(sign-det): prove producer replay and share checked root tables

kim-em:
Successful BKR tree construction now has an algebraic proof that its independent replay passes, and full-root extraction reuses one checked table instead of replaying it for every root.

The companion tracks the actual rational/scaled solver returns, literal input orders, retained rank certificate, indexed reductions and prepared Tarski certificates. It proves local node acceptance under the shared generic coefficient interpretation laws, propagates checked preprocessing slices and child/support bindings through the actual balanced recursion, and shows that a successful tree cannot fail `buildPrepared`'s final replay guard. A conformance theorem instantiates the result on noncanonical coefficient representations without introducing a field instance on them. `buildNode_complete` proves the finite assembly step for both solver paths: a checked candidate system whose values equal the actual prepared-query values produces a node with the same counts, with the supplied inverse bound exactly when present. This lower-level lemma neither supplies complete support nor assumes a headline query-soundness theorem. `Node.parent_system` transports the finite child-product system to the exact row dimension and tensor inverse consumed by node construction, using the shared `parentInverse` definition. Executable construction and companion statements share their preparation and per-row reduction definitions.

For positive-degree heads, `buildRoots` now extracts count-one descriptors through proof-backed table membership. It avoids per-root whole-tree replay, derivative reconstruction and full-index distinctness checks. `rootsFromTable_eq` proves exact agreement with the previous literal checking path, including diagnostics; permutation and conditional strict-sortedness results now reach the public `buildRoots` entry point. `buildRoots_raw` also retains the exact caller context, head, interval and canonical full derivative indices; `buildRoots_constant` proves that successful constant-head enumeration is empty by descriptor shape alone. Constant heads retain the literal per-descriptor checking path. Full-index lists and head checks still repeat during insertion; their possible O(N²n) cost remains explicit. Supplied unchecked certificates still undergo full literal replay. A new full two-derivative tree and the shared descriptor constructor are checked in the ordinary kernel.

Validation: full `lake build`; computational/companion conformance and emitter builds; ordinary-kernel literal probes and axiom inventories; actual filtered oracle harness (101 exact cases, 16 adversarial Python tests, unchanged fixtures); DAG, copyright, source-size and diff checks.

Refs #10377. This proves producer acceptance, not total producer success or arbitrary-certificate semantic soundness. General finite producer completeness, domain-exact total APIs and full root/sign correspondence remain open; semantic proofs require #10389 and the separately required Tau Ceti BKR/Thom foundations. Shared domain verification inside the query tree, finite DAG serialization, nested/refined contexts and supplied-proof integration, final total sample-consumer APIs, extension/infinitesimal conformance with pinned Z3, and full Phase-4 runtime/fresh-module/comparison/profile evidence remain required. Cross-polynomial comparison still constructs four tables/prepared domains. The eliminated repeated work is a structural property of the new code; no measured runtime, Phase-4 completion or full issue completion is claimed. #10352 remains the separate rank-performance obligation.


kim-em:
Claude Opus completed an independent source review and found no soundness defect. Commit 76572c320 addresses the substantive findings: shared node preparation/reduction definitions, public `buildRoots` extraction/permutation/sortedness proofs, one vector-transport lemma, and an exact parent-system dimension/inverse adapter. Documentation now states the remaining insertion guard cost and avoids asserting unproved constant-root semantics.

The suspected tactic issue was checked against a successful build and requires no change. Full local `lake build`, computational/companion conformance, ordinary-kernel probes, axiom inventories, 101 FLINT cases with 16 adversarial tests, and source/DAG/registry checks pass. The held subsequent infinitesimal tranche also passes its 26 Z3 cases and 14 adversarial tests, but is excluded from this PR. These checks do not establish the outstanding semantic or Phase-4 gates.

kim-em:
The focused Claude Opus follow-up again found no soundness defect or behavioral change. Commit 0183d25b1 addresses its substantive remaining findings: the executable and companion now use one `parentInverse` definition, and `rootsFrom_raw` / `buildRoots_raw` retain the literal domain and canonical full derivative slots through sorting. `buildRoots_constant` proves the constant-head success case structurally. The generic transport lemma is now `Vector.toList_transport`.

Validation: full local lake build; computational and companion conformance including the new theorem axiom inventories; all 101 FLINT cases and 16 adversarial tests; all held-tranche 26 Z3 cases and 14 adversarial tests. No semantic foundation or totality gate is claimed. The suggested direct equality of returned count vectors is not used: the returned node has its own dependent size, so the list equality avoids adding a size transport to that lower-level interface.

PR #10405, opened 2026-09-22
test(sign-det): verify nested infinitesimal roots against pinned Z3

kim-em:
Sign determination and Thom operations now have independent exact conformance over one and two infinitesimal levels. The new 29-case emitter exercises the corrected de Moura–Passmore polynomial, nested roots, zero/repeated queries, invalid domains, partial/full descriptors, completion, selected signs, cross-polynomial comparisons and re-encoding. It also rejects changed derivative-query/context/head bindings and multi-query tables presented as leaves. The leaf-arity checks are not identity-preserving incomplete-support forgeries.

The Lean fixtures reuse lawful `RationalFn` arithmetic with test-only least-coefficient sign callbacks. A separate Python oracle reconstructs every serialized coefficient in a fresh Z3 RCF context, computes exact roots/signs and verifies the results without decimal approximations. The package and numeric runtime version are pinned to z3-solver 4.15.4.0. Required cases enforce their coefficient depth and the corrected Passmore polynomial, and cover all five descriptor-error reasons. Sixteen adversarial Python tests reject omitted support, malformed coefficients and contexts, wrong order/completion/re-encoding and version drift. The oracle extends the existing single CI job.

Validation: full local `lake build`; 29 exact infinitesimal fixtures and 16 adversarial tests; existing 101 FLINT fixtures and 16 adversarial tests; source, DAG and conformance-registry checks. The existing factorization measurement freshness check has a literal baseline-to-current Lake-file exemption: the only new target changes are confined to conformance, so factorization runtime inputs remain unchanged.

Refs #10377. These are test coefficient providers, not production `HexOrderedFn` registrations, a tower implementation or semantic coefficient certificates. General producer totality, final total consumer APIs, arbitrary-certificate root/sign correspondence through #10389 and the separately required Tau Ceti Thom/BKR foundations, shared/nested coefficient-proof graph serialization and transport, and the complete Phase-4 comparison/coverage/proof evidence remain open. Oracle agreement does not discharge those proof or performance gates.


kim-em:
Addressed all seven Opus findings in 0f0fa8645:

- Corrected the negative-replay descriptions to match the actual context/head/query mutations and leaf-arity guard.
- Enforced required coefficient depths and the corrected Passmore polynomial, including adversarial substitution checks.
- Added absent, malformed and invalid-domain descriptor cases. All five error reasons now occur in the extension fixtures.
- Kept the exact package pin, switched to the numeric runtime version, removed the unverified commit-provenance claim, and tested both version checks.
- Added Z3 to the required-oracle preflight and documented regeneration commands in both READMEs.
- Moved only generic serialization validators into a shared module; both exact root computations remain separate.

Validation: emitter build passed; 29 Z3 fixtures and 16 rejection tests passed; existing 101 FLINT fixtures and 16 rejection tests passed; source/DAG/registry checks passed. The earlier CI freshness failure is fixed by the runtime-neutral Lake-file exemption in 47269eb27. Proof and performance gates remain as listed in the PR description and #10377.

PR #10406, opened 2026-09-22
feat(sign-det): replay and encode shared certificate graphs

kim-em:
Repeated BKR subtrees can share literal graph entries while retaining both ordered child edges. The checker validates each entry against an already accepted prefix, rejects cycles and forward references, checks unreachable entries, and preserves exact context, domain, query and witness bindings. Accepted graph replay produces an accepted literal tree, and graph evidence can validate a supplied root descriptor without repeating the whole tree check.

The encoder visits a supplied tree in left-to-right postorder and interns entries by exact equality; hashes only narrow the lookup. Proofs follow the actual encoder and checker: structural expansion returns the exact original tree for every input, and checking an encoded graph returns exactly the original tree checker result. Encoding preserves rejection as well as acceptance. Checked graph replay also returns the same literal tree as structural expansion. Collision tests change an inverse witness without changing the entry hash and verify that the bad node is retained and rejected. Ordinary-kernel probes cover shared children, descriptor extraction, cycles, truncation and an encoded forged inverse witness, with exact axiom audits.

Descriptor extraction checks raw shape and context before replay, and `descriptor_replay` proves exact descriptor agreement on any accepted supplied graph's actual replay result; `descriptor_encode` extends agreement to rejection of encoded trees. A public failure lemma avoids opening the descriptor's private constructor boundary. Evaluated expansion probes cover shared values, cycles, forward references and missing roots; structurally valid false unreachable witnesses expand but still fail checked replay. Query/chain/rank/system/reduction/node equality instances live in their owning modules. Explicit field decisions avoid proof-induced casts during kernel reduction, and ordinary-kernel probes check complete equal and unequal query certificates, rank witnesses and nodes, including nonempty reduction/preparation records that also pass replay. Cheap context and ordinary-field comparisons precede support-length computation. Both expansion and checked replay use one generic memo-prefix and hash-consing invariant. The computational SPEC records the same-level component separately from nested coefficient evidence.

Validation: exact-tranche full `lake build` passed (14,671 jobs); graph conformance and ordinary-kernel/axiom probes passed (369 jobs); source/DAG/conformance-registry checks passed. All 105 freshness tests and the actual factorization/graph-isomorphism freshness checks passed for the unchanged Lake registration. The Lake-file exemption records only the graph conformance registration under the standard `HexSignDet.CrossCheck` module name.

Refs #10377. This tranche concerns same-level BKR graph structure and literal replay. It does not serialize coefficient values or lower-level coefficient proofs, prove arbitrary-field root/sign semantics, or complete Phase-4 evidence. General producer totality and the final total consumer interfaces, nested coefficient-proof serialization/transport, semantic correctness through #10389 and the separately required Tau Ceti Thom/BKR foundations, and remaining performance/proof evidence stay open.


kim-em:
Addressed the independent Opus review:

1. Added evaluated `expand?` probes for the complete/shared trees, cycles, forward edges, missing roots and invalid unreachable references.
2. Proved `Dag.descriptor_encode`: graph and tree descriptor extraction return exactly the same `Option` for every encoded tree, including rejection. Added its exact axiom audit.
3. Moved exported equality instances to their owning modules. Explicit field decisions now cover query chains, Tarski certificates, rank witnesses, systems, reductions and nodes. New ordinary-kernel probes caught nonreducing derived rank/system equality; those implementations were replaced and complete equal/unequal node probes now pass too.
4. Factored memo growth, encoder validity and collision-safe insertion into `DagMemo`. Structural expansion and checked replay instantiate that one invariant; their short recursive inductions retain the distinct certificate premises.
5. Moved cheap descriptor well-formedness and context guards before graph replay.
6. Added false but structurally valid unreachable-witness rejection and split-arity probes; documented the deliberate difference between structural expansion and arithmetic replay.
7. Documented occurrence-based encoder cost and entry-remapping graph transformations. No allocation or Phase-4 claim is inferred from this change.
8. Recorded that cross-level coefficient-sign evidence needs a separate type, with this fixed-domain DAG as its same-level component. No speculative binding redesign was added.
9. Renamed the newly introduced conformance module to the standard `HexSignDet.CrossCheck` name and updated its existing CI registration; no workflow change or CI fan-out.

The exact revised tranche passes full `lake build` (14,671 jobs), conformance/kernel/axiom probes (369 jobs), all 105 freshness tests and actual factorization/graph-isomorphism freshness checks. All root-semantic, nested-evidence and remaining performance gates on #10377 stay open.

kim-em:
The focused follow-up review confirmed the invariant generalization, descriptor case split and instance ownership. Its remaining actionable findings are addressed:

- `Dag.descriptor_replay` now proves exact descriptor agreement for any accepted supplied graph and its actual returned tree. `descriptor_encode` uses it on acceptance and retains its complete rejection argument.
- Removed `import all HexSignDet.Descriptor`. The owning module exposes only the `Descriptor.ofReplay_none` failure theorem needed by the proof; its constructor stays private.
- Added a node with nonempty, valid query preparation and product-reduction chains. Ordinary-kernel probes compare all those records, reject a changed preparation, and verify actual replay of the node.
- Added acceptance of a valid unreachable duplicate entry alongside the existing rejection controls.
- Renamed the generic encoder invariant's field to `fold`, appropriate to both structural expansion and checked replay.
- Node equality compares context and ordinary fields before computing support lengths; exact whole-node equality still reduces in the ordinary-kernel probes.
- The computational SPEC now describes same-level graph replay/expansion/encoding and distinguishes it from the separate nested-evidence requirement. Occurrence-based encoder cost and the remaining allocation/performance gates stay recorded under #10377; no performance completion is claimed.

The revised exact tranche passes full `lake build` (14,671 jobs), `HexSignDet.CrossCheck` and all ordinary-kernel/axiom probes (369 jobs), plus source/DAG/line-count/copyright/conformance-registry checks. CI will check this updated commit.

PR #10407, opened 2026-09-22
Reuse validated Tarski domains in BKR node and graph replay

kim-em:
BKR replay previously checked the same endpoint guards and squarefree signed-remainder witness separately for every moment. Graph replay now validates the first entry's domain once and offers it to all nodes. Each node prefers that domain when it matches its first witness; otherwise nodes with several moments validate their own first domain once. Single-moment nodes without a matching graph domain use full replay directly. Later literal matches reuse the selected evidence, while different witnesses fall back to full replay. Acceptance and rejection are unchanged for arbitrary supplied certificates.

The shared Tarski kernel factors literal bindings, domain checks and query checks. `Domain.replay?` creates checked evidence only after the actual supplied domain replay succeeds, retaining the entire context, head, interval and squarefree witness under fixed sign operations. `checkCached_eq` proves identical Boolean results to full replay for every cache and certificate. `checkMoment_eq` and `Node.check_eq` lift that agreement through the actual moment and node algorithms, including empty systems and rejection. Existing companion proofs use the expanded complete-checker contract; no root-sum premise or producer-provenance assumption replaces replay.

The complete cached query checker is exposed through the HexSturm frontend. The guarded cache-hit operation is named `checkHit` and explicitly rejects cache misses; it shares the query-dependent body with the complete fallback checker. The cache uses the existing literal equality instances in the signed-remainder and Tarski modules. Their field decisions reduce in the ordinary kernel. Conformance exercises accepted shared replay, malformed witnesses, failed local cache construction, one-moment nodes, changed contexts and endpoints, and alternate valid witnesses in later BKR moments and graph nodes. Proof audits permit only the expected standard axioms.

Validation on this exact tranche: full `lake build` passed (14,672 jobs); the final graph conformance and ordinary-kernel/axiom probes passed (370 jobs), and shared-query, Sturm/real-roots/sign-determination companion conformance passed (9,510 jobs); source, dependency DAG, line-count, copyright, conformance registry and whitespace checks passed. The broader development branch also passed its full build and the Mathlib-free check (58 benches and 2,691 build-only proof probes).

Refs #10377. Each node selects one validated domain witness; later witnesses that differ from that selected domain still incur full replay at each occurrence. Cache performance on heterogeneous witnesses and one-moment nodes remains an explicit measurement obligation. Lower-level coefficient-proof dependencies and byte serialization, the remaining producer/consumer and descriptor interfaces, and the full Phase-4 performance evidence remain open. These checks establish finite replay agreement, not the pending arbitrary-field root/sign semantics. Full semantic correctness requires #10389 and the separately required Tau Ceti Thom/BKR foundations. No speedup or performance-gate completion is claimed.


kim-em:
Addressed the independent Opus review in 3d8fca4e2:

- Graph reuse now composes with node-local reuse: a mismatch against the node's first witness selects that node's own validated domain. Later differing witnesses still take the complete fallback path.
- Nodes with at most one moment do not construct a local cache. A matching existing graph cache can still be reused.
- Added direct compiled probes for alternate-domain cache selection, a one-moment node with no cache, and tree rejection after invalid first-domain evidence prevents local cache construction. Existing arbitrary-cache equality proofs and ordinary-kernel soundness/axiom probes still pass.
- Added `Sturm.checkCached` and its unconditional equality theorem, keeping BKR behind the prepared-query frontend. Renamed the guarded cache-hit API to `checkHit`, documented its rejection on a miss, and shared its query-dependent body with `checkCached`. Removed the reviewed theorem defaults and bare simplifications.

The full build passed (14,672 jobs), shared-query and BKR graph conformance passed (370 jobs), and DAG, line-count, copyright, whitespace and actual factorization/graph-isomorphism freshness checks passed. A fresh follow-up Opus review is running alongside CI.

I agree that cache speed must be measured. This partial PR claims unconditional finite replay equivalence, not a speedup or Phase-4 completion. The held benchmark work is not yet in this tranche; cache-specific adjacent AB/BA measurements for heterogeneous witnesses and single-moment nodes remain explicitly open under #10377. The separate moment-power comparison currently running does not discharge that cache-specific obligation.

kim-em:
The follow-up Opus review found no correctness/soundness defect or merge blocker and confirmed the earlier substantive fixes. The final follow-up tightens the matching-shared-domain probe for a one-moment node, asserts that the supplied domain really validated in the alternate-witness probe, and annotates `Node.cache`'s return type. `checkBody` now explicitly states that it omits domain validation and cannot certify a Tarski result alone. A private exposed helper was rejected by the module's public-definition boundary, so its public component role is documented, matching the existing `checkQuery` surface.

The remaining low-severity cache heuristics are explicitly documented for performance measurement: one initial graph domain may have only one use; selection is based on moment zero rather than scanning for the most frequent witness; and literal binding is checked again when each moment is replayed. These are finite cost obligations, not soundness gaps, and this PR makes no optimality or speedup claim.

The final conformance passed (370 jobs), the full build passed (14,672 jobs), and line-count/copyright/whitespace checks passed. CI is checking the updated head. Required semantic and Phase-4 gates remain open under #10377.

